PR #272, opened 2026-02-04
AMS Lite to Spoolman

adrilop002:
## Description

Adds a new helper script scripts/bambuddy_etl_spools.sh to extract filament usage from bambuddy.db into a local tracking.db, generate CSV summaries, and (optionally) sync spool remaining weight to Spoolman.

The script is designed to be safe for repeated runs and uses INSERT OR IGNORE to avoid duplicates. It also warns if spoolman_map (tag_uid -> spool_id) mappings are missing, and only syncs mapped spools.

## Related Issue

Fixes #

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

* Added scripts/bambuddy_etl_spools.sh to ingest print archive data into tracking.db and generate:
- out/filamento_por_color.csv
- out/filamento_por_bobina.csv (includes spool color name, dominant filament type, and optional spool_id)

* Added basic color mapping (color_map) and spool mapping table (spoolman_map) support.

* Added Spoolman sync: PATCH /api/v1/spool/{spool_id} updating remaining_weight (overwrite mode), with fail-soft behavior and logging to out/spoolman_push.log. (remaining_weight and used_weight are not set together.)

## Testing

Ran ./scripts/bambuddy_etl_spools.sh against a real bambuddy.db and verified generated CSV output.

Confirmed Spoolman updates by checking the remaining weight values after running the script.

## Checklist

- [ ] My code follows the project's coding style


maziggy:
If we want to add the script, I need some docs for the Wiki.

adrilop002:
This script (stored under the scripts/ folder) runs an ETL pipeline to compute filament usage from Bambuddy’s print archives and prepare the results for reporting and Spoolman synchronization.

- It reads from bambuddy.db (Bambuddy’s database, read-only) and processes only completed jobs (status='completed').

- It creates or updates a local working database, tracking.db, where it stores normalized tables for: print facts (print_facts), AMS/tray snapshots per print (print_trays), and per-spool consumption allocations in grams (allocations).

- The workflow is incremental: it stores the last processed archive_id in ingest_state, so subsequent runs only ingest new prints instead of recalculating everything.

- If a print contains an AMS/extruder mapping, the script assigns filament usage to a specific spool (tag_uid) with high confidence; if mapping data is missing, it attempts a color-based backfill (only for single-color prints) using color_to_tag_uid.

- It creates an out/ directory (if missing) and writes two CSV outputs:

** filamento_por_color.csv: total grams used per color (plus a “MULTI” bucket for unassignable multi-color usage).

** filamento_por_bobina.csv: per-spool breakdown (tag_uid) including used grams, nominal grams, and an estimated remaining amount.

- (Optional) If SPOOLMAN_URL is set, it syncs each spool’s remaining_weight to Spoolman via its API and logs the outcome to out/spoolman_push.log.

- To avoid manual mapping for new spools, it can auto-fill spoolman_map by querying Spoolman: when a spool is linked, it resolves spool_id using the tray_uuid stored in Spoolman under extra.tag.

- It can also auto-populate the local color_map from Spoolman using filament.color_hex and filament.name, so new colors are labeled automatically (while preserving any local overrides you define).

maziggy:
Hey @adrilop002, 
thanks for this PR — you've identified a real gap: AMS Lite users (A1/A1 Mini) don't have RFID on their spools, so the native per-filament tracking we merged in PR #277 can't identify their spools and won't work for them.

  Your color-matching approach to allocate usage is a creative solution to that problem. However, we have some concerns about the implementation as a standalone bash script:

  - Architecture: A 586-line bash script with embedded SQL that reads bambuddy.db directly operates outside Bambuddy's application layer. This makes it hard to maintain
  and could conflict with concurrent database access.
  - Security: Shell variables interpolated into SQL strings ($tag_uid, $spool_id) create injection risks. The Spoolman PATCH calls also lack authentication support.
  - Localization: Comments and output are in Spanish ("Falta comando", "filamento_por_bobina", "desconocido"), which would need translating for the broader user base.
  - Spoolman conflict: The script overwrites remaining_weight from scratch, which would conflict with PR #277's incremental tracking for users who have both AMS and AMS
  Lite.

  What we'd love to see instead is the color-based spool matching integrated into the existing Python tracking service (backend/app/services/spoolman_tracking.py). The  logic would fit naturally as a fallback in _resolve_spool_tag() — when tray_uuid and tag_uid are both empty (AMS Lite), fall back to matching by tray_color. That way AMS Lite users get the same native tracking experience as full AMS users.

  Would you be interested in taking that approach? Happy to help point you to the right places in the codebase. The CSV export idea is also cool — if you want to pursue  that separately, we could add it as a Python API endpoint.

  Thanks for the contribution and for highlighting the AMS Lite gap!

PR #361, opened 2026-02-13
Add material mismatch check

Keybored02:
## Description

Add material mismatch check when assigning spools. This is implemented for non-BL spools only. 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made
- Check spoolmaterial against traymaterial (only types, no colors)
- Ask user for confirmation in case of mismatch with confirm modal
- Updated locales

## Screenshots
image

## Testing

## Checklist


maziggy (changes requested):
Bug — spool assignments query too restrictive:

  PrintModal/index.tsx:
  enabled: isLibraryFile && mode === 'reprint',
  But the warning check in handleSubmit fires for mode === 'reprint' || mode === 'add-to-queue' and doesn't require isLibraryFile. So the warning will never fire for
  archive reprints or add-to-queue — spoolAssignmentsByPrinter will always be empty. Should be:
  enabled: (mode === 'reprint' || mode === 'add-to-queue') && assignmentMode === 'printer',

  Bug — mismatch dialog closes before mutation completes:

  AssignSpoolModal.tsx handleConfirmMismatch() calls setShowMismatchConfirm(false) immediately after assignMutation.mutate(), hiding the dialog before the mutation
  finishes. This makes the isLoading prop on the ConfirmModal useless, and if the mutation fails the user just sees a toast with no way to retry. Remove the state cleanup
   from handleConfirmMismatch and let onSuccess (which already does the same cleanup) handle it:
  const handleConfirmMismatch = () => {
    if (!pendingAssignId) return;
    assignMutation.mutate(pendingAssignId);
  };

  Untranslated strings:

  de.ts, fr.ts, and ja.ts have English text for printModal.insufficientFilamentTitle, insufficientFilamentMessage, insufficientFilamentLine, and printAnyway. Only it.ts
  has proper translations. (The assignMismatch* keys are correctly translated in all locales.)

  Minor — confirmClassName prop:

  The warning variant already defines text-black in its button styles. Adding a generic confirmClassName prop to ConfirmModal just to pass !text-black for one caller
  suggests the Button component is overriding the text color. Might be simpler to fix the specificity in the existing warning variant styles rather than adding a new
  prop.

Keybored02:
Yeah, sorry, forgot to switch branches when committing the weight check feature, it's not ready yet. Will fix asap.

Keybored02:
Yes, I've tested the feature a bit more with the fixes you proposed the last review and it's working as intended. If it's fine by you I would leave it here. Otherwise you can reject it and I can open a new one just for that. 

Sorry for the trouble.

maziggy:
Doesn't work for me?!

Screenshot 2026-02-15 at 09 47 54
Screenshot 2026-02-15 at 09 48 17

maziggy:
Ahhh now I see. This is for assigning spools to the inventory.

This was already completely refactored and it now only shows matching spools.

Need to revert your PR.

For the future I would suggest to just check with prior to submitting a PR. I'm working heavily on the code and things are changing quickly.

Keybored02:
I see, that's partially why I'm eager to push, the dev speed around here is really fast.
Are you sure you want to force showing only compatible filaments? Some users do not match printer profile to filament to bypass some FW locks, like no TPU on L nozzle on H2D, or no brittle filaments in the R. So auto matching would prevent you form tracking in those cases.

Issue #184, opened 2026-01-30
[Feature]: Plate Swap Integration (A1/A1 Mini) Swapmod and Jobox


MicGier:
### Problem or Use Case

For the A1 and A1 Mini models, there are modifications that allow for automatic plate changing. These modifications are implemented via the corresponding G-code.

### Proposed Solution

It would be great if this feature could be integrated as an option into the queue to create a production line. For the A1 Mini this would be Swapmod for example. For the A1 it would be Jobox.

### Feature Category

Print Queue & Scheduling

### Priority

Would improve my workflow

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Need way more details.

MicGier:
I’ve been chatting with the developers of the Plate Swap System on Discord. They mentioned they’ve been in touch with you about this.

maziggy:
Nope.

MicGier:
Oh, ok. 

So I'm talking about these two projects:

https://swap-systems.com  (for the A1 mini)

Swap Code can be found here:

https://swap-systems.com/automation-approaches/


and

https://jobox.tech (for the A1)

SWAP GCODE:

`Plates swapping gcode
;This code ejects a plate and loads a new one
G90
G0 Y260 F2000
G0 Y265 F500
G0 X-33 F2000
G0 Z-12 F2000
G0 Y250 F2000
G0 Y240 Z250 F2000
G0 Y100 F1000
G0 Y0 F2000
G0 Y15 F2000
G0 Y33 F500
G0 Y260 F2000
G0 Y210 F2000
G0 Y10 F2000
G0 Y245 F2000
G0 Y265 F500
G0 Y-2 F2000
G0 Y15 F2000
G0 Y-2 F2000
G0 Y20 F2000
G0 Y100 F2000
Plate load only gcode
;This code only loads a new plate (if there is none)
G90
G28 XY
G0 X-33 F2000
G0 Z30 F5000
G0 Y0 F2000
G0 Y260 F2000
G0 Y210 F2000
G0 Y265 F2000
G0 Y-2 F2000
G0 Y15 F2000
G0 Y-2 F2000
G0 Y20 F2000
G0 Y100 F2000`

markrobotsmith:
This would be a great feature. I've written a webapp for this. I can look at integrating it here. Some things to note:
[...]

PR #390, opened 2026-02-15
First attempt to automation, Location == automation means printer aut…

pleite:
## Description

First attempt to automation, Location == automation means printer aut…o clear on successful completion

## Related Issue

Implementation of Feature #184 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- For a printer with Location as the single word automation when a print is successful the plate is automatically cleared
- There are no changes to the printer model or the UI on this commit
- The UI will briefly show the Clear plate button while the upload and MQTT command are being delivered on the background.

## Testing

## Checklist

- [ ] My code follows the project's coding style

## Additional Notes

This is plain hack to get the feature started, with some guidelines on what to transpose to the UI and printer model

### DIFF
--- a/backend/app/services/printer_manager.py
+++ b/backend/app/services/printer_manager.py
@@ -106,6 +106,7 @@ class PrinterManager:
         self._current_print_user: dict[int, dict] = {}  # {printer_id: {"user_id": int, "username": str}}
         # Track plate-cleared acknowledgments for queue flow
         self._plate_cleared: set[int] = set()  # printer_ids where user confirmed plate is cleared
+        self._automation: dict[int, bool] = {} # Cache printer is automated, auto clear
 
     def get_printer(self, printer_id: int) -> PrinterInfo | None:
         """Get printer info by ID."""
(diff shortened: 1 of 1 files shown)


maziggy:
Thanks for kicking off the automation discussion! I love the enthusiasm for #184.

  I'm going to close this one though — using the location field as a magic keyword isn't the right approach for this. A proper implementation would need:
  - A dedicated automation/auto-clear setting per printer (in the UI, not a location string)
  - Safety checks (e.g. what if the print "completed" but the model is still stuck to the plate?)
  - There's also a bug where self._automation[printer_id] would KeyError for printers connected before the dict is populated

  That said, auto-clear for print farm/queue workflows is definitely something we want to support. I'll keep #184 open for a proper implementation.

pleite:
I can spend some time to bring you a proper PR, any naming conventions you want to share for the feature to present it to the UI, probably the same way as the auto archive checkbox? 

Second, if a print is left on the bed, for example with GCode that misses the automation there were two considerations, the plate check fails as the printer does the tap touch and misses the proper plate height and we could rely on the camera machine learning - thus making it more robust. There are not magic tricks here, the owner needs to know what is doing.

Third by fixing the first the data should be populated outside of the printer manager. The intention here was to provide a quick walkthrough of the result, what was the minimal effort to achieve.

Thanks for taking a look, next one is going to be more elaborate.

PR #421, opened 2026-02-17
Added translation to Brazilian Portuguese on existing strings.

wreuel:
## Description
Added translation for Brazilian Portuguese

## Related Issue

Fixes #

## Type of Change
Translation cretaed

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Just added a file with the translations.

## Testing

## Checklist

## Additional Notes
There still some places that don't use translation that need to have it.


wreuel:
> Thanks for adding Brazilian Portuguese @wreuel! Great to see more language support.
> 
> However, there are a few issues that need to be fixed before this can be merged:
> 
> 1. Truncated value
> 
> common.no is 'ão' — it should be 'Não'. The leading "N" was cut off. This key is used throughout the entire UI.
> 
> 2. Untranslated English strings
> 
> * queue.itemCount_plural: '{{count}} items' should be '{{count}} itens'
> * settings.email.securityOptions: all three values are verbatim English — "Port" should be "Porta", "None" should be "Nenhum"
> 
> 3. Major: key structure doesn't match en.ts
> 
> This is the main issue. Many sections in pt-BR.ts have completely different key names than the current en.ts. For example:
> 
> * system section: en.ts has cpuUsage, memoryUsage, diskUsage — pt-BR has instanceId, osInfo, pythonVersion (different keys entirely)
> * library section: en.ts has addFilament, editFilament, vendor — pt-BR has importFromMakerworld, downloads, rating
> * inventory: en.ts has ~80+ keys — pt-BR has ~25 with different names
> * Same pattern in editArchive, kProfiles, virtualPrinter, modelViewer, spoolman, and others
> 
> When the keys don't match, i18next can't find them and silently falls back to English, so these sections will appear untranslated to users.
> 
> 4. ~29 top-level sections are completely missing
> 
> Including smartPlugs, dashboard, restoreBackup, configureAmsSlot, photoGallery, tagManagement, notificationTemplates, richTextEditor, and many more.
> 
> Recommendation:
> 
> The easiest approach would be to start fresh using the current en.ts as the exact template — copy it to pt-BR.ts and translate only the string values, keeping all key names and nesting structure identical. That way every key maps correctly and you get full coverage. The sections you've already translated well (nav, common, printers, archives, queue, settings core) can be reused.

Thanks a lot for the detailed review and for taking the time to point these things out — I really appreciate it 🙏

Regarding your notes:

• The truncated “Não” and the untranslated strings you mentioned are absolutely valid — thanks for catching those. I’ll fix them right away.

• About the key structure mismatch, I tried to verify this on my side but couldn’t find the examples you referenced. For instance, I searched the pt-BR file for instanceId, osInfo, and pythonVersion, and none of those keys appear there.

The en.ts file I used as the base was the one located at:
./frontend/src/i18n/locales/en.ts

On the 0.2.1b branch it contains 3441 lines, which matches the file currently in main as well. So as far as I can tell, the pt-BR file was created directly from that version as a template.

It’s possible we might be looking at different revisions or that the structure changed in another branch — could you point me to the exact source/version you’re comparing against? That would help me align things properly.

Thanks again for the thorough feedback — I’ll start by fixing the confirmed issues while we clarify the structure differences 👍

PR #414, opened 2026-02-17
Fix: Spool Empty Weight Catalog Matching by ID

Keybored02:
## Description

The `SpoolWeightPicker` component matched catalog entries by weight value only, using `find()` which returns the first match. This approach is ambiguous when duplicate weights exist in the catalog.
Implemented unique identification using catalog entry IDs:
- Store the catalog entry ID (unique primary key) whenever a user selects from the catalog
- When editing, use the ID to look up the exact catalog entry

## Related Issue

Fixes #413

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition or update

### Backend
- Added `core_weight_catalog_id: Mapped[int | None]` field to track which catalog entry was used
- Added `core_weight_catalog_id: int | None = None` to `SpoolBase` schema
- Added `core_weight_catalog_id: int | None = None` to `SpoolUpdate` schema
- Added database migration in `run_migrations()`

### Frontend
 - Updated `SpoolFormData` interface to include `core_weight_catalog_id: number | null`
 - Updated `defaultFormData` to initialize `core_weight_catalog_id: null`

- Modified `SpoolWeightPicker` component signature:
     - Added `catalogId` and `onCatalogIdChange` parameters
     - Removed local `selectedId` state (replaced with `catalogId` prop)
   - Updated matching logic:
     - Skips auto-matching if catalog hasn't loaded yet (prevents race conditions)
     - Only clears selection if no catalogId exists
     - Keeps valid selections even when multiple entries have same weight
   - Updated display logic to prioritize ID-based lookup over weight-based matching
   - Changed dropdown button click handler to store both weight AND catalog ID
   - Added `core_weight_catalog_id` to form initialization when editing spools
   - Updated form submission data object to include `core_weight_catalog_id`
   - Added `core_weight_catalog_id` to passed props in `SpoolWeightPicker` call
   - Added `core_weight_catalog_id: number | null` to `InventorySpool` interface

### Tests
- "preserves core_weight_catalog_id when editing other fields"
Verifies catalog ID persists when editing unrelated fields. Edit note → submit → assert `core_weight_catalog_id: 5` in PATCH.

- "includes core_weight_catalog_id when selecting from catalog"
[...]

PR #455, opened 2026-02-19
Allow external materials/brands to be used when adding spool

### LINKED ISSUE #419: [Feature]: Custom filaments color unavailable when adding spool, not all manufacturers appear in drop down menu
### Bug Description

Some manufacturers from the Color Catalog are not available for selection when adding a spool. Custom colors are also not present when adding a spool.

### Expected Behavior

I can select all materials, colors, and manufacturers synced from the external database when adding a spool. All of the above can be chosen in the dropdown menu.

### Steps to Reproduce

1. Sync with external database
2. Verify FormFutura is present in the dropdown menu
3. Add a new spool
4. Try to select FormFutura as the manufacturer

Also:
1. Add a new color to Settings->Filament
2. Try to recall that color via the Add Spool color search function

### Printer Model

H2C

### Bambuddy Version

0.2.0

### Printer Firmware Version

01.01.05.00

### Installation Method

Docker

### Operating System

Docker

```shell

```

### Screenshots

[bambuddy-support-20260217-184937.zip](https://github.com/user-attachments/files/25371401/bambuddy-support-20260217-184937.zip)

### Additional Context

I can work on this issue. 

### Checklist

- [ ] My printer is set to LAN Only mode

Keybored02:
## Description
This PR integrates the materials and brands fetched from the external database into the Add Spool form. It also allows for custom colors, brands and materials in Color Catalog to be included when adding a spool. It fixes a few minor UI inconsistencies and sizing issues. 

## Related Issue

Fixes #419

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made
- The Add spool modal now fetches from both the hard-coded list of materials/brands and the external database.
- A bilateral filter kicks in when the user selects a brand or material, filtering the other options based on what is available in the pair (custom fields are still allowed).
- Ui fixes:
  - Drop down menus are now unified in theme and scrollbar style
  - The add form for the Color Catalog has been reworked to enlarge the color selection button
  - When a numeric input field content is deleted it's no more populated with a 0 automatically
 

## Screenshots
Screenshot 2026-02-17 210909
Screenshot 2026-02-17 210924
Screenshot 2026-02-18 155312
Screenshot 2026-02-18 155320

## Testing

## Checklist


maziggy:
> Instead of filtering for weight class and fetching by name, empty spools are now referenced by a unique Id.

Can you please explain? This might have a high risk of breaking other functions.

Keybored02:
My bad, it's a duplicate of something already merged in  #414 . Let me edit the PR message.
To be clear, there are non changes relative to that in this PR.

Keybored02:
1. FilamentSection now allows custom materials via a button and onBlur commit
2. ColorSection filters color search within matchedCatalogColors
3. CatalogDisplayColor type is moved to types.ts
4. The ml-12 margin in ColorCatalogSettings is replaced with a gap utility for better layout
5. Cleaned up filteredCatalogColors dependencies
6. The catalog color section will now only appear if at least one brand or material is selected, or if there are matched catalog colors.

maziggy (changes requested):
A few things I noticed:

  **Bugs**

  1. Material dropdown onBlur doesn't commit typed values that match an existing list item

  The new material dropdown's onBlur handler only commits when the typed value is NOT in the filtered list:

  onBlur={() => {
    if (materialSearch && !filteredMaterials.includes(materialSearch)) {
      updateField('material', materialSearch);
      setMaterialDropdownOpen(false);
    }
  }}

  If a user types "PLA" (which IS in filteredMaterials) and clicks somewhere else on the form, the value is never committed — the dropdown closes via the outside-click
  handler and the input reverts to the previous value. The user would need to explicitly click the dropdown item.

  Also, filteredMaterials.includes() is case-sensitive, so typing "pla" when the list has "PLA" would commit "pla" as a custom entry rather than selecting the existing one.

  Suggested fix: commit the value on blur regardless — if it matches a list item, select it; if it doesn't, treat it as custom.

  2. .replace('brand', 'material') is dead code

  {t('inventory.useCustomBrand', { brand: materialSearch }).replace('brand', 'material')}

  The {{brand}} interpolation is already resolved by i18next before .replace() runs, so the output (e.g. Use "PLA") never contains the literal word "brand". The .replace() does nothing. Either add a dedicated useCustomMaterial i18n key or just drop the .replace() since the translated text is generic enough.

  3. CatalogDisplayColor type duplicated

  This type is defined in both types.ts (exported) and locally in ColorSection.tsx. They also differ slightly — material is string | null locally vs optional string in types.ts. Should import from one place.

  **UX suggestion**

  4. Brand-only selection shows all colors for that manufacturer

  When a brand is selected but material is not, the new code path returns ALL catalog colors for that manufacturer across all materials. This could show many swatches
  spanning PLA, PETG, ABS, etc. with no grouping. The tooltip already shows manufacturer — consider showing the material too (e.g. "White (Bambu Lab — PLA Basic)") so
  users can tell which material a swatch belongs to.

  **Minor**

  - w-20\n                   h-10 in ColorCatalogSettings.tsx has a line break inside the className string — should be w-20 h-10 on one line.

Keybored02:
Ok, I've removed the onBlur commit, it was a mess. Hopefully this is the last.

PR #502, opened 2026-02-23
Add material mismatch and insufficient filament checks

Keybored02:
## Description

Add material mismatch check when assigning non-BL spools. The warning can be dismissed. Addedchecks prior to printing for insufficient filament on the spool. This is a re-merge of #361 for branch 0.2.1b4 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made
- Check spoolmaterial against traymaterial (only types, no colors)
- Ask user for confirmation in case of mismatch with confirm modal
- Added checks for insufficient material to PrintModal
- Added toggle in Setting, Filament to disable insufficiency warning
- Updated locales

## Screenshots
image
image

## Testing

## Checklist

### DIFF
--- a/frontend/src/pages/PrintersPage.tsx
+++ b/frontend/src/pages/PrintersPage.tsx
@@ -2974,7 +2974,7 @@ function PrinterCard({
                                               trayInfo: {
                                                 type: filamentData.profile,
                                                 material: tray?.tray_type ?? undefined,
-                                                color: filamentData.colorHex ?? undefined,
+                                                color: filamentData.colorHex ?? '',
                                                 location: `${getAmsLabel(ams.id, ams.tray.length)} Slot ${slotIdx + 1}`,
                                               },
                                             }) : undefined,
@@ -3210,7 +3210,7 @@ function PrinterCard({
                                           trayInfo: {
                                             type: filamentData.profile,
                                             material: tray?.tray_type ?? undefined,
-                                            color: filamentData.colorHex ?? undefined,
+                                            color: filamentData.colorHex ?? '',
                                             location: getAmsLabel(ams.id, ams.tray.length),
                                           },
                                         }) : undefined,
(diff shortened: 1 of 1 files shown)


maziggy:
Please also rebase to resolve conflict.

Keybored02:
Regarding the material matching: yes that's an issue. Especially because of how the filament catalog is structured, with relevant info divided between material itself and subtypes. So Silk PLA will match with PLA, and brand names for base polymers like ApolloX will not match ASA even tough they are ASA. The other idea I had was to check for the filament profile assigned when adding the spool and compare it to the tray profile. But this would require a really extensive profiling effort on the user to work reliably. Maybe it can be a fallback? check for types matching first, if false check for profiles?

Keybored02:
Tested a few options, and came to the conclusion that there's no easy way to make the matching work 100% of the time. For example, the matcher would need to know that `selectedSpool?.material` ePLA or rPLA can be matched to `trayInfo.type` PLA, but PLA-CF cannot. Cleanup of symbols and abbreviations (-, + ,CF ,GF, Plus, R) before matching would break stuff, including `subtypes` in the query would break stuff, and doing neither definitely breaks stuff. The safest way would be to have a table with all correct assignments, but writing and maintaining it would be nightmarish. 

So, the middle ground I've chosen is: match full word first, `trayInfo.type` against `selectedSpool?.material`. If it's a perfect match, check the `trayInfo.profile` and `selectedSpool?.profile`. If they do match, no prompt to confirm. If `type` cannot be matched to anything, or it matches partially with some char excluded, or the profiles do not match, prompt to confirm. As long as the user gives accurate info on the spool, should work for 99% of the cases. The remaining 1% are edge cases where the subtype is key to identify the material, and the profile chosen does not match the actual filament properties.

The issue here is that I cannot get the `trayInfo.profile` and `selectedSpool?.profile` to get passed to AssignSpoolModal reliably. I think this has something to do with #524 . Will continue to check.

Keybored02:
Closed, will reopen after a rebase.

PR #531, opened 2026-02-26
Add customizable low stock threshold, add low stock filter

Keybored02:
## Description

Make the low spool threshold customizable instead of hard-coded to 20%. Add a new filter to the Filament page to show only low spools. 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Added UI and logic to let users edit and persist the "low stock" threshold (default 20%) via localStorage
- Introduced a "Low Stock" filter button in the inventory page

## Screenshots
image
image

## Testing

## Checklist

### DIFF
--- a/frontend/src/__tests__/pages/InventoryPageLowStock.test.tsx
+++ b/frontend/src/__tests__/pages/InventoryPageLowStock.test.tsx
@@ -317,7 +317,7 @@ describe('InventoryPage - Low Stock Threshold', () => {
 
       // Too high
       await user.clear(input);
-      await user.type(input, '150');
+      await user.type(input, '0');
 
       const saveButton = screen.getByRole('button', { name: /save/i });
       await user.click(saveButton);


maziggy (changes requested):
1. Dead CSS (no effect)
  The <style> block targeting input[type=number] spinners has no effect since the input is type="text" with inputMode="decimal". Same for the MozAppearance: 'textfield'
  inline style. Both can be removed.

  2. Update the UsageFilter type
  'lowstock' is added as an inline union on the state (UsageFilter | 'lowstock'), but the UsageFilter type definition on line 23 isn't updated. Please add 'lowstock' to the type itself so it stays in sync: type UsageFilter = 'all' | 'used' | 'new' | 'lowstock';

  3. Wrong tooltip on edit button
  The pencil icon uses title={t('inventory.editSpool')} ("Edit Spool") — should be something like "Edit threshold" instead. Needs a new i18n key or a more fitting
  existing one.

  Minor nits (optional):
  - The || 'Threshold must be between 0.1 and 99.9' fallback on the toast isn't needed since the key exists in all locale files
  - localStorage.setItem on save isn't wrapped in try-catch (the read is) — could throw in private browsing

Keybored02:
I think localStorage is fine for this setting.

maziggy:
I changed my mind, sorry!

Think we should not use localStorage to keep the unique approach and to not limit it to a single browser.

Keybored02:
Ok, will take bit to reimplement and test. Brb.

maziggy (changes requested):
**Issues**

  **1. No backend validation for threshold range ⚠️**

  backend/app/schemas/settings.py — Both AppSettings and AppSettingsUpdate accept any float value. The frontend validates 0.1–99.9, but the API is unprotected — a direct
  API call with -5 or 9999 would be accepted and break the filter.

  Fix:
  # AppSettings
  low_stock_threshold: float = Field(
      default=20.0, ge=0.1, le=99.9,
      description="Low stock threshold percentage (%) for inventory filtering and display"
  )

  # AppSettingsUpdate
  low_stock_threshold: float | None = Field(default=None, ge=0.1, le=99.9)

  **2. Validation test has no assertions**

  InventoryPageLowStock.test.tsx:367-391 — The "validates threshold input range" test types 150, clicks Save, then has only comments:

  // Should show error (implementation depends on toast system)
  // The value should not be saved

  This passes regardless of whether validation works. It should assert that the PUT endpoint was NOT called (e.g., capture updatedSettings like the test above it does,
  and verify it remains null).

  **3. thresholdInput state initialization**

  InventoryPage.tsx:560 — thresholdInput initializes from settings?.low_stock_threshold ?? 20 at mount time, before settings have loaded. The edit button's onClick re-syncs it which mostly covers this, but adding a useEffect would be more robust:

  useEffect(() => {
    if (!showThresholdInput) {
      setThresholdInput(lowStockThreshold.toString());
    }
  }, [lowStockThreshold, showThresholdInput]);

  **4. Field ordering in AppSettingsUpdate (minor)**

  The new field is inserted between mqtt_password and mqtt_topic_prefix, splitting the MQTT group. Should be appended at the end to match how AppSettings places it.

  **5. Orphaned i18n key (minor)**

  The old inventory.lowStockThreshold key (e.g. '<20% remaining' in en) is now unused across all 6 locale files — replaced by the dynamic < {lowStockThreshold}% rendering. Should be removed.

PR #617, opened 2026-03-05
Add camera grid streaming with multiplexed MJPEG, print controls, and UX improvements

aneopsy:
## Description

Adds a real-time camera grid view to the Printers page — a single multiplexed MJPEG stream serves frames for all visible printers, decoded in a Web Worker and rendered on canvas. Includes inline print controls (pause/resume/stop), HMS error notifications, quality switching, and comprehensive backend stream management with shared producers and automatic cleanup.

Frontend:

image
image

Backend optimisation - CPU usage for camera stream:

Before:
image

After:
image

## Type of Change

## Changes Made

**Backend (`backend/app/api/routes/camera.py`)**
- Shared stream hub (`_StreamHub`) — single producer per camera, multiple consumers
- Multiplexed `/camera/grid-stream` endpoint with binary-framed multi-printer MJPEG
- Print control endpoints (pause, resume, stop) on camera cards
- Periodic cleanup of stale frame buffers and orphaned ffmpeg processes
- RTSP semaphore to limit concurrent ffmpeg spawns
- Race condition fix in `_ensure_producer` for concurrent client requests

**Frontend (`frontend/src/pages/PrintersPage.tsx`)**
- `CameraGrid` component with canvas-based rendering and dynamic layout
- Web Worker (`cameraGridDecoder.worker.ts`) for off-main-thread frame decoding
- `IntersectionObserver` to pause/resume streams for off-screen cards
- Reconnect logic with exponential backoff
- Quality/scale/FPS switching controls
- Inline pause, resume, stop print controls on camera cards
- HMS error notification badges with modal details
- ETA tooltip on progress display

**i18n**
- Added camera grid translation keys to all 7 locale files
- Italian nozzle info keys

## Testing

- [x] Backend: 1794/1795 tests pass (1 flaky unrelated FTP test)
- [x] Frontend: 1038/1038 tests pass
- [x] Ruff lint/format clean
- [x] TypeScript + ESLint clean

## Checklist


maziggy (changes requested):
There are some files which should not be there, because they are overwriting my pipeline!!!

aneopsy:
I'm trying to fix the camera feed in general

maziggy:
Huh? You PR would overwrite my .gitignore and .pre-commit-config.yaml. Furthermore there are old index files included.

aneopsy:
Got it, I rewrote the branch history to remove the changes from these files.

maziggy (changes requested):
**Bugs:**

  1. Progress sort is broken for idle vs offline — Infinity - 1 === Infinity and Infinity - 2 === Infinity in IEEE 754, so idle and offline printers sort randomly relative to each other. Use large finite numbers instead (e.g. 1e9 for offline, 1e9 - 1 for idle).
  2. reconnectAttemptsRef.current read during render won't update the UI — Refs don't trigger re-renders, so reconnectAttempt passed to CameraGridCard will always show
  the value from the last render (likely 0). Need to track this in state or derive from existing state.
  3. CancelledError / GeneratorExit not re-raised in generate_chamber_mjpeg_stream and generate_rtsp_mjpeg_stream — these are BaseExceptions that should be re-raised
  after logging so task cancellation propagates. The finally block runs either way. This matters because SharedStreamHub.stop() cancels producer tasks and waits for them
  — if cancellation isn't acknowledged it may hang on the 5s timeout.

  **Minor / cleanup:**

  4. setCacheTick(t => t + 1) and setStatusTick(t => t + 1) — the t parameter shadows useTranslation()'s t. Not a bug currently but a hazard. Use prev => prev + 1.
  5. f-string logger calls in external_camera.py (_capture_rtsp_frame) — rest of codebase uses %s-style. Would fail Ruff G004 if enabled.
  6. generate_chamber_image_stream return type annotation says asyncio.StreamReader | None but actually returns tuple[asyncio.StreamReader, asyncio.StreamWriter] | None.
  7. No max jpegLen sanity check in the binary frame protocol — a corrupt frame header with a huge length would cause the client buffer to grow unboundedly. Consider
  capping at e.g. 10MB.

maziggy (changes requested):
A couple of minor things that could be cleaned up please.

  1. _ensure_producer and camera_grid_stream reach into _hub._lock / _hub._streams directly rather than going through a public SharedStreamHub method — works fine, just
  an encapsulation thing.
  2. CameraGrid keeps both reconnectAttemptsRef (for logic) and reconnectAttempt state (for UI) — works correctly, just a bit unusual to maintain two sources.

aneopsy:
Fixed an additional security issue by adding camera stream view permissions; also added unit tests

maziggy (changes requested):
Sorry, but the more often I look at the code, the more bugs I find.

Not sure if there are also possible "old" bugs.


  **1. Global tray ID calculation (PrintersPage.tsx ~line 3052)**

  The regular AMS section uses ams.id * 4 + slotIdx directly, but the HT AMS section (line ~3272) uses getGlobalTrayId(). Should use getGlobalTrayId() in both places for
  consistency — the raw formula would break for AMS IDs ≥128 or if tray.id doesn't match the array index.

  **2. Untracked setTimeout in onMutate (PrintersPage.tsx ~line 1976)**

  The 2-second minTimePassedRef timer handle is discarded. If another re-read triggers within 2s, a second untracked timer fires and can clear the spinner prematurely.
  Save the handle in a ref and clear it on re-trigger/unmount, same as refreshTimeoutRef.

  **3. localStorage parse without try/catch (PrintersPage.tsx ~line 3837)**

  JSON.parse(saved) for cameraWindowState can throw on corrupted data. EmbeddedCameraViewer already wraps the same pattern in try/catch — this should too.

  **4. closePlateCheckModal state ordering (PrintersPage.tsx ~line 2050)**

  setPlateCheckResult(null) closes the modal before await setChamberLight(false) resolves. If the user reopens plate-check during that async window, the light state gets
  out of sync and the chamber light stays on. Move setPlateCheckResult(null) after the await.

  **5. Imports inside streaming loops (camera.py)**

  import time and import uuid appear inside while True frame loops (~lines 121, 244, 305, 391, 462). Python caches them, but it's still a sys.modules lookup per frame.
  Move to top-level imports.

  **6. Chamber stream disconnect cleanup (camera.py ~line 527)**

  The finally block uses asyncio.sleep(0.1) to let the inner generator clean up, but if the chamber stream is blocked on reader.readexactly() with a 30s timeout, the
  disconnect event won't fire until the next I/O completes. Closing writer doesn't necessarily cause the SSL reader to raise immediately. Consider explicitly calling
  aclose() on the inner generator or closing the reader's transport as well.

maziggy:
ject literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(497,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(498,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(499,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(500,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(501,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(502,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(503,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(504,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(505,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(521,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(522,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(523,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(524,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(525,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(526,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(527,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(528,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(529,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(530,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(545,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(548,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(549,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(550,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.31 src/i18n/locales/ja.ts(551,5): error TS1117: An object literal cannot have multiple properties with the same name.
#16 25.67 npm notice
#16 25.67 npm notice New major version of npm available! 10.9.4 -> 11.11.0
#16 25.67 npm notice Changelog: https://github.com/npm/cli/releases/tag/v11.11.0
#16 25.67 npm notice To update run: npm install -g npm@11.11.0
#16 25.67 npm notice
#16 ERROR: process "/bin/sh -c npm run build" did not complete successfully: exit code: 2

#19 [stage-1 6/9] RUN --mount=type=cache,target=/root/.cache/pip     pip install --root-user-action=ignore -r requirements.txt
#19 10.18   Installing build dependencies: finished with status 'done'
#19 10.18   Getting requirements to build wheel: started
#19 CANCELED
------
 > [frontend-builder 6/6] RUN npm run build:
25.31 src/i18n/locales/ja.ts(545,5): error TS1117: An object literal cannot have multiple properties with the same name.
25.31 src/i18n/locales/ja.ts(548,5): error TS1117: An object literal cannot have multiple properties with the same name.
25.31 src/i18n/locales/ja.ts(549,5): error TS1117: An object literal cannot have multiple properties with the same name.
25.31 src/i18n/locales/ja.ts(550,5): error TS1117: An object literal cannot have multiple properties with the same name.
25.31 src/i18n/locales/ja.ts(551,5): error TS1117: An object literal cannot have multiple properties with the same name.
25.67 npm notice

maziggy:
Please always run the test locally before pushing.

maziggy (changes requested):
**Critical**

  **1. stop_camera_stream kills hub-managed ffmpeg processes**

  The stop handler's docstring says "The shared producer (SharedStreamHub) is NOT stopped here" and it intentionally doesn't call _hub.stop(). However, the handler still
  iterates _active_streams and terminates every ffmpeg process matching {printer_id}-*. Hub-managed producers ARE registered in _active_streams (via _ensure_producer →
  generate_rtsp_mjpeg_stream → _active_streams[stream_id] = process).

  When a single-camera viewer closes and the frontend calls POST /{id}/camera/stop via sendBeacon, the hub producer's ffmpeg gets killed — disconnecting all other viewers
   (grid on another tab, embedded viewer on another device, etc.). This contradicts the stated multi-viewer design.

  Fix: skip entries that belong to active hub producers in the stop handler (e.g., check _hub.is_active(printer_id) before terminating), or don't register hub producer
  processes in _active_streams at all.

  **2. Print control buttons ignore printers:control permission**

  CameraGrid passes onPause/onStop/onResume unconditionally to CameraGridCard. The existing PrinterCard gates these behind hasPermission('printers:control'). Read-only
  users see clickable pause/stop/resume buttons that produce 403 error toasts.

  onClearPlate right below shows the correct pattern:
  onClearPlate={hasPermission('printers:clear_plate') ? () => clearPlateMutation.mutate(p.id) : undefined}
  Same pattern should be applied to all three print-control callbacks.

  **3. xxl: breakpoint classes don't exist in Tailwind**

  GRID_LAYOUT_COLS uses xxl:grid-cols-* but the Tailwind config has no custom xxl breakpoint — only the defaults (sm, md, lg, xl, 2xl). The xxl: classes are never
  generated, so the grid caps at the xl: column count on wide screens (compact stops at 5 columns instead of 6).

  Fix: replace xxl: with 2xl: throughout.

  **Important**

  **4. RTSP semaphore held for entire stream lifetime**

  generate_rtsp_mjpeg_stream acquires the semaphore (_MAX_RTSP_FFMPEG = 20) and holds it for the entire ffmpeg process lifetime — the read loop runs inside async with
  semaphore:. After 20 simultaneous RTSP streams, any new stream (including a hub producer restart after a network hiccup) blocks at semaphore acquisition indefinitely.

  The semaphore appears intended to limit concurrent ffmpeg launches, not total concurrent streams. Fix: narrow the async with semaphore: scope to cover only the
  create_subprocess_exec call and the initial failure check, then release before entering the read loop.

  **5. hasQueuedJobs always false in camera grid**

  CameraGrid reads queryClient.getQueryData(['queue', p.id, 'pending']) but never fetches this query. The cache is only populated when PrinterQueueWidget renders (in
  normal card view). In camera grid mode, getQueryData returns undefined, so the "Clear Plate" button never appears even when jobs are queued. Either add a query fetch
  inside CameraGrid, or remove the dead prop if intentionally omitted.

  **6. ja.ts introduces duplicate keys within the printers object**

  The insertion block at @@ -486 adds keys that already exist earlier in the same printers object:
  - deleteConfirm (line 146), camera (line 216), skipObject (line 217)
  - openCameraOverlay (line 410), openCameraWindow (line 411), clickToViewHmsErrors (line 445)

  JavaScript silently takes the last definition. Values are identical now but will cause confusion if either copy is ever updated independently. The ~46 genuinely new
  backfilled keys in the block are fine — just the duplicates need removing.

  **7. zh-CN missing printers.status.failed and printers.sort.progress**

  The PR adds these keys to all 6 other locales but not zh-CN.

  **Minor**

  **8. snapshotBlobUrl not cleared on mode switch in CameraPage.tsx**

  When switching from snapshot to stream and back, the old blob URL is revoked but snapshotBlobUrl state isn't set to null. On switch-back, the  briefly renders with
   the revoked URL causing a broken-image flash. Adding setSnapshotBlobUrl(null) in switchToMode('snapshot') fixes it.

maziggy (changes requested):
**Build breaker:**

  1. ja.ts duplicate keys — The printers object still has ~31 duplicate keys (lines 491–552 repeat keys already defined at lines 147–447). This causes the TypeScript build to fail. Remove the duplicate block and keep only the genuinely new keys (cameraGridError, cameraGrid.*, viewHmsErrors, etc.).

  **Still open from latest review:**

  2. Print control buttons ignore printers:control permission — onPause, onStop, onResume are still passed unconditionally to CameraGridCard (lines 2037–2039).
  onClearPlate on line 2047 shows the correct pattern:
  onPause={hasPermission('printers:control') ? () => setConfirmAction(...) : undefined}
  3. xxl: breakpoint classes don't exist — GRID_LAYOUT_COLS (lines 1568–1570) still uses xxl:grid-cols-*. Tailwind only has 2xl: — replace xxl: with 2xl:.
  4. hasQueuedJobs always false in camera grid — Line 2057 reads from queryClient.getQueryData(['queue', p.id, 'pending']) but this query is only fetched when PrinterCard renders (normal card view). In grid mode, the cache is never populated. Either prefetch the query in CameraGrid or remove the prop.
  5. zh-CN missing printers.status.failed and printers.sort.progress — These were added to all other locales but not zh-CN.
  6. snapshotBlobUrl not cleared on mode switch — switchToMode('snapshot') in CameraPage.tsx (line 204) doesn't call setSnapshotBlobUrl(null), causing a brief
  broken-image flash when switching back from stream mode.
  7. Hub producer silently swallows CancelledError — The except asyncio.CancelledError at line 356 in camera.py logs but doesn't re-raise. The chamber and RTSP stream
  handlers correctly re-raise — this one should too for consistent cancellation propagation.

  Please run cd frontend && npm run build locally before pushing — that would have caught #1 immediately.

maziggy:
Each review round takes significant time on my end, so please make sure all feedback items are addressed and the build passes locally 
 
 cd frontend  && npm run build && cd ..
 ./test_frontend.sh
 ./test_backend.sh
 ./test_security --full
 
 before pushing. I can't keep re-reviewing the same issues.

aneopsy:
Sorry about that, I missed your comments, I was just pushing a last fix for the security/auth on the camera and WebSocket so I can test it directly on the prod server. Also, the i18n issue was due to a rebase error, my bad.

maziggy:
Please rebase with branch 0.2.3b1 and also change target branch to 0.2.3b1.

maziggy (changes requested):
**Bugs (will crash/malfunction in normal use)**

  1. useZoomPan.ts — setUlPanOffset is undefined
  handleZoomOut calls setUlPanOffset({ x: 0, y: 0 }) but the state setter is setPanOffset. This will throw TypeError every time a user zooms out to level 1.
  2. CameraPage.tsx — stalledPaused typo useStreamReconnect accepts stallPaused but the call passes stalledPaused (extra d). The prop is silently ignored, so stall detection runs during snapshot mode and transitions, triggering spurious reconnects.
  3. useMjpegStream.ts — clean EOF treated as error 
When the stream ends normally (done === true from the reader), the code calls onErrorRef.current?.() which triggers reconnection. A clean server-side close shouldn't
  be treated as an error — this will cause reconnect loops whenever the backend intentionally ends the stream.

  **Security**

  4. external_camera.py — RTSP URLs not sanitized before passing to ffmpeg
  _capture_rtsp_frame() and _stream_rtsp() pass the raw url directly as ffmpeg's -i argument. The MJPEG path correctly calls _sanitize_camera_url() and the USB path uses strict regex validation, but the RTSP path has neither. Since external_camera_url is user-configurable, someone could store a file:// or concat: URL to read arbitrary
   local files via the MJPEG stream output. Fix: apply _sanitize_camera_url(url, allowed_schemes=("rtsp", "rtsps")) at the top of both functions.
  5. external_camera.py — _stream_usb uses weak startswith check   _capture_usb_frame correctly validates with regex.match(r"^/dev/video(\d{1,2})$") and reconstructs the path, but _stream_usb only checks startswith("/dev/video"),
  allowing paths like /dev/video../../etc/passwd. Should use the same regex validation.

  **Resource leaks / race conditions**

  6. PrintersPage.tsx (CameraGrid) — stall timer leaks across reconnect stallTimer is declared inside try, so when the catch block starts a reconnect delay, the old timer is still pending. When it fires, it cancels the new connection's readerRef. Declare stallTimer in the outer loop scope and clear it in the catch before the reconnect wait.
  7. PrintersPage.tsx (CameraGrid) — transferred ArrayBuffer may be accessed after detach buf.buffer.slice(jpegStart, jpegStart + jpegLen) gets a slice of the underlying buffer, then postMessage(..., [jpeg]) transfers it. If buf hasn't been reallocated, the
   remaining buf.buffer is now detached. Use new Uint8Array(buf.buffer, jpegStart, jpegLen).slice().buffer to always produce a fresh copy before transfer.
  8. external_camera.py — ffmpeg zombie on RTSP timeout _capture_rtsp_frame calls process.kill() in the TimeoutError handler but never await process.wait(), leaving a zombie process.
  9. camera.py — generate_chamber_image_stream return type annotation
  Annotated as -> asyncio.StreamReader | None but actually returns (reader, writer) tuple. The writer reference is important for cleanup.
  10. camera.py — orphan stream cleanup incomplete cleanup_orphaned_streams pops from _active_streams but doesn't clear _last_frames / _last_frame_times for the same printer, causing stale values in camera_status.

  **Minor**

  - EmbeddedCameraViewer.tsx: resetZoom missing from fullscreen useEffect dependency array
  - cameraGridDecoder.worker.ts: self as unknown as {...} cast — consider using declare const self: DedicatedWorkerGlobalScope for proper typing

maziggy:
How do we want to proceed with this one?

aneopsy:
> How do we want to proceed with this one?

I will address the reported issues. I am currently testing the feature across 30 printers and encountering network/printer instability that corrupts the video feed and crashes the FFMPEG process. I am continuously improving the system to ensure reliable 24/7 video stream handling.

PR #625, opened 2026-03-05
Feature: Force Color Match

### LINKED ISSUE #604: [Feature]: Force Color Match
### Problem or Use Case

When sending a print with a filament color that is not loaded into any AMS, it would be required in some cases to default to "Force color match" and have the job held in the queue until a printer with the matching color and material. 
Currently the person printing can be surprised by the actual color printed without any indication it will change.

### Proposed Solution

Add a checkbox to the filament mapping options, "Force color match". This would send the job to the queue and mark the job as "No Matching Material/Color" until loaded. I think this should be set to true by default. Must also work for dual nozzle prints.

### Alternatives Considered

_No response_

### Feature Category

Print Queue & Scheduling

### Priority

Critical for my use case

### Mockups or Examples

_No response_

### Contribution

### Checklist

cadtoolbox:
## Description

This PR introduces a new **Force Color Match** feature to improve how Bambuddy matches filament/color information during print scheduling when selecting "Any Model" as the print method. When enabled, the system will **prefer a strict color match** (rather than “best effort” / fallback matching) to reduce incorrect filament assignments and improve archive consistency—especially for libraries with similarly-named filaments or overlapping color metadata.

Bambuddy currently performs color/filament matching in a way that can occasionally:
- Choose an unintended filament when multiple candidates are “close enough”
- Fall back to a near match when an exact match exists but is not selected due to ordering/heuristics
- Produce inconsistent assignments between imports depending on metadata completeness

**Force Color Match** adds an explicit option to make matching behavior deterministic and stricter where users want accuracy over flexibility.

## Related Issue

Fixes #604 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Adds a **Force Color Match** option in the UI where color/filament matching is configured/triggered.
- When enabled, matching logic will:
  - Prefer **exact (or explicitly-defined) color matches** first
  - Avoid fallback candidates unless explicitly allowed/necessary
  - Provide clearer outcomes when a strict match cannot be found (e.g., prompt / warning / “unmatched” state)

## Testing

- [X] I have tested with my printer model: H2D Pro, X1E

Testing involved:
- Adding a print job with matching material/color as normal.
- Adding a print job with no matching material/color.
- Adding a print job with matching material/color after a no match job. Job jumps queue.
- Adding filament to the AMS of desired color while a no matching material/color job is waiting.

## Checklist

## Additional Notes

Additional testing with different printers, AMS combos and filament would be ideal.


cadtoolbox:
@maziggy I'm not quite sure I have enough context to understand comment 3 or 4. Can you elaborate?

maziggy:
**Comment 3 — PrintersPage.tsx queueCount filter:**

  The badge count that shows how many queued jobs are compatible with each printer (around line ~1792 in PrintersPage.tsx) still uses the old some() logic that treats all filament overrides the same. It doesn't distinguish between force-color overrides and preference-only overrides like PrinterQueueWidget.tsx now does. This means the
  badge could show a job as compatible with a printer even if that printer doesn't have the exact color loaded for a force-color slot — defeating the purpose of the feature.

  The fix would be to split the overrides by force_color_match in the queueCount filter too, so force-color slots require an exact match while preference slots keep the current "close enough" behavior.

  **Comment 4 — loadedFilaments?.size guard:**

  In your PR you fixed one place where loadedFilamentTypes?.size was used as a truthy check — the problem being that an empty Set has size === 0 which is falsy, so the check would be skipped entirely for a printer with no filaments loaded (making it appear compatible with everything). That fix was correct.

  But the same bug exists at line ~1792 with loadedFilaments?.size — it's the same pattern, just a different variable. A printer with 0 filaments loaded would skip the color compatibility check and appear compatible with all jobs. Should be changed to loadedFilaments !== undefined (or similar) to match your other fix.
  
 Does that help?

cadtoolbox:
@maziggy I "think" I have all your comments addressed.  The Commit 940810c incorporation from your recent changes got a little confusing. Let me know if you find anything else that needs to be updated.

cadtoolbox:
@maziggy I just reviewed the files changed again on this PR. I swear I didn't change half of those; I rebased with 0.2.2b2. 🙄

maziggy:
Please rebase with branch 0.2.3b1 and also change target branch to 0.2.3b1.

cadtoolbox:
@maziggy I think it all looks correct now.

maziggy:
The PR is still targeting the worng branch.

Please rebase with branch 0.2.3b1 and also change target branch to 0.2.3b1.

maziggy:
> [frontend-builder 6/6] RUN npm run build:
3.728 src/components/PrintModal/FilamentOverride.tsx(166,13): error TS1128: Declaration or statement expected.
3.728 src/components/PrintModal/FilamentOverride.tsx(168,9): error TS1381: Unexpected token. Did you mean `{'}'}` or `&rbrace;`?
3.728 src/components/PrintModal/FilamentOverride.tsx(168,11): error TS1381: Unexpected token. Did you mean `{'}'}` or `&rbrace;`?
3.728 src/components/PrintModal/FilamentOverride.tsx(170,5): error TS1005: ')' expected.
3.728 src/components/PrintModal/FilamentOverride.tsx(171,3): error TS1109: Expression expected.

cadtoolbox:
@maziggy with this branch to merge I don't see your changes in the filament selection for any model to show the matte or basis and now the settings page labels aren't reading correctly. Is this an issue with my branch or 0.2.3b1?

Sorry, just saw your comment. I'll work on fixing these issues.

maziggy:
Cherry-picked 940810cba3203d1450c83553bb4cac1167df68a4 into branch 0.2.3b1.

cadtoolbox:
@maziggy Last comments have been incorporated. Thanks for your help on seeing this one to the finish line.

Issue #639, opened 2026-03-06
[Bug]: Spoolman Not showing link spool with spoolman configured


errolsancaktar:
### Bug Description

Have everything set up and can see the comms between bambuddy and spoolman.  spoolman has spools available

### Expected Behavior

Should let me link spools

### Steps to Reproduce

set up per docs

### Printer Model

A1

### Bambuddy Version

v0.2.1.1

### Printer Firmware Version

1.07.02.00

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

### Relevant Logs / Support Package

```shell

```

### Screenshots

<img width="825" height="747" alt="Image" src="https://github.com/user-attachments/assets/ebf3f06e-fc91-4c69-b844-c91aea81e38b" />
<img width="1417" height="321" alt="Image" src="https://github.com/user-attachments/assets/10c42813-00fd-4ea0-9a03-de807a93d5c6" />
<img width="460" height="1008" alt="Image" src="https://github.com/user-attachments/assets/218bea57-aba8-45ba-9a17-82a0b266eb90" />

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I am using the latest version of Bambuddy
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled

maziggy:
What spools are you talking about? Need some more details.

errolsancaktar:
maybe i'm misunderstanding the process but heres my scenario:

- i dont use bambu or any rfid spools right now
- i store my spools in spoolman
- id like to be able to load a spool into my a1 and have it record usage against it in spoolman

based on what i'm seein in the code however, it seems that spool tracking in bambuddy bases the spool on the tag_uuid of the loaded filament.  

Some tests i've run through:
- the interaction between bambuddy and spoolman seems fine, i can pull filaments, spools, etc through the api of bambuddy
- everything is conifugred and bambuddy accepts that spoolman is there
- if i set a dummy 32 bit hex val in tag i can force "link" the spool to a tray in bb but i still dont see any spoolman specific stuff
- tried running current beta with no change
- does this have to do with the A1 loading null for the tray_uuuid or is that how all bambus work?

maziggy:
As per documentation, this is only available for the internal inventory.

There's already a PR to add this also for Spoolman, but I didn't check it so far.

PR #653, opened 2026-03-07
[Fix]: Non-Bambu Lab Spools can now be fully linked/unlinked to Spoolman

shrunbr:
## Description

**Third PR Update:** Sorry for the confusion, I didn't see the weird rebase that happened. Should be all resolved now.

This fixes the issue where the "Link to Spoolman" button was not showing for non-Bambu Lab spools. Generic tags are generated for each non-Bambu spool using Printer ID + AMS ID + Tray ID. This tag is then used in the extra.tag field within Spoolman. 

In addition to fixing the spoolman linking issue, I've also added an Unlink from Spoolman button to non-Bambu spools which are linked so that the tag field can be cleared right from the interface rather than going into Spoolman to do it manually.

**COPILOT WRITTEN SUMMARY BELOW**

This pull request refactors and expands the Spoolman spool linking functionality to support both tray UUIDs and generic tag UIDs, improves validation and error handling, and adds support for unlinking spools. It also updates frontend components and tests to match these backend changes, ensuring a consistent user experience and robust integration.

## Related Issue

#639 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [X] Test addition or update

## Changes Made

**FRONTEND CHANGES**

- LinkSpoolModal now uses the Spoolman API and not the Inventory API
- linkTagToSpool in LinkSpoolModal has been simplified to linkSpool(spoolId, spoolTag)
- Added "Unlink from Spoolman" button below the "Open in Spoolman" button only on non-Bambu filaments
- Updated FilamentHoverCard to show the Spoolman link button even when trayUuid is missing using failback tag generation
- Added unlinkSpoolMutation in PrintersPage with cache invalidation for linked/unlinked spool maps
- Implemented getFallbackSpoolTag() in PrintersPage to generate the 16-char ID for slots without RFID identifiers
- Added onUnlinkSpool callbacks for standard AMS, AMS-HT and external tray slots
- Updated linkSpool in client.ts to accept spoolTag parameter instead of object request
- Added unlinkSpool(spoolId) method in client.ts
- Updated all test expectations in LinkSpoolModal.test.tsx to use new Spoolman API methods, data structure and "Select Spool" UI text

**BACKEND CHANGES**

- Updated LinkSpoolRequest in spoolman.py to accept spool_tag, tray_uuid or tag_uid
- Modified link validation in spoolman.py to accept both 16 and 32-char hex tags
- Added POST /spoolman/spools/{spool_id}/unlink endpoint in spoolman.py to clear extra.tag field
- Updated link validation test expectation to "16 or 32 hex characters" in test_spoolman_api.py
- Added unlink success test validating tag is cleared to empty JSON string in test_spoolman_api.py

**TRANSLATIONS**

- Added translations for "noTrayUuid" key for missing tag scenarios

## Screenshots

link-to-spoolman-button

link-modal

open-spoolman-unlink

## Testing

- [X] I have tested with my printer model: P2S

## Checklist

## Additional Notes

N/A


maziggy (changes requested):
**A few things I'd like to address before merging:**

  **Design question — fallback tag stability:**
  The fallback tag uses the printer's database ID (getFallbackSpoolTag(printerId, amsId, trayId)). If a user removes and re-adds a printer, it gets a new ID, which would
  silently break all existing fallback tag links. Would using the printer's serial number (or a hash of it) be more stable here?

  **Code issues to fix:**

  1. vendor field in FilamentHoverCard — The unlink button checks data.vendor !== 'Bambu Lab', but vendor isn't in the FilamentData interface. Can you confirm this
  compiles cleanly? If not, vendor needs to be added to the interface.
  2. import json inside the unlink function body — The existing link_spool had this too (pre-existing debt), but since json is already used at module level, please move
  the import to the top of the file.
  3. hasUnlinkedSpools prop cleanup — The Link button's disabled state now uses !spoolman.onLinkSpool instead of hasUnlinkedSpools, but the prop is still passed and
  declared in the interface. If it's no longer driving any UI behavior in FilamentHoverCard, it should be removed.
  4. noTrayUuid translation key — With the fallback tag generation, spoolTag in LinkSpoolModal should never be falsy (since getFallbackSpoolTag always returns a value). Is this a dead code path now? If so, it can be removed. If not, the message "No tray UUID - cannot link" doesn't match the new behavior.
  5. No confirmation on unlink — Unlinking is a one-click action with no confirmation. A simple confirm dialog would help prevent accidental unlinks.

shrunbr:
> **A few things I'd like to address before merging:**
> 
> **Design question — fallback tag stability:** The fallback tag uses the printer's database ID (getFallbackSpoolTag(printerId, amsId, trayId)). If a user removes and re-adds a printer, it gets a new ID, which would silently break all existing fallback tag links. Would using the printer's serial number (or a hash of it) be more stable here?
> 
> **Code issues to fix:**
> 
> 1. vendor field in FilamentHoverCard — The unlink button checks data.vendor !== 'Bambu Lab', but vendor isn't in the FilamentData interface. Can you confirm this  compiles cleanly? If not, vendor needs to be added to the interface.
> 2. import json inside the unlink function body — The existing link_spool had this too (pre-existing debt), but since json is already used at module level, please move the import to the top of the file.
> 3. hasUnlinkedSpools prop cleanup — The Link button's disabled state now uses !spoolman.onLinkSpool instead of hasUnlinkedSpools, but the prop is still passed and declared in the interface. If it's no longer driving any UI behavior in FilamentHoverCard, it should be removed.
> 4. noTrayUuid translation key — With the fallback tag generation, spoolTag in LinkSpoolModal should never be falsy (since getFallbackSpoolTag always returns a value). Is this a dead code path now? If so, it can be removed. If not, the message "No tray UUID - cannot link" doesn't match the new behavior.
> 5. No confirmation on unlink — Unlinking is a one-click action with no confirmation. A simple confirm dialog would help prevent accidental unlinks.

Good call on the printer ID vs Serial Number. I've gone ahead and updated that for example, here's what it looks like for me now.

image

As for the other stuff:
1. "Bambu Lab" is mentioned in the FilamentData interface it's either Bambu Lab or Generic for the vendor. You're using this exact same logic (or similar) also in FilamentHoverCard.tsx on line 196, 200, and 353.

2. Done, it's been moved.

3. Done, it's removed.

4. Good catch, this isn't needed anymore.

5. Great idea, done as well!

---

I've also tested these updates locally and everything seems to be working as expected for me.

maziggy (changes requested):
A few things remaining before this can merge:

  **Must fix:**

  1. TypeScript type error in LinkSpoolModal — spoolTag is string | undefined (from trayUuid || tagUid) but gets passed to api.linkSpool(spoolId, spoolTag) which expects
  string. The button is correctly disabled when !spoolTag, so it's safe at runtime, but this won't compile. Quick fix: api.linkSpool(spoolId, spoolTag!) or add a guard in the mutation function.
  2. Unlink toast uses wrong i18n keys — unlinkSpoolMutation.onError uses  ('spoolman.linkFailed') (link, not unlink), and onSuccess falls back to    t('spoolman.unlinkSpool') which is a button label, not a success message. Please add dedicated spoolman.unlinkSuccess and spoolman.unlinkFailed keys.
  3. Tag display vs. tag sent mismatch (Copilot also flagged this) — LinkSpoolModal computes spoolTag = trayUuid || tagUid but the UI elsewhere shows tagUid || trayUuid.
  If both exist, the displayed tag and the one sent to the backend are different. They should use the same priority order.

  **Nice to have (can be follow-up):**

  4. The Spoolman client init + health check boilerplate is now copy-pasted across link_spool, unlink_spool, and others. A small helper would clean that up, but fine to
  leave for later.
  5. LinkSpoolRequest accepts {} since all 3 fields are optional — a Pydantic @model_validator requiring at least one would be more robust, but the route-level validation covers it.

shrunbr:
> A few things remaining before this can merge:
> 
> **Must fix:**
> 
> 1. TypeScript type error in LinkSpoolModal — spoolTag is string | undefined (from trayUuid || tagUid) but gets passed to api.linkSpool(spoolId, spoolTag) which expects
>    string. The button is correctly disabled when !spoolTag, so it's safe at runtime, but this won't compile. Quick fix: api.linkSpool(spoolId, spoolTag!) or add a guard in the mutation function.
> 2. Unlink toast uses wrong i18n keys — unlinkSpoolMutation.onError uses  ('spoolman.linkFailed') (link, not unlink), and onSuccess falls back to    t('spoolman.unlinkSpool') which is a button label, not a success message. Please add dedicated spoolman.unlinkSuccess and spoolman.unlinkFailed keys.
> 3. Tag display vs. tag sent mismatch (Copilot also flagged this) — LinkSpoolModal computes spoolTag = trayUuid || tagUid but the UI elsewhere shows tagUid || trayUuid.
>    If both exist, the displayed tag and the one sent to the backend are different. They should use the same priority order.
> 
> **Nice to have (can be follow-up):**
> 
> 4. The Spoolman client init + health check boilerplate is now copy-pasted across link_spool, unlink_spool, and others. A small helper would clean that up, but fine to
>    leave for later.
> 5. LinkSpoolRequest accepts {} since all 3 fields are optional — a Pydantic @model_validator requiring at least one would be more robust, but the route-level validation covers it.

1, 2 and 3 of your Must Fix have been resolved with the latest set of commits.

shrunbr:
Saw there were some commits to 0.2.3b1 here. I did testing with my changes and those commits, still seem to be working fine.

shrunbr:
> **1. Tag priority inconsistency between lookup and linking**
> 
> The trayTag lookup in PrintersPage.tsx uses tray_uuid || tag_uid || fallback: const trayTag = (tray?.tray_uuid || tray?.tag_uid || getFallbackSpoolTag(...))?.toUpperCase();
> 
> But in LinkSpoolModal.tsx, the tag sent to Spoolman uses the opposite priority: const spoolTag = tagUid || trayUuid;
> 
> If a spool ever has both tray_uuid and tag_uid, it would be looked up by tray_uuid but linked by tag_uid — meaning a linked spool wouldn't show as linked.
> 
> Quick fix — change LinkSpoolModal.tsx to match the lookup order: const spoolTag = trayUuid || tagUid;
> 
> **2. Fallback tag stability — ams.id**
> 
> The serial hash fixed the printer portion of the fallback tag, but getFallbackSpoolTag still uses ams.id: getFallbackSpoolTag(printer.serial_number, ams.id, slotIdx)
> 
> If ams.id is the database auto-increment ID, removing and re-adding an AMS unit would change it — silently breaking all existing fallback tag links for that AMS. Same class of issue you already fixed for the printer ID. Can you confirm whether ams.id here is the DB ID or the hardware unit index (0-3)?
> 
> **3. Verify Unlink icon import**
> 
> FilamentHoverCard.tsx uses but I don't see Unlink added to the lucide-react import. Please verify it's imported — otherwise this will fail at runtime.
> 
> **4. Spoolman client init boilerplate**
> 
> The unlink_spool endpoint copies the same 9-line client init + health check block that's already duplicated across link_spool, sync_printer_ams, get_spools, get_unlinked_spools, get_linked_spools, etc. A small helper like async def get_ready_spoolman_client(db) that handles the enabled check, client init, and health check would clean this up. Not blocking, but this PR adds another copy so it's a good time to consolidate.

I've fixed your 1.

2 sounds like a niche use case and should be fixed by keeping AMS entries in the database even upon deletion (soft delete) and actually removing them after a certain amount of time, this seems like a more logical approach. What is the use case around someone removing and re-adding an AMS?

3 Look at the screenshots for yourself

---

Whatever AI other than Copilot you're using to check pull requests will find issues the more you ask it to. You need to start double checking it's findings. Your 1 in this reply for example _"change LinkSpoolModal.tsx to match the lookup order: const spoolTag = trayUuid || tagUid;"_ is talking about flipping back the tag priority that you wanted me to change already because of a copilot suggestion. Also 3, screenshots clearly show an icon there.

This is a feature that [per the documentation](https://wiki.bambuddy.cool/features/spoolman/#adding-spools) should be working. I've taken the time to get it working because I am one of those people who can benefit from it and think it'll be a game changer for this. All of the testing you've setup has passed and any issues from here on out are pure speculation. This is my last edit to the PR, it can either be merged or closed.

maziggy:
You're right, and I owe you an apology. I went back and checked — point 1 was asking you to undo a change I asked for in the previous round, and point 3 was wrong since Unlink is already imported. I need to do a better job verifying review findings before posting them.

The code looks good. Tag priority is consistent (tagUid || trayUuid) across both lookup and linking, the unlink flow has proper confirmation, and tests pass. Merging — thanks for your patience and for getting this feature working.

PR #719, opened 2026-03-15
[Fix]: Send Bambu RFID Tags to Spoolman, Allow Unlink in Manual Mode, Stop Location Clearing for Generic Spools

### LINKED ISSUE #714: [Bug]: Manual "Link to Spoolman" rewrites RFID tag to short legacy value, breaking existing AMS/Spoolman matching
### What happened?

I am using Bambuddy with Spoolman and only genuine Bambu Lab RFID spools.

Auto-sync appears to identify AMS spools using a **long RFID tag** (32 hex chars), for example:

- detected by auto-sync: `F6CE8D9015614E658E6D69303DE38A22`

When I manually click **"Link to Spoolman"** for that spool and select an existing Spoolman spool, Bambuddy appears to overwrite the tag in Spoolman with a **different short tag** instead, for example:

- after manual link: `0AE9D2FE00000100`

After that:
- the Printers tab shows **"Link to Spoolman"** again instead of **"Open in Spoolman"**
- spool fill level is missing in the Printers tab
- auto-sync may no longer match the spool correctly unless I manually restore the long RFID tag in Spoolman

So it looks like:
- **auto-sync / AMS detection** uses the long RFID tag
- **manual "Link to Spoolman"** writes a different short/legacy tag
- those two code paths do not agree on the spool identifier

### Expected Behavior

Manual linking should preserve/use the same spool identifier that AMS auto-sync uses.

If the spool is already matched by its RFID tag, linking it manually should not replace the tag with a different value.

After linking, the UI should consistently show:
- **"Open in Spoolman"** instead of **"Link to Spoolman"**
- correct spool fill level in the Printers tab

### Actual behavior
Manual linking changes the Spoolman tag from the long RFID tag to a short different value, which seems to break Bambuddy's linked-state/UI and causes inconsistent behavior between matching and linking.

### Steps to Reproduce

1. Use a genuine Bambu Lab RFID spool in AMS
2. Let Bambuddy auto-sync to Spoolman
3. Observe that Bambuddy detects/creates/updates the spool using a long RFID tag, e.g.:
   - `F6CE8D9015614E658E6D69303DE38A22`
4. In Spoolman, ensure that spool has this long tag
5. In Bambuddy, click **"Link to Spoolman"** on that spool and select the correct existing Spoolman spool
6. Check the tag in Spoolman again
7. Observe that the tag was changed to a different short value, e.g.:
   - `0AE9D2FE00000100`
8. Go back to the Printers tab
9. Observe that Bambuddy now shows **"Link to Spoolman"** again instead of **"Open in Spoolman"**, and fill level may be missing

### Printer Model

X1 Carbon

### Bambuddy Version

v0.2.2b3

### Printer Firmware Version

01.11.02.00

### Installation Method

Docker

### Operating System

Docker

### Logs

Relevant behavior from logs:

- auto-sync creates/updates using long tags, e.g.:
  - `Creating new spool in Spoolman for PLA Wood (tag: F6CE8D9015614E65...)`
  - later:
  - `Updating existing spool 24 for tag F6CE8D9015614E65...`

This suggests AMS matching is using the long RFID tag correctly.

### Environment

- Bambuddy version: v0.2.2b3
- Deployment: Docker / Portainer on Synology
- Printer: Bambu Lab X1C
- Spool type: only genuine Bambu Lab RFID spools
- Spoolman enabled

### Additional notes

This may be related to the linked/unlinked UI state from #210, but this issue seems deeper:
the actual identifier written during manual linking appears to differ from the identifier used by AMS auto-sync.

### Screenshots

_No response_

### Additional Context

_No response_

### Checklist

shrunbr:
## Description

This PR has trayUuid take precedence over tagUid so that the Bambu spool RFID is sent to Spoolman and not the generic placeholder needed for generic spools. While fixing this, it became clear to me that if someone were running sync in manual mode they may want the ability to unlink Bambu spools from Spoolman themselves so the button now appears on Bambu spools but only if Spoolman is in Manual mode.

After testing sync in manual mode, I noticed that even though Bambuddy said it was skipping non-Bambu spools it would clear the location from non-Bambu spools in Spoolman. To fix that, we now look for the 32-char tag on the spool before clearing it since the generic ones we generate are 16-chars. Now generic spool locations are not cleared from Spoolman upon sync, they are truly skipped.

**COPILOT GENERATED SUMMARY**
This pull request improves the handling of Bambu Lab spools and their integration with Spoolman, focusing on more accurate identification of spools using tray UUIDs, better sync mode awareness, and clearer logic for linking/unlinking spools. It also updates the UI and tests to reflect these changes.

## Related Issue

Fixes #714 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [X] Test addition or update

### Backend: Bambu Lab Spool Handling

* The backend now only clears locations for removed spools if they are Bambu Lab spools, identified by a 32-character RFID tag, ensuring generic spools are not affected.
* Unit tests for clearing locations have been updated to use realistic 32-character tags, improving test accuracy for Bambu Lab spool scenarios.

### Frontend: Spool Linking and Unlinking Logic

* The logic for linking spools in `LinkSpoolModal` now prioritizes `trayUuid` (Bambu Lab path) over `tagUid` (generic path), ensuring the correct identifier is used for Bambu Lab spools.
* The `FilamentHoverCard` and related UI now conditionally show the unlink option based on the sync mode, hiding it for Bambu Lab spools when auto-sync is enabled. 
* The `PrintersPage` and `PrinterCard` components now fetch and propagate Spoolman sync mode settings, enabling context-aware UI behavior for linking and unlinking spools.
* The `LinkSpoolModal` tests have been expanded to cover both Bambu Lab (trayUuid present) and generic (trayUuid missing) linking paths, ensuring correct behavior in both scenarios.

## Screenshots

Unlink button on a Bambu spool while in Manual sync mode
image

Bambu Lab RFID tags showing when available
image

Generic tags showing when RFID tags are not available
image

RFID tag in Spoolman
image

## Testing

- [X] I have tested with my printer model: 

## Checklist


shrunbr:
While I recognize copilots suggestion I think that it's incorrect. Our generated placeholder tags are only 16 characters and as of right now Bambu Lab spool RFID tags are only 32 characters, if any of that changes we can adjust the code.

shrunbr:
> One issue I spotted:
> 
> len(spool_uuid) != 32 gate in clear_location_for_removed_spools() can leave stale locations
> 
> sync_ams_tray() falls back to storing the 16-char tag_uid in Spoolman when tray_uuid isn't available (line 808-809). The new len != 32 check will skip those spools entirely, so their locations will never get cleared when they're removed from the AMS.
> 
> The problem is that Bambuddy-generated fallback tags (from getFallbackSpoolTag) are also 16 chars, so you can't distinguish them from real 16-char RFID tags by length alone.
> 
> A couple of options:
> 
> * Align with is_bambu_lab_spool() and accept both 16 and 32 char hex tags, then find another way to exclude the generated fallback tags (e.g. check against the known
>   fallback format)
> * Store a marker in the extra field at sync time (e.g. "source": "fallback" vs "source": "rfid") so clear_location_for_removed_spools() can check that directly

That's similar to the suggestion copilot had. Let me at least state my case because I adjust further :P 

The gate is specifically checking for 32 char hex tags because that is just how the RFID tags work. There are some people who have [done research into the Bambu Lab RFID tags](https://github.com/Bambu-Research-Group/RFID-Tag-Guide/blob/main/BambuLabRfid.md). Each block of the RFID tag is 16 bytes which is represented by two hex characters, 16 bytes x 2 hex characters per byte = 32 characters. Anything less/more is a corrupted RFID tag and shouldn't be messed with anyway but I think there would be a larger problem outside of our control if that were to happen.

16 characters are what we are using as our generated tag for generic spools so we don't want to mess with those. Since those are supposed to be "skipped" upon sync and we're not really skipping them by adjusting the location.

maziggy:
Ahhh.....I see.

PR #750, opened 2026-03-18
Improve Home Assistant notifications

### LINKED ISSUE #656: [Feature]: allow notifications to send via Home Assistant
### Problem or Use Case

send notifications via home assistant not possible

### Proposed Solution

as there is already a HA intergration use this api access to send notifications to HA as HA also supports push to mobile app. 

### Alternatives Considered

_No response_

### Feature Category

Notifications

### Priority

Nice to have

### Mockups or Examples

_No response_

### Contribution

### Checklist

mrtncode:
## Description
Here again ☺️

## Related Issue

Fixes #656 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Add Home Assistant notify services for notifications
- Adjust frontend -> Add input field
-

## Testing

## Checklist

## Additional Notes

Two things :)
1. I dont know if the frontend should be built by me.. I have built the frontend, if that shouldnt be, we can just revert the last commit.
2. You mentioned in the last PR that I should use i18n keys. But the only frontend change is in AddNotificationsModal.tsx. The other input fields have hardcoded strings (not t('...') keys)

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -75,7 +75,7 @@ jobs:
         run: docker build -t bambuddy:security-scan .
 
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@0.34.0
+        uses: aquasecurity/trivy-action@v0.35.0
         with:
           image-ref: 'bambuddy:security-scan'
           format: 'sarif'
@@ -91,7 +91,7 @@ jobs:
           category: trivy
 
       - name: Run Trivy for Dockerfile/IaC
-        uses: aquasecurity/trivy-action@0.34.0
+        uses: aquasecurity/trivy-action@v0.35.0
         with:
           scan-type: 'config'
           scan-ref: '.'


maziggy (changes requested):
**A few things to address before merging:**

  **Blocker**

  - The image_data=image_data was removed from the _send_webhook() call (line 571). That's unrelated to HA and breaks camera snapshots for webhook notifications — please
  revert that line.

  **Should fix**

  - The HomeAssistantConfig class in schemas/notification.py isn't imported or used anywhere — it's dead code. Please remove it. The config is stored as a freeform JSON dict like all other providers, and the service layer reads it via config.get("service") which is fine.
  - config.get("ha_service") fallback on line 524 — there's no existing data with that key, and the frontend field is service. Drop the ha_service fallback to keep it clean.
  - Please remove the static/ changes (built frontend assets). I build those as part of the release process, so they shouldn't be in the PR.
  - The service string gets concatenated directly into the URL. Would be good to validate that the domain/service parts only contain [a-zA-Z0-9_] to prevent path traversal
  (even though it only reaches the user's own HA instance).
  - Re: i18n — you're right that the existing fields in that modal are hardcoded too, so it's consistent. I'll clean that up across the board separately.

mrtncode:
Thank you for detailed feedback! It seems there was some conflics as I ported this to the dev branch (deleted lines, ..). 
I will look into this later when coming back from work :)

mrtncode:
@maziggy Thank you for good feedback! I think I fixed all :)

maziggy:
Thanks!

PR #761, opened 2026-03-20
Feature: Admin Set Default Nav-Menu Order

### LINKED ISSUE #586: [Feature]: Left hand navigation menu order
### Problem or Use Case

It would be nice if the left-hand navigation menu order when set by an admin would propagate to all users.

### Proposed Solution

Make this an admin setting to set menu order.

### Alternatives Considered

_No response_

### Feature Category

Other

### Priority

Would improve my workflow

### Mockups or Examples

_No response_

### Contribution

### Checklist

cadtoolbox:
## Description

This feature allows an Admin when User Authentication is enabled to set the default menu order based on how the Admin has the menu order configured. 

1. Admin customizes their own menu order.
2. Admin uses the new 'Set Default' toggle within the Admin; General; Sidebar Order settings.
3. New user's menu order updates to new default upon first login.
4. User is allowed to re-configure menu afterwards as normal.

## Related Issue

Fixes #586 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- New 'Set Default' button within the Admin; General; Sidebar Order settings.

## Testing

- [X] I have tested with my printer model: N/A

## Checklist


maziggy (changes requested):
**Bug: "Set Default" button visible to non-admin users**

  The button is gated on authEnabled, but that's true for all authenticated users, not just admins. A non-admin will see it, click it, and get a 403 with a generic error
  toast. Add a permission check:

  {authEnabled && hasPermission('settings:update') && (
    <Button
      variant="secondary"
      size="sm"
      className="whitespace-nowrap"
      onClick={handleSetDefaultSidebarOrder}
    >
      <ListOrdered className="w-4 h-4" />
      {t('settings.setDefault')}
    </Button>
  )}

  **Minor: JSON.parse outside try/catch**

  In handleSetDefaultSidebarOrder, the localStorage parse happens before the try block:

  const stored = localStorage.getItem('sidebarOrder');
  const orderArr = stored ? JSON.parse(stored) : defaultNavItems.map(i => i.id);

  If localStorage is corrupt, this throws unhandled. Move it inside the try block.

  **Design: No way to clear the default**

  Once set, there's no way to remove the default (reset to "no default"). Would be good to add a "Clear Default" option that sets locked_sidebar_order to "", so admins can go
   back to letting everyone start with the built-in order.

  **Nit: Field naming**

  The DB field/schema still says locked_sidebar_order from the earlier lock/unlock iteration, but the behavior is now "set default" (users can still reorder after).
  default_sidebar_order would be clearer. Not a blocker — just something to consider cleaning up.

cadtoolbox:
@maziggy Fixed bugs per your comments. The Reset button now also resets the order for all users, not just the Admin.

maziggy (changes requested):
**Bug: Reset clears default for ALL users**
  handleResetSidebarOrder currently pushes the default nav order to the server when an admin clicks "Reset". This means resetting your own sidebar also silently resets every
  user's sidebar on their next refresh. The "Reset" button should only reset the admin's local sidebar — clearing the server-side default should be a separate, intentional
  action (or not happen at all from "Reset").

  **Dead code: locked_sidebar_order fallback**
  The backend endpoint falls back to reading locked_sidebar_order, but nothing writes that key. If this was from an earlier iteration it should be removed, or if it's for
  migration from a previous version, it needs a comment explaining what writes it.

  **Design concern: No opt-out for users**
  Once an admin pushes a default, users get their sidebar overwritten on next load. Any new push (same order, new timestamp) overwrites user customizations again. Users have no way to say "keep my own order." Consider only applying the default when a user has never customized their sidebar, or adding an opt-out mechanism.

  **Minor: useEffect dependency array**
  The useEffect in Layout.tsx uses setSidebarOrder and saveSidebarOrder but they're not in the dependency array. ESLint's exhaustive-deps rule will likely flag this.

cadtoolbox:
I'll think through these new comments. I tend to error on "Admin knows best" which isn't always doable for everyone.

cadtoolbox:
@maziggy I made some changes to make it a little less obtrusive. If the admin sets their order and then toggles the "Set Default" button, any NEW user will have the menu order set to that default initially. Users are free to customize the order without it ever changing. Existing users have no impact to the menu order.

Issue #713, opened 2026-03-15
[Feature]: Multiple Switches for a single Printer


ManuelW77:
### Problem or Use Case

If i have multiple switches for a single printer, a heater as example, i only have the second switch as a script link. It would be nice, to have the row as the first switch is.

### Proposed Solution

A second row for a second switch with options for auto off.
For a heater as example set off as soon as the print is done.

<img width="637" height="117" alt="Image" src="https://github.com/user-attachments/assets/6403cecd-085a-4b63-a411-d4b40e73e5f8" />

### Feature Category

Smart Plugs

### Priority

Critical for my use case

### Contribution

- [x] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Since we are talking about HA switch entities, why don't you automate it on the HA side?

ManuelW77:
Because i don't want automations here and there and it is critical function to disable a heater. So it's more safe to have it at one place.
One of ,my reasons to switch to bambuddy is, to have all at one place.

maziggy:
Got it. This would require a complete refactoring of the power plugs implementation. Not sure if it's woth the work.

Anyway, I leave it in the queue to pick it up on a later point in time again.

ManuelW77:
Ok, can i do this?

maziggy:
Yes...sure, already assigned the issue to you. But please follow our rules. Thanks!

ManuelW77:
i give my best, hope its fine now :)

maziggy:
Had to revert and clode your PR. See commment in PR.

PR #786, opened 2026-03-22
Feature/multiple switches per prtinter

ManuelW77:
## Description

Have more than one switch at the printer card, because of heater or something else.
Adds other switches as a new line below.

## Related Issue

https://github.com/maziggy/bambuddy/issues/713

Fixes #

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

Backend
backend/app/api/routes/smart_plugs.py — +11 Zeilen

Neuer Endpoint GET /smart-plugs/by-printer/{printer_id}/all
Gibt alle Smartswitches eines Druckers als Array zurück (kein Filter)
Auth-geschützt via SMART_PLUGS_READ Permission
Frontend API
frontend/src/api/client.ts — +1 Zeile

Neue Methode getAllSmartPlugsByPrinter(printerId) → SmartPlug[]

## Screenshots

SCR-20260322-lyod

## Testing

- [X] I have tested with my printer model: 

## Checklist

### DIFF
--- a/frontend/src/components/SmartPlugCard.tsx
+++ b/frontend/src/components/SmartPlugCard.tsx
@@ -81,7 +81,6 @@ export function SmartPlugCard({ plug, onEdit }: SmartPlugCardProps) {
       // Also invalidate printer-specific smart plug queries to keep PrintersPage in sync
       if (plug.printer_id) {
         queryClient.invalidateQueries({ queryKey: ['smartPlugsByPrinter', plug.printer_id] });
-        queryClient.invalidateQueries({ queryKey: ['smartPlugsByPrinter', plug.printer_id] });
       }
     },
   });


ManuelW77:
Thanks, i have corrected this.

maziggy:
Unfortunately we had to revert the merge because of some issues that surfaced during testing.

  Core issue: You can't actually add a second switch to a printer. The Add modal filters out printers that already have a switch assigned, which defeats the purpose of the feature.

  Beyond the frontend, there are several backend code paths that assume a single plug per printer and will crash (MultipleResultsFound) or misbehave once multiple plugs exist:

  1. print_scheduler.py:1341-1344 — _get_smart_plug() uses scalar_one_or_none() which throws if a printer has >1 plug. This breaks auto power-on when the queue starts a print and auto power-off after completion.
  2. print_queue.py:806-807 and :823-824 — The stop endpoint also uses scalar_one_or_none() twice (initial lookup + re-fetch inside cooldown), same crash.
  3. print_queue.py:827 — Hardcodes tasmota_service.turn_off(plug) instead of using smart_plug_manager.get_service_for_plug(), so HA plugs would be turned off via the wrong service.
  4. smart_plug_manager._get_plug_for_printer() — Only returns the first non-script plug. If someone has multiple power plugs, only one gets auto-controlled. This might be intentional but needs to be a conscious design choice.

I did not do a full check, but there might be additional changes required.

PR #814, opened 2026-03-25
Spoolbuddy init and diagnostic improvements

Keybored02:
## Description

This pull request refactors and enhances the diagnostic scripts for the PN5180 NFC reader and NAU7802 scale.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Refactored both scripts to use shared driver classes from the daemon directory, eliminating code duplication.
- Added and decoded additional diagnostic registers (SIGPRO_CONFIG, TEMP_CONTROL, SYSTEM_STATUS) with human-readable output, including baudrate and temperature decoding.
- Added self-tests for SPI device and GPIO pin accessibility.

## Testing

- [x] I have tested with my printer model: H2C/ RPi 3B+

## Checklist

### DIFF
--- a/spoolbuddy/scripts/pn5180_diag.py
+++ b/spoolbuddy/scripts/pn5180_diag.py
@@ -69,10 +69,6 @@ def _check_spi_device_access() -> str:
     return spi_path
 
 
-def _pin_state_name(value: gpiod.line.Value) -> str:
-    return "ACTIVE" if value == gpiod.line.Value.ACTIVE else "INACTIVE"
-
-
 def _self_test_control_pins(nfc: PN5180):
     """Toggle NSS and RST pins and print observed line state.
     Uses public set_pin/get_pin methods to avoid direct access to driver internals.


maziggy (changes requested):
**Daemon driver side effects**

  The changes to nau7802.py affect the daemon too, not just the diagnostic script:

  - logger.debug() was replaced with print() — the daemon service would spam stdout with init messages during normal operation. The shared driver should keep using logging; the diagnostic script can capture or wrap output separately.
  - The init() sequence now flushes 4+2 readings and runs AFE calibration, adding ~1.5s+ to startup time vs the previous ~200ms. Is this fixing an actual issue you saw, or is
   it a "nice to have"? If the daemon init was working fine before, this change should either be opt-in or limited to the diagnostic path.
  - The module docstring was changed to say "Scale Diagnostic" but this is the shared daemon driver, not the diagnostic script.
  - Type annotations were removed from __init__ — should keep those.

  **pn5180_diag.py cleanup**

  After removing the local PN5180 class, there's a fair amount of dead code left behind:
  - All CMD_* constants (the driver handles SPI internally now)
  - BUSY_PIN (driver handles it)
  - Module-level REGISTER_NAMES dict (replaced by REGISTER_NAMES_DUMP inside run_diagnostics() — pick one and delete the other)

  Also _self_test_control_pins() accesses the private nfc._lines attribute directly — this will break if the driver internals change. Would be better to expose a small public
   method on the driver for this, or at minimum add a comment acknowledging the coupling.

Keybored02:
- The init() sequence now flushes 4+2 readings and runs AFE calibration, adding ~1.5s+ to startup time vs the previous ~200ms. 

It's the manufacturer's suggested power up and calibration sequence.

PR #847, opened 2026-03-29
Fix nozzle mapping logic for single active extruder scenarios in 3MF …

### LINKED ISSUE #827: [Bug]: Wrong filament mapping – dual extruder (H2C) has only one active nozzle
### Bug Description

The H2C allows a nozzle size to be selected in the Vortek system that is not available in the left extruder—for example, 0.4 mm on the left and 0.2 mm on the right.

If a print is then started that exclusively uses the corresponding nozzle on the right extruder, a setup is created which erroneously claims that the nozzle in use is the one on the left. This occurs because only one extruder is detected, leading the system to assume the groupID must address that specific one.

Consequently, the filament assignment selected in the slicer cannot be chosen within the UI of Bambuddy.

Happens with docker and git clone, Linux and MacOS

### Expected Behavior

The same filament mapping/slot should be possible between Bambuddy and Bambu Studio

### Steps to Reproduce

1. Use different nozzles in the vortek system than in the left nozzle (e.g., only 0.2 in L but at least one 0.4 in R)
2. Sync the nozzles with Bambu Studio. You should receive Left Nozzle (0), Right Nozzle (1+) if you select 0.4 as nozzle size
3. Slice the model with 0.4 and select the corresponding filament from the AMS (I used AMS B)
4. Upload print to Bambuddy, press print and look into the filament mapping, this shows now only the option to select filament from AMS A

### Printer Model

H2C

### Bambuddy Version

0.2.2.1

### Printer Firmware Version

01.01.05.00

### Installation Method

Manual (git clone)

### Operating System

macOS

_No response_

### Screenshots

Same 3mf

Current Bambuddy version. Only AMS A, wrong colors:
Issue

Fix (and correct color mapping)
Solution

### Additional Context

Already fixed, if interested and assigned I happily create the corresponding pull request

### Checklist

behrinml:
## Description

The H2C allows a nozzle size to be selected in the Vortek system that is not available in the left extruder—for example, 0.4 mm on the left and 0.2 mm on the right.

If a print is then started that exclusively uses the corresponding nozzle on the right extruder, a setup is created which erroneously claims that the nozzle in use is the one on the left. This occurs because only one extruder is detected, leading the system to assume the groupID must address that specific one.

Consequently, the filament assignment selected in the slicer cannot be chosen within the UI of Bambuddy.

## Related Issue
Fixes #827 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- short check during filament mapping if only one extruder is active. If this is true, the filament mapping leads to groupId 0 as this is the only one reported from the slicer.

## Screenshots

Previous:
issue
Afterwards:
solution

## Testing

- [X] I have tested with my printer model: H2C

## Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -8,6 +8,18 @@ body:
       value: |
         Thanks for taking the time to report a bug! Please fill out the form below.
 
+  - type: dropdown
+    id: component
+    attributes:
+      label: Component
+      description: Which part of the project is affected?
+      options:
+        - Bambuddy
+        - SpoolBuddy
+        - Both
+    validations:
+      required: true
+
   - type: textarea
     id: description
     attributes:
@@ -59,7 +71,7 @@ body:
         - Multiple printers
         - Not printer-related
     validations:
-      required: true
+      required: false
 
   - type: input
     id: version
@@ -71,13 +83,18 @@ body:
       required: true
 
   - type: input
+    id: spoolbuddy_version
+    attributes:
+      label: SpoolBuddy Version
+      description: If SpoolBuddy-related, which version is running? (Check SpoolBuddy Settings → Updates)
+      placeholder: e.g., 0.1.0
+
+  - type: input
     id: firmware
     attributes:
       label: Printer Firmware Version
-      description: Which firmware version is your printer running? (Check printer screen or Bambu Handy app)
+      description: Which firmware version is your printer running? (Check printer screen or Bambu Handy app). Leave blank if not printer-related.
       placeholder: e.g., 01.08.00.00
-    validations:
-      required: true
 
   - type: dropdown
     id: installation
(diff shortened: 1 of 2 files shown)


maziggy (changes requested):
Thanks for the fix — the diagnosis is spot-on. A few things to address before merge:

  **Bug: hardcoded extruder 0 doesn't generalize**

  When only one extruder is active, the fix maps all filaments to 0:
  nozzle_mapping[int(filament_elem.get("id"))] = 0

  This works for right-extruder-only prints (right = 0), but if someone prints using only the left extruder (1), all filaments would incorrectly map to the right. The value should come from the active extruder's physical ID:

  active_idx = active_extruders.index(1)
  target_extruder = int(physical_extruder_map[active_idx])
   ...
  nozzle_mapping[int(filament_elem.get("id"))] = target_extruder

  Edge case: sum(active_extruders) == 0

  If extruder_nozzle_stats exists but all entries show zero usage, sum < 2 is true and we'd still enter the branch. Using sum(active_extruders) == 1 instead of < 2 would be safer.

  **Minor: comment on stats format**

  The partition("#")[2] parsing is a bit opaque — a short comment showing what extruder_nozzle_stats looks like (e.g. "0.4#3|0.2#0") would help future readers.

PR #889, opened 2026-04-04
[Feature] Improve AssignSpoolModal filtering logic

Keybored02:
## Description
This PR improves Assign Spool usability in kiosk mode and tightens the spool filtering flow. It adds a viewport-safe modal layout so the Assign Spool dialog fits on smaller screens, and it refines the filtering logic with localized fields plus a “view all” toggle.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- The old logic always narrowed the list by tray profile when profile data existed, which was good for safety but too strict when the metadata was incomplete or when one needed to browse more broadly. The new logic keeps the search filter, but makes the tray/profile filter optional behind a “Show all spools” toggle, so you can still fall back to the full manual spool list without changing the underlying assignment rules.
- Now the modal is capped to 90vh, the body scrolls internally, and the footer stays available, which keeps the Assign / Cancel actions accessible on small screens.

## Screenshots

image

## Testing

## Checklist

## Additional Notes
Tested on SpoolBuddy too.
-->


maziggy (changes requested):
**1. Profile filter should use normalizeValue() (bug)**

  The new tray profile filter does its own .trim().toUpperCase(), but the existing mismatch warning in handleAssign uses normalizeValue(). These should stay consistent to avoid drift. Replace:

  const trayProfile = (trayInfo.profile || trayInfo.type || '').trim().toUpperCase();
  // ...
  const spoolProfile = (spool.slicer_filament_name || spool.slicer_filament || '').trim().toUpperCase();

  With:

  const trayProfile = normalizeValue(trayInfo.profile || trayInfo.type);
  // ...
  const spoolProfile = normalizeValue(spool.slicer_filament_name || spool.slicer_filament);

  normalizeValue is already defined in the same file and does the same trim + uppercase, so no reason to duplicate.

  **2. Missing i18n keys for de/ja**

  showAllSpools was added to en.ts but needs entries in de.ts and ja.ts too.

  **3. Reset toggle state on reopen**

  disableFiltering persists across modal open/close cycles. If a user toggles "Show all spools" on for one tray slot, it's still on when they open the modal for a
  different slot. Add a reset when the modal opens, e.g.:

  useEffect(() => {
    if (isOpen) {
      setDisableFiltering(false);
    }
  }, [isOpen]);

maziggy (approved):

maziggy (approved):

PR #916, opened 2026-04-08
feat: Two-Factor Authentication & OIDC/SSO – full implementation with admin UI

netscout2001:
.


maziggy:
Your PR would overwrite my complete Github pipeline and also internal configuration files, which would be a disaster. This is not acceptable at all.

PR #920, opened 2026-04-08
Feature/printer search filter

### LINKED ISSUE #852: [Feature]: Search/filter printers
### Problem or Use Case

With a lot of printers to manage, it can be a little hectic trying to find a printer in the list. Ctrl+f in the web browser works fine, but it's a bit cumbersome on mobile.

### Proposed Solution

It'd be great if there was a search/filter bar at the top of the printers page that let me search for printers by name, with live filtering as the text is typed into the bar.

### Alternatives Considered

Ctrl+F in a web browser, but this is a bit clunky on mobile clients.

### Feature Category

UI/UX

### Priority

Nice to have

### Mockups or Examples

Image
Image

### Contribution

### Checklist

legend813:
## Description

Adds live search and filter controls to the Printers page to make it easier to find specific printers in large setups, especially on mobile where browser Ctrl+F is impractical. Filtering is purely client-side — no backend changes made.

## Related Issue

Fixes #852 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Added a live search bar below the page title that filters printers by name, model, location, and serial number (includes a clear button)
- Added a status filter dropdown (All statuses / Printing / Paused / Idle / Finished / Error / Offline)
- Added a location filter dropdown — only visible when at least one printer has a location configured
- All three filters are combinable; search bar and dropdowns are hidden when no printers are configured yet
- Added i18n keys for all 7 supported locales (EN, DE, FR, IT, JA, PT-BR, ZH-CN)

## Screenshots

image

image

image

image

image

## Testing

- [X] I have tested with my printer model: X1C

## Checklist

## Additional Notes

The status filter uses the existing client-side `queryClient` cache for status data, consistent with how the existing status sort (`sortBy=status`) works. Status filter results reflect the cache state at the time of selection — same behavior as the existing status sort, not live-reactive to WebSocket updates.


maziggy (changes requested):
Thanks for this — really clean implementation and nice that it's purely client-side!

  A couple of things I noticed:

  Empty state bug when using dropdown filters without text search

  The noSearchResults fallback only triggers when search.trim() is truthy:

  ) : sortedPrinters.length === 0 && search.trim() ? (

  If someone filters by status (e.g. "Printing") or a specific location and nothing matches, but the search box is empty, they'll see a blank page — no printers and no empty-state message. This should also check the dropdown filters:

  ) : sortedPrinters.length === 0 && (search.trim() || statusFilter !== 'all' || locationFilter !== 'all') ? (

  Related: the noSearchResults i18n key only interpolates the search term (No printers match "{{search}}"), so when filtering by status/location with no text search it would show No printers match "". Might be worth adding a more generic "No printers match your filters" message as a fallback when search is empty but dropdowns are
  active.

  Minor: accessibility

  The search input doesn't have an aria-label — the placeholder helps sighted users but isn't picked up by screen readers. Adding aria-label={t('printers.search')} would fix that.

maziggy (changes requested):
Thanks for this — scope is tight, i18n coverage across all 7 locales is thorough, and reusing the existing printers.status.* keys for the filter dropdown is the right call. A few things before merge:

  **Must-fix**

  1. Search term isn't trimmed before matching (PrintersPage.tsx ~line 6075):
  if (search.trim()) { 
    const q = search.toLowerCase();  // ← still has leading/trailing spaces
  2. The outer search.trim() guard only checks whether there's non-whitespace, but q itself is untrimmed, so " X1C" ends up searching for " x1c" inside printer names and
  fails to match. One-line fix:
  const q = search.trim().toLowerCase();

  **Should-fix**

  1. Status filter isn't reactive to WebSocket updates. The filteredPrinters memo depends on [printers, search, statusFilter, locationFilter, queryClient]. queryClient is a stable reference and the ['printerStatus', p.id] cache entries don't trigger a rerun when new status arrives. Concretely: filter by "Printing", a print finishes → the now-idle printer stays in the list until you reselect the filter.

  2. I know you noted this mirrors sortBy=status, and that's true, but the UX impact is worse for filtering than for sorting — sort staleness is cosmetic (wrong order),
  filter staleness is functional (wrong set of printers). Could you either:
    - subscribe to the query cache (queryClient.getQueryCache().subscribe(...)),
    - use useQueries to watch all printer statuses so React re-renders on update, or
    - at minimum mention the caveat in the UI so users aren't surprised?

  3. The clear-search <button> is missing type="button" and aria-label — small a11y polish:
  <button type="button" aria-label={t('common.clear')} onClick={() => setSearch('')} ...>
  4. Minor inconsistency worth noting: the "error" filter treats state === 'FAILED' || hmsErrors.length > 0 as error, while the existing status sort only uses
  hmsErrors.length > 0 for priority. The filter's behavior is arguably more correct — just flagging.
  5. A vitest test for filteredPrinters (search matching + combined filters + empty-state) would be nice insurance against regressions, not blocking.

  Once these issues  are fixed, this is good to merge.

legend813:
I changed everything accordingly - thanks for the hints :-)

I also fixed the failing `test_backend.sh` on Windows-Machines. If that's not okay and I shall remove it, just tell me.

maziggy:
No changes to internal/pipeline files please. I've deleted test_background.sh.

PR #932, opened 2026-04-09
Feature: print files directly from project view (closes #930)

### LINKED ISSUE #930: [Feature]: Print files directly from Project and mark as project related print
### Problem or Use Case

Inside the project view i would find it very helpful, when i can print a file directly.

### Proposed Solution

I'm talking about that view: 
Image

I would suggest to add not only a link to the File Manager, but a view like this: 

Image

Would save time by reducing the clickpath to the archive and linking it to the project.

### Alternatives Considered

_No response_

### Feature Category

Other

### Priority

Would improve my workflow

### Mockups or Examples

_No response_

### Contribution

### Checklist

legend813:
## Description

Show printable files from linked library folders directly in the project detail page, with Print Now and Add to Queue buttons per file. Removes the detour through the File Manager for common reprint workflows.

When Print Now or Add to Queue is triggered from the project detail page, the resulting archive is automatically linked to that project — no manual assignment needed afterwards.

## Related Issue

Closes #930

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Fetch files for all linked folders in parallel via useQueries
- Render each folder with its file list below the folder header link
- Show Play / CalendarPlus action buttons on sliced files (.gcode, .gcode.3mf)
- Wire PrintModal in reprint and add-to-queue modes
- Add noFiles / print / addToQueue translation keys to all 7 locales

- project_id flows explicitly from the UI through the full stack:
- PrintModal accepts projectId prop (only set from ProjectDetailPage)
- printLibraryFile API options + PrintQueueItemCreate carry project_id
- FilePrintRequest and PrintQueueItemCreate schemas accept project_id
- PrintDispatchJob stores project_id through background_dispatch
- archive_print() sets project_id on the created PrintArchive
- print_scheduler passes item.project_id when creating archive from queue

## Screenshots

image

image

image

## Testing

## Checklist

## Additional Notes

Prints are only flagged as project prints when they were started in the ProjectDetailView - the linkage only is not enough.


maziggy (changes requested):
A few things worth considering before merge:

  **Should consider**

  1. N+1 request pattern for folder contents. useQueries fires one GET /library/files?folder_id=X per linked folder. Fine for typical projects (2–3 folders), but worth
  noting if someone links a lot. Not a blocker.
  2. No server-side existence check on project_id. Backend passes it straight through to archive_print() / PrintQueueItem(). A bogus ID would hit the FK constraint at
  insert time → 500. Since the value only comes from internal UI state (never user-typed), exploitability is effectively zero, but a pre-check would produce a cleaner 404
  if something ever gets out of sync.

  **Nits**

  3. isSlicedFilename uses lower.includes('.gcode.') which would also match foo.gcode.bak. Unlikely in practice — fine as-is, or tighten to .endsWith('.gcode.3mf').
  4. The flex-shrink-0 → shrink-0 sweep across BOM / timeline / header is unrelated to this feature. Harmless, just scope creep.
  5. PR body says "Features #930" — should be Closes #930 if you want the issue to auto-close on merge. (Title already has (closes #930) so it may be covered.)
  6. No vitest coverage for the new file-list rendering or the conditional print-button logic.
  7. The reprintArchive branch in PrintModal/index.tsx doesn't receive project_id, which is actually correct (ProjectDetailPage only triggers the printLibraryFile branch).
   A one-line comment making that intent explicit would help future readers.

legend813:
Thanks for the thorough review @maziggy! Here's what I addressed:

**Should Consider**

1. **N+1 pattern** → replaced with a single bulk query. Extended `GET /library/files` with an optional `project_id` param that JOINs on `LibraryFolder` (no new endpoint needed, `LibraryFile` already had `folder_id`). `useQueries` in `ProjectDetailPage` is gone — replaced with one `useQuery` + `useMemo` grouping files by folder for the section-based render.

2. **project_id existence check** → added a pre-check in the library file print endpoint (same pattern as `create_folder` / `update_file` in the same file). A bogus `project_id` now returns 404 instead of hitting the FK constraint at insert time.

**Nits**

3. **isSlicedFilename** → tightened to `.endsWith('.gcode.3mf')` in both `ProjectDetailPage.tsx` and `FileManagerPage.tsx`. Added a regression test specifically for `.gcode.bak` to guard this.

4. **flex-shrink-0 sweep** → agreed, purely opportunistic cleanup while reading adjacent components. All equivalent in Tailwind v3+, leaving as-is.

5. **PR body** → changed to "Closes #930"

6. **Test coverage** → added 9 tests:
   - `ProjectDetailPage.test.tsx` (new): 4 `isSlicedFilename` cases (`.gcode` ✓, `.gcode.3mf` ✓, `.gcode.bak` ✗, `.stl` ✗), folder file rendering, and print modal open trigger
   - `PrintModal.test.tsx`: 2 `project_id forwarding` tests — verifies `printLibraryFile` sends `project_id` and `reprintArchive` intentionally does not

7. **reprintArchive comment** → added a two-line comment explaining the intentional absence of `project_id` on that branch.

maziggy (approved):

PR #939, opened 2026-04-10
Enhance plate-clear tracking and visibility in printer cards

### LINKED ISSUE #935: [Feature]: Enhance plate clear tracking and visibility
### Problem or Use Case

The ideal (to me) workflow for job queuing is:
submit jobs via orca to the job queue -> jobs automatically sent to available printers -> when job is completed mark the printer as ready again. 

Currently this works well as you can click "Plate cleared, start next job" button on the printer however this button is only visible if a print job is waiting in the queue: (autoDispatchQueue.length > 0) That means a finished printer with no waiting job has no per-printer “plate cleared” control in the normal card UI and therefore you cannot mark a printer "cleared" at the time of clearing relying on me remembering!

An [API endpoint](https://wiki.bambuddy.cool/reference/api/?h=cleare#clear-plate) exists to mark the printer as cleared but this is (to the best of my knowledge) not exposed in the webUI (until the point mentioned above where there are queued jobs).

Additionally tracking of the plate status flag is poor in the UI - there is not (to my knowledge) any visible status indicator to the user outside of querying the api directly.

### Proposed Solution

Implement a "Plate cleared" button as part of the printer card visible when the job state is not printing (e.g. idle or failed / finished) and only when the flag for plate cleared is false:

Image

Implement a status tracker that indicates the plate status to the user as part of the printer info with states: Plate Not Cleared (idle but not cleared, cannot receive job), Plate Cleared (ready for next job), Plate in use (during printing):
 

Image

Image

Image

These two changes will provide better management of plate status and provide a better overview of printer status hopefully making availability and clearing tasks clearer to the user. 

### Alternatives Considered

_No response_

### Feature Category

Print Queue & Scheduling

### Priority

Would improve my workflow

### Mockups or Examples

_No response_

### Contribution

### Checklist

EdwardChamberlain:
## Description
This PR improves the printer card workflow around plate-clear confirmation so the existing `plate_cleared` flag is visible and actionable directly in the main printer UI, without relying on the queue-specific flow.

Previously, the web UI only exposed a per-printer plate-clear action widget when a compatible queued job was waiting. As a result, a finished printer with no waiting queue item had no normal card-level way to be marked as cleared, even though the backend already supported that state through the existing API.

This change adds plate-clear visibility and control directly to printer cards, while keeping behaviour aligned with the existing scheduler semantics and workflow setting.

## Related Issue
Fixes #935 

## Documentation
**Pick one**:
- [ ] Docs PR(s) linked above
- [x] No docs update required — reason: UI behaviour changed, but no standalone documentation update is needed for this change

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made
 - Added a plate status badge to printer cards showing "Plate Cleared", "Plate in use", "Plate Not Cleared"
 - Added a Mark plate as cleared button to printer cards visible when not printing and uncleared. 
 - `Require plate-clear confirmation` button hides both indicator and button for those not interested. 
 - Updated the `Require plate-clear confirmation` setting description to clarify that disabling plate-clear confirmation also hides the plate status badge and card button
 - Added and updated frontend tests to cover the new printer card behaviour

## Screenshots
image

image

image

## Testing

- [X] I have tested with my printer model: P1S

All tests passed locally.

## Checklist

## Additional Notes
Behaviour when Require plate-clear confirmation is enabled:
 - RUNNING / PAUSE printers show Plate in Use
 - FINISH / FAILED printers with plate_cleared = false show Plate not Clear
 - FINISH / FAILED printers with plate_cleared = true show Plate Clear
 - The Mark plate as cleared button appears only for connected FINISH / FAILED printers where plate_cleared = false

Behaviour when Require plate-clear confirmation is disabled:
 - The plate status badge is hidden
 - The Mark plate as cleared button is hidden
 - Scheduler behaviour remains unchanged and does not rely on the flag for FINISH / FAILED printers

The existing bulk clear-plate action is intentionally left untouched but it might be worth considering also disabling this option when `Require plate-clear confirmation` is disabled as it doesn't serve any purpose. 

I only speak English - translated strings were generated using AI and **not** checked for accuracy by a native speaker.


maziggy (changes requested):
**One thing worth addressing**

  The "Mark plate as cleared" button is rendered inside the status.temperatures && viewMode === 'expanded' IIFE in PrintersPage.tsx, while the status badge lives in the
  always-visible indicator row. That means in compact view a user sees the "Plate not Clear" warning badge but has no way to act on it without first expanding the card.
  Either move the button out of the temperatures block so it renders in both view modes, or if that's intentional it'd be good to call out in the PR description.

  **Small things**

  - IDLE + plate_cleared: false shows "Plate Clear". Semantically fine (the scheduler only cares about FINISH/FAILED) and your test pins this behaviour, but worth noting
  it can briefly mislead a user who glances at the card right after starting a new print.
  - clearPlateMutation.onSuccess calls setQueryData({plate_cleared: true}) and then invalidateQueries — the refetch can theoretically flip the badge back for a frame if
  the backend is slow. Since clear-plate is synchronous it's unlikely to surface; you could drop the optimistic setQueryData and just rely on the invalidation for simplicity.
  - Missing tests: no FAILED state case (the description claims parity with FINISH), no permission-disabled button case, no compact-view case. Your 5 new tests cover the
  happy paths well, these would just be nice-to-haves.

EdwardChamberlain:
Thanks for the feedback. I will take a look at implementing at the weekend. It's not forgotten!

Regarding hiding of the button:
I think this just comes down to stylistic choice and I don't have a strong preference. 

It's annoying having to switch views to get to the button but that behaviour is inline with that of other controls in the printer card. The button adds visual noise for users selecting the most compact view who assumedly want a denser display of information

I suppose the decision can be boiled down into compactness vs utility but I think based on consistency of behaviour and making the compact view as compact as possible my preference is to keep the behaviour as in the PR currently. If we think this button is required in the compact view then we could look at adding a smaller rendition of the button (icon only with tool tip) for that view?

Keen to hear your thoughts as you have a better grasp of the direction for this project.

maziggy:
> I suppose the decision can be boiled down into compactness vs utility but I think based on consistency of behaviour and making the compact view as compact as possible my preference is to keep the behaviour as in the PR currently. If we think this button is required in the compact view then we could look at adding a smaller rendition of the button (icon only with tool tip) for that view?

I agree.

maziggy:
Any updates?

EdwardChamberlain:
Added a new icon to symbolise the plate clearing action and now use that on both the existing button for plate clearing and a new small button that appears as required on the "small" printers card view giving actionability from the small view.

image

image

maziggy:

maziggy (changes requested):
Thanks for the PR — the workflow improvement makes sense and the card layout looks clean in the screenshots. I hit a few issues when verifying the branch locally
  that are blockers before we can merge:

  **1. Frontend build fails**

  npm run build on the PR head (887541b) fails with:

  src/pages/PrintersPage.tsx(2853,17): error TS1005: ')' expected.
  src/pages/PrintersPage.tsx(2868,15): error TS1128: Declaration or statement expected.
  src/pages/PrintersPage.tsx(2868,16): error TS1005: '}' expected.
  src/pages/PrintersPage.tsx(2869,13): error TS1381: Unexpected token.
  src/pages/PrintersPage.tsx(2869,17): error TS1381: Unexpected token.

  The second "Mark plate as cleared" button (the full-width one, lines 2853–2867) is inserted after the closing </div> of the temperatures grid, but still inside the
  IIFE's return (...). An IIFE return can only hold a single JSX root, so the two siblings need a Fragment wrapper or a shared parent:

  return (
    <>
      <div className="flex items-stretch gap-1.5 flex-wrap">…</div>
      {showClearPlateButton && (<button>…</button>)}
    </>
  );

  (Our CI currently only runs security scans on PRs — no frontend-build check — which is why this wasn't flagged automatically. I'll add one in a separate change.)

  **2. plate_cleared field isn't in the API response**

  The component reads status?.plate_cleared ?? false, but backend/app/schemas/printer.py exposes awaiting_plate_clear, not plate_cleared. The new tests work because they mock plate_cleared into the status payload, but in production that field is undefined, which means:

  - The button will always render on connected FINISH/FAILED printers whenever require_plate_clear is on.
  - After clicking, the optimistic cache is set to plate_cleared: true, but the next /status poll overwrites it with undefined (→ coerces back to false), so the button
  flickers back a few seconds later.

  Two options:
  - Frontend-only: use !status?.awaiting_plate_clear as "cleared" and drive the button off status?.awaiting_plate_clear directly. Simpler, no backend change.
  - Backend change: add a dedicated plate_cleared: bool field to PrinterStatus in the schema and printer_manager, and reset it to false on every print start.

  I'd lean toward Option 1 — awaiting_plate_clear already carries the "finished + not yet acknowledged" semantic the card needs.

  **3. The plate status badge isn't implemented**

  The PR description mentions a badge showing "Plate Cleared / Plate in use / Plate Not Cleared", and you added plateStatus.cleared / notCleared / inUse to all 7 locale
  files — but those keys are never referenced in PrintersPage.tsx. Only plateStatus.markCleared is used.

  Three of the new tests assert on those badge strings:

  - 'shows plate clear status and action on finished printers when not cleared' → getAllByText('Plate not Clear')
  - 'shows plate clear status but no action while idle' → getAllByText('Plate Clear')
  - 'shows plate in use status while printing and hides the clear action' → getAllByText('Plate in Use')

  These can't be passing against the current diff since none of those strings are rendered. Could you double-check whether the local npx vitest run against the new test
  file actually succeeded? If the badge got dropped in a rebase, easiest path is to add it back next to getStatusDisplay(...) in both the compact and expanded views,
  gated on the same requirePlateClear flag.

  **Smaller items**

  - PrintersPage.tsx:1649 — plateCleared is only read once; fine to inline into needsPlateClear.
  - PrintersPage.tsx:2858 — small-view button uses title={... : undefined} while the compact-view version at 2636 uses title={... :
  t('printers.plateStatus.markCleared')}. Worth aligning for consistency.

maziggy:
Thanks!

maziggy:
Unfortunately we've experienced some bugs which were introduced by your PR. Will fix them now. Next time we'll need to be more careful.

maziggy:
Root cause — PR #939 (b046c2c, "Enhance plate-clear tracking and visibility") added awaiting_plate_clear but the flag lives on printer_manager._awaiting_plate_clear (not on PrinterState). printer_state_to_dict() — which builds every WebSocket printer_status payload — never emits it:

  - backend/app/services/printer_manager.py:796-870 (the result = {...} dict) has no awaiting_plate_clear key.
  - Only the HTTP endpoint fills it in (backend/app/api/routes/printers.py:634).

  So after a print completes:

  1. main.py:2470 sets the flag in-memory — correct.
  2. Backend sends WS print_complete, but useWebSocket.ts:234-239 intentionally skips invalidateQueries(['printerStatus']) (comment says "causes re-render cascade and browser freeze"; relies on "printer_status websocket messages will naturally update the status" — except they don't carry the field).
  3. WS printer_status merges keep streaming and setQueryData preserves the stale awaiting_plate_clear: false.
  4. The only path that surfaces true is the 30 s HTTP fallback poll (PrintersPage.tsx:1430). And since each WS merge calls setQueryData, React Query's dataUpdatedAt is pushed forward on every status tick — on a chatty printer that delays the refetch further, which is why it can sit at 5 min.

  "Plate Clear" pill with no Clear-Plate button = classic stale false: the "cleared" label shown in PrintersPage.tsx:1672-1675 is the fallback when awaiting_plate_clear is falsy.

  Fix (smallest diff): add awaiting_plate_clear to printer_state_to_dict. Since it already takes printer_id, it can read printer_manager.is_awaiting_plate_clear(printer_id)
  directly. One key in the returned dict, zero new wiring on the frontend — the existing merge does the rest.

  Affects A1/H2D/X1C equally because the path is transport-agnostic. H2D stuck in FINISH without the button is the same bug, just over a longer poll slip.

maziggy:
There are even more errors caused by this PR. Need to sort it out.

EdwardChamberlain:
I think this was caused during the rebase. It was a significant rebase with a lot of conflicts, and the PR was merged before I had a chance to properly test against the new base.

That was my mistake. I should not have pushed the rebased branch without testing it first, and I should have marked the PR as draft until that was done.

I have submitted a follow-up PR to address the duplicated buttons as a first step.

PR #963, opened 2026-04-13
feat: add embedded GCode viewer

### LINKED ISSUE #952: feat: plugin system — install, manage and convert OctoPrint plugins
## Summary

I'd like to propose adding a first-class plugin system to Bambuddy. I've built a working implementation on my fork and would like to contribute it upstream.

## Motivation

OctoPrint has a rich plugin ecosystem. Many users who migrate from OctoPrint to Bambuddy miss plugins like GCode visualisers, custom notifications, filament-change scripts, etc. A lightweight plugin system would let the community extend Bambuddy without requiring core changes for every feature.

### Plugin loader & base mixins
- Plugins are Python packages in `data/plugins/` — auto-discovered on startup
- Supported mixins: `StartupPlugin`, `ShutdownPlugin`, `EventHandlerPlugin`, `SettingsPlugin`, `AssetPlugin`, `SimpleApiPlugin`
- Plugin lifecycle events fire on print start/complete, archive creation, startup/shutdown

### Plugin API (`/api/v1/plugins`)
- List all plugins with metadata, enabled state, load status
- Upload a `.zip`, detect type (Bambuddy native or OctoPrint), preview conversion, confirm install
- Enable / disable (restart required)
- Per-plugin settings (GET/PUT merged with defaults)
- Static asset serving (`/{key}/assets/{path}`) with path-traversal protection
- `SimpleApiPlugin` GET and command dispatch

### OctoPrint plugin converter
- AST-based detection: checks for `import octoprint`, `__plugin_pythoncompat__`, class inheritance
- Maps `StartupPlugin`, `SettingsPlugin`, `AssetPlugin`, `SimpleApiPlugin`, `EventHandlerPlugin` 1:1
- Stubs unsupported mixins (`TemplatePlugin`, `BlueprintPlugin`) with clear comments
- Shows a preview of converted code before committing the install

### Frontend
- **Plugins** sidebar nav item
- Plugin cards: name/version/author/description, Loaded/Restart/Disabled badges, enable toggle, settings editor
- Upload modal with type detection, mixin support preview, and generated code diff
- "Open Viewer" button for plugins that expose a `static/index.html`

### Bundled example: PrettyGCode
Includes a conversion of [OctoPrint-PrettyGCode](https://github.com/Soopahfly/OctoPrint-PrettyGCode) as a reference plugin:
- Full Three.js 3D GCode viewer
- Layer highlight synced to Bambu `layer_num` via MQTT (byte-offset map built by parsing the GCode file)
- Nozzle animation using synthetic GCode commands (Bambu doesn't expose serial echo, so we reconstruct it)
- File picker from Bambuddy's library, printer selector with model-based bed sizes

## Implementation notes

- No breaking changes — the plugin system is entirely additive
- `docker-compose.yml` gets a `./plugins:/app/data/plugins` volume mount
- `PluginRecord` DB table added via SQLAlchemy (existing `init_db()` path)
- All plugin static assets served through the existing FastAPI layer

## Questions for maintainers

1. Is the `data/plugins/` location acceptable, or would you prefer a separate `PLUGINS_DIR` env var?
2. Should the OctoPrint converter live in core, or be a separate optional dependency?
3. Are there CI/linting requirements I should be aware of for the AST-based converter?

Happy to adjust the implementation based on feedback before submitting the PR.

Soopahfly:
## Summary

Adds [PrettyGCode](https://github.com/Kragrathea/pgcode) as a built-in GCode visualiser embedded directly in the Bambuddy layout, so users can preview and inspect GCode files without leaving the dashboard.

- **GCode Viewer** nav entry (Layers icon) loads the visualiser in the main content pane with the sidebar still visible
- **File picker** browses and searches the Bambuddy library for `.gcode` files across all folders, with live search filtering
- **Printer selector** populates from the Bambuddy API; switching printers updates the print bed dimensions automatically for X1C, P1S, A1, A1 Mini and other Bambu models
- Currently printing file **auto-loads** when a printer is active
- **Layer slider** scrubs through the model layer by layer; **play button** animates through layers at adjustable speed (1×, 3×, 10×, 25×)
- Viewer runs in a same-origin `<iframe>` so it does not interfere with React app state or routing

## Technical notes

- `gcode_viewer/` — PrettyGCode static assets with a custom `bambuddy_adapter.js` that bridges OctoPrint API patterns to Bambuddy's REST/WebSocket API, a `slider-shim.js` replacing the bootstrap-slider dependency, and a CSP meta tag
- `GCodeViewerPage.tsx` — iframe wrapper with a recursion guard (prevents infinite nesting if the SPA catch-all serves the React app into the iframe)
- Explicit `@app.get` routes for `/gcode-viewer/*` placed before the SPA catch-all (`app.mount()` loses to `/{full_path:path}` in some Starlette versions)
- Vite dev-server plugin serves `gcode_viewer/` at `/gcode-viewer/` without proxying to uvicorn during development
- Path traversal guard added to `library.py` `to_absolute_path()`
- `nav.gcodeViewer` i18n key added to all 7 locale files

## Docs

This adds a new nav item and feature — a wiki page will be needed. Happy to write it once this PR is reviewed.

## Screenshots

Screenshot 2026-04-12 165450

## Test plan

- [ ] Navigate to GCode Viewer in the sidebar — viewer loads in main pane with sidebar visible
- [ ] Click Load file — file picker opens and lists `.gcode` files from the library
- [ ] Select a file — 3D model renders in the viewer
- [ ] Layer slider scrubs through layers; play button animates
- [ ] Printer selector populates with connected printers; switching updates bed size
- [ ] Dev server (`npm run dev`) serves the viewer correctly at `/gcode-viewer/`

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


maziggy (changes requested):
Some issue to fix before we can merge.

  🔴 Blockers

  1. Security: slicer-download auth bypass (backend/app/api/routes/library.py:2502,2521)

  The PR removes await from two async def functions:
  token = create_slicer_download_token("library", file_id)   # returns a coroutine!
  if not verify_slicer_download_token(token, "library", file_id):  # coroutine is always truthy
  verify_slicer_download_token is async def → the unawaited call returns a coroutine object → not <coroutine> is always False → the 403 guard never fires, and any request
  with any token value downloads the file. create_... is equally broken — the response sends a coroutine instead of a token string. These changes are unrelated to the GCode
  viewer and must be reverted.

  2. Path-traversal guard has prefix-confusion bug (library.py:114, main.py:4428)

  if not str(resolved).startswith(str(base)):
  With base = /data/app, this permits /data/app_evil/secret. Use Path.is_relative_to(base) (3.9+) or startswith(str(base) + os.sep).

  3. to_absolute_path now raises on legacy absolute paths

  Previously an absolute path was returned verbatim (comment in code: "for backwards compatibility during migration"). The PR now calls .resolve() and raises ValueError if
  it's outside base_dir. Installs with legacy DB rows pointing to absolute archive paths will start throwing. Either log-and-skip, or preserve the old pass-through for
  absolute paths.

  🟡 Important

  4. Adapter misses sessionStorage token (bambuddy_adapter.js:33)

  App stores the JWT in sessionStorage when the user opts out of "remember me" (see frontend/src/api/client.ts:10). The adapter only reads localStorage, so users without
  persistent login get an unauthenticated viewer and every API call 401s. Fix:
  const token = sessionStorage.getItem('auth_token') ?? localStorage.getItem('auth_token');

  5. No tests for new surface

  Zero tests added for: /gcode-viewer route ordering, path-traversal guard on static dir, or to_absolute_path behaviour change. At minimum add one integration test covering
  /gcode-viewer/../main.py → 403 and one for legacy absolute paths in to_absolute_path.

  🟢 Minor 

  6. Licensing — PrettyGCode is AGPL-3.0 (compatible with Bambuddy's AGPL-3.0), but the bundled three.min.js, jquery.min.js, dat.gui.js, OBJLoader.js, etc. ship without any
  LICENSE/NOTICE file in gcode_viewer/. Add a gcode_viewer/VENDORED.md (or THIRD_PARTY_NOTICES.md) listing source URL, version, and license for each vendored lib.

  7. Static /gcode-viewer/* route has no auth gate. Static assets only — acceptable, but when auth is enabled the viewer page still renders; only API calls fail. Consider
  wrapping index.html serving behind the same auth check the SPA uses, so un-auth'd users see the login redirect.

  8. Supply-chain surface: ~15k lines of vendored JS (jQuery 3.x, three.js, dat.GUI). No SRI, no version pins documented. Document versions in the VENDORED.md above so
  future CVE triage is possible.

  9. Dev vite-plugin + proxy /gcode-viewer fallback is redundant — the middleware handles it; the proxy will never fire. Not a bug, just dead config.

  10. GCodeViewerPage uses height: '100vh' — inside the existing Layout this may overflow the viewport by the sidebar/header height and cause double scrollbars. Consider height: '100%' with a flex parent or a calc(100vh - <header>).

Soopahfly:
Hows that look now?

maziggy:
Content-Security-Policy: The page’s settings blocked the loading of a resource (frame-ancestors) at http://192.168.255.16:8000/ because it violates the following directive: “frame-ancestors 'none'” [gcode-viewer](http://192.168.255.16:8000/gcode-viewer/)

​
Screenshot 2026-04-14 at 09 31 59

Soopahfly:
Sorry, I hadn't tested in Firefox.  Should be working now.

maziggy:
The Gcode viewer should also replace the existing one on Archives page. Each archive card has a button and context menu item to open print file in Gcoede Viewer.

Screenshot 2026-04-19 at 11 17 36
Screenshot 2026-04-19 at 11 17 24

Soopahfly:
OK, I'll take a look

maziggy:
I'll take care about the remaining issues. Would like to get it merged. Alrready working on it.

PR #1013, opened 2026-04-17
feat(cloud): support China region for token-based login

Minidoracat:
## Description

The `/cloud/token` endpoint previously always used the global Bambu API
endpoint, so users with China-region access tokens could not validate
their token. The password-login flow already exposes a region selector;
this brings the token-login flow to parity.

## Related Issue

N/A — small enhancement mirroring the existing password-login region
option. Happy to open a tracking issue first if preferred.

## Documentation

**Companion docs PRs**: not opened (see reason below)

**Pick one**:
- [ ] Docs PR(s) linked above
- [x] No docs update required — reason: the new region selector mirrors
      the existing password-login control. No new concepts, config keys,
      or endpoints. Happy to add a wiki note if maintainers prefer.

## Type of Change

## Changes Made

- Backend: `CloudTokenRequest` gains a `region` field (default `"global"`)
- Backend: `bambu_cloud` exposes `reset_cloud_service(region)` so the
  singleton can be rebuilt against the requested region
- Backend: `/cloud/token` calls `reset_cloud_service` before `set_token`
  so the China endpoint is used when selected
- Frontend: `cloudSetToken` accepts a `region` argument
- Frontend: `ProfilesPage` token-auth step now shows the same
  Global / China region dropdown as the password-auth step

## Screenshots

image

image

## Testing

- [x] I have tested this on my local machine (Docker build + swap, token
      flow with Global and China regions)

## Checklist

- [x] My code follows the project's coding style (`ruff check` + `ruff format` pass)
- [x] My changes generate no new warnings (`tsc --noEmit` + `eslint` pass)

## Additional Notes

Follow-up observation (out of scope for this PR, happy to file separately):
the `/cloud/login` handler already accepts `CloudLoginRequest.region`
from the schema and the frontend sends it, but the handler doesn't read
`request.region` — it just calls `get_cloud_service()`. So the
password-login region selector is currently a UI-only affordance.


maziggy:
Next time pleasefollow our rules from CONTRIBUTING.md. As per instructions you need an assigned issue before starting to work on PR's.

Anyway....thanks for your contribution.

maziggy (changes requested):
Thanks for the contribution, @Minidoracat! The direction is right — China-region users should have parity with the password flow — but I'd like a couple of things addressed before we merge.

  **Blocking**

  1. Region isn't persisted. The User model stores cloud_token / cloud_email but no cloud_region column, and the Settings fallback doesn't store it either. After
  reset_cloud_service("china") the in-memory singleton is China, but on next backend restart get_cloud_service() lazily constructs a fresh instance with the default
  "global" region — the stored China token then gets used against api.bambulab.com and every call silently fails. Token setup looks like it succeeded, then breaks at the
  next restart.
  2. Global-singleton side effect. reset_cloud_service() mutates the process-wide singleton, and there are ~17 get_cloud_service() call sites across routes/cloud.py,
  routes/inventory.py, and services/github_backup.py. In a multi-user auth-enabled deployment, if user A sets a China token, user B's next /cloud/settings / firmware /
  device call runs their global-region token against the China base URL. Cross-user breakage.
  3. Minor: the prior _cloud_service._client (httpx.AsyncClient) isn't closed before overwrite.

  Suggested approach

  Either of these works — second one is smaller:

  - Add a cloud_region column to User (and the Settings fallback key), thread it through store_token / get_stored_token, and have routes that call cloud.set_token(token)
  also rebuild/reconfigure the base URL from the stored region. This at least survives restart and scopes per-user.
  - Or, drop the singleton pattern here and construct a BambuCloudService(region=...) per request from the stored region.

  While you're in there

  You flagged this in the description already: /cloud/login accepts request.region in the schema but the handler ignores it. Same class of bug, same fix — please handle
  both flows here so we don't ship half a region selector. That would make the password-login selector actually functional, not just a UI affordance.

  **Nits**

  - A route test confirming a China-region token request hits api.bambulab.cn (e.g. via an httpx mock) would be great.
  - region: str = Field(default="global", ...) → Literal["global", "china"] so invalid values fail at validation rather than silently falling into the China branch.
  - Please double-check profiles.login.region, regionGlobal, regionChina exist in all 7 locale files.

  Requesting changes for now. Happy to re-review as soon as persistence + scoping are sorted. Thanks again for jumping on this.

Minidoracat:
Thanks for the detailed review — noted on the issue-first workflow, will open one before the next contribution.

Pushed a follow-up commit addressing all blocking items and the nits:

**Persistence (blocking item 1)**
- New `users.cloud_region` column + `bambu_cloud_region` Settings fallback key, threaded through `store_token` / `get_stored_token` / `clear_token`.
- Legacy rows with NULL region normalise to `"global"` so existing tokens keep working after the migration.
- Verified with a real P1S setup: stored `cloud_region="china"` survives container restart and subsequent `/cloud/*` calls hit `api.bambulab.cn`.

**Singleton dropped (blocking items 2 and 3)**
- Removed `_cloud_service` / `get_cloud_service()` / `reset_cloud_service()`.
- Every cloud-touching route now builds a per-request `BambuCloudService(region=...)` via a new `build_authenticated_cloud(db, user)` helper, seeded from that user's stored token, and `await cloud.close()`s it in `finally`.
- No more cross-user region bleed; no more leaked `httpx.AsyncClient` on region switch.

**`/cloud/login` honours `request.region` (the "half a region selector" note)**
- The login handler used to ignore `request.region` even though the schema and frontend sent it. It now builds a region-aware service and stores the region on success.
- `/cloud/verify` gained a `region` field so TOTP reaches the correct TFA endpoint (`bambulab.com` vs `bambulab.cn`) without relying on singleton state carried over from `login_request`. Frontend `cloudVerify()` and `ProfilesPage` pass it through.

**Nits**
- `CloudLoginRequest` / `CloudVerifyRequest` / `CloudTokenRequest`: `region: Literal["global", "china"]`, so invalid values fail in Pydantic instead of silently routing to the China branch.
- Tests: unit tests assert `BambuCloudService(region="china")` routes login + TOTP to the `.cn` endpoints (with httpx mocked); integration tests cover region round-trip through `store_token` / `get_stored_token` / `build_authenticated_cloud` for both per-user and Settings-fallback paths.
- i18n keys (`profiles.login.region`, `regionGlobal`, `regionChina`) confirmed present in all 7 locale files.

`ruff check` / `ruff format` / `tsc --noEmit` / `eslint` all clean. Ready for another look.

maziggy (changes requested):
Thanks for the rework — singleton removal is clean, migration looks safe, and the storage round-trip coverage is solid.

  Before merge, a few things worth tightening:

  **Tests**
  Please add route-level tests that cover the actual region plumbing end-to-end. The service unit tests prove BambuCloudService(region="china") hits .cn, and the storage
  tests prove region persists, but nothing asserts that a POST /cloud/login {region: "china", ...} actually reaches api.bambulab.cn — which is the core bug this PR fixes.
   Similarly for /cloud/verify (TFA endpoint selection) and /cloud/token.

  **Small API polish**
  /cloud/status could include the active region in its response so the frontend can reflect "Connected (China)" after a reload without relying on local state. Minor, but
  it's the one piece of plumbing that doesn't round-trip yet.

  **Efficiency**
  - get_auth_status calls get_stored_token and then build_authenticated_cloud, which internally calls get_stored_token again. Worth collapsing so /cloud/status hits the
  DB once.
  - Now that each request constructs its own BambuCloudService, every call pays a fresh TCP + TLS handshake. Fine for cold paths like login/status, but
  /cloud/filament-info can chain many get_setting_detail calls and the connection pool no longer carries across requests. Not a blocker, but if you'd like to keep the old
   pool reuse, sharing a single httpx.AsyncClient at the app level and passing it into the service would preserve it without bringing back the region-bleed bug.

  Happy to re-review once the route-level tests land.

Minidoracat:
Thanks — addressed everything from the second round.

**Route-level tests**
- Three tests now exercise the routes end-to-end by patching `httpx.AsyncClient` at the class level and asserting on the captured URLs:
  - `POST /cloud/token {region: "china"}` → `get_user_profile` hits `api.bambulab.cn`, never `api.bambulab.com`.
  - `POST /cloud/login {region: "china"}` → `login_request` hits `api.bambulab.cn`.
  - `POST /cloud/verify {region: "china", tfa_key: ...}` → TOTP goes to `bambulab.cn/api/sign-in/tfa`.
- Two more tests cover the new `/cloud/status` payload (authenticated / unauthenticated).

**`/cloud/status` returns `region`**
- `CloudAuthStatus.region: Literal["global", "china"] | None` (only populated when authenticated).
- Frontend `ProfilesPage` renders a compact region badge next to "Connected as …" — see screenshot below.
image


**Single DB round-trip in `get_auth_status`**
- Collapsed — the handler reads token/email/region once and constructs the per-request service inline instead of going through `build_authenticated_cloud`, which would re-query.

**App-scoped shared `httpx.AsyncClient`**
- Lifespan registers a single `AsyncClient` via `set_shared_http_client(...)`.
- `BambuCloudService` picks it up automatically and skips `close()` when it doesn't own the client; scripts/tests that skip the lifespan still get a self-owned client, so the existing unit tests (which patch `cloud._client.post`) keep working.
- Only the transport pool is shared — region + token state stays per-instance, so the region-bleed bug does not regress.

`ruff check` / `ruff format` / `tsc --noEmit` / `eslint` all clean. Ready for another look.

maziggy (approved):

maziggy:
Thanks!

wintsa123:
Note that the option for personal pages has been removed from China's bambu official website
image

wintsa123:
@Minidoracat i cant get token url.
image

Minidoracat:
@wintsa123  For China-region token login, you can get the token from your browser cookies:

1. Open MakerWorld in your browser and make sure you are logged in.
2. Press `F12` to open Developer Tools.
3. Go to **Application** → **Cookies** → the MakerWorld domain.
4. Select the `token` cookie.
5. Copy the full value from **Cookie Value** at the bottom, then paste it into Bambuddy as the token.

Please keep the token private and do not paste it into the PR thread.

How to get the MakerWorld token

wintsa123:
I noticed that bambu cookie has settings for long and short tokens. It seems that using short tokens doesn't seem to be effective for a long period of time. Should the backend be changed to use refreshToken instead?

@Minidoracat

Minidoracat:
> I noticed that bambu cookie has settings for long and short tokens. It seems that using short tokens doesn't seem to be effective for a long period of time. Should the backend be changed to use refreshToken instead?
> 
> @Minidoracat

Good point, thanks for checking this.

I also checked the current Profiles token-login flow in Bambuddy. Right now it stores and sends the copied cookie value as `access_token` / `cloud_token`; it does not persist or use `refreshToken` for a server-side refresh flow.

In my current MakerWorld login session, the `refreshToken` cookie has the same value as the `token` cookie, so simply switching the backend from `token` to `refreshToken` would not change the actual token being sent.

I also tested this token-login flow over a longer period on my side, and it stayed logged in without being logged out quickly.

So I think we should keep this PR using the `token` cookie for now. If we can confirm a real refresh endpoint / renewal flow, or if someone can reproduce a case where `token` expires quickly while `refreshToken` is different and still valid, that would be a good follow-up improvement.

PR #1067, opened 2026-04-21
fix(virtual-printer): Tailscale cert provisioning — hardening + maintainer feedback

### LINKED ISSUE #701: [Feature]: Support Standardized Root Trust via Let's Encrypt / Tailscale
### Problem Description
Currently, each **Bambuddy** installation generates its own unique self-signed Root CA. Per the [Bambuddy Virtual Printer documentation](https://wiki.bambuddy.cool/features/virtual-printer), this certificate must be manually added to the certificate store of slicers like Bambu Studio or Orca Slicer.

The current architecture creates a "Single Trust" bottleneck:
* **The Slicer Limitation:** Slicer certificate stores typically handle only **one active Bambuddy CA file at a time**.
* **Domain Friction:** Users who interact with multiple Bambuddy instances (e.g., School, Work, and Home) must manually swap or hand-edit the `printer.cer` file in the slicer directory every time they switch environments.
* **Administrative Fatigue:** While modifying `C:\Program Files` is always a high-friction task, doing it repeatedly throughout the day as a user migrates between organizational domains is a significant barrier to use.

### Proposed Enhancement
Transition from per-installation self-signed certificates to a standardized trust model using **Let's Encrypt**. 

Under this model, a user performs a **one-time manual operation**: adding the Let's Encrypt Root CA to their slicer's CA store. Once this is done, any Bambuddy instance using a Let's Encrypt-issued certificate will be automatically trusted by the slicer without further file manipulation.

### Benefits
1.  **Seamless Roaming:** After the initial one-time setup, users can switch between any number of Bambuddy instances (Home, School, Work) without ever touching a `.cer` file again.
2.  **Reduced Admin Privileges:** Removes the need for recurring Administrator access to sensitive application folders (`C:\Program Files`) for certificate swapping.
3.  **Standardized Security:** Leverages industry-standard, automated certificate lifecycles rather than indefinite self-signed CAs.

### Proposed Implementation Path
1.  **Tailscale Integration (Primary):** Many users deploy Bambuddy within a Tailnet. Leveraging Tailscale’s built-in [HTTPS/Certs feature](https://tailscale.com/kb/1153/enabling-https) would allow Bambuddy to automatically fetch and renew Let's Encrypt certificates for each node without any additional configuration or separate ACME client.
2.  **Built-in ACME Client (Secondary):** For non-Tailscale users, provide an optional ACME client (e.g., via a lightweight Python library) to handle HTTP-01 or DNS-01 challenges for users with valid FQDNs.
3.  **Fallback Logic:** Retain the current self-signed method as a fallback for offline or air-gapped installations.

### Alternatives Considered

To further reduce friction, file PRs for Bambu Slicer and Orca Slicer to include the Let's Encrypt CA in their CA cert store.

### Feature Category

Other

### Priority

Would improve my workflow

### Mockups or Examples

_No response_

### Contribution

### Checklist

legend813:
## What this PR does

Adds automatic Tailscale TLS certificate provisioning for Virtual Printers. When Tailscale is installed and connected, Bambuddy obtains a Let's Encrypt certificate via `tailscale cert` and applies it to all VP TLS services (MQTTS, FTPS, Bind). SSDP then advertises the Tailscale FQDN instead of a local IP — so slicers connect via a hostname covered by a publicly trusted cert, eliminating manual CA installation.

Falls back silently to the existing self-signed cert when Tailscale is absent or provisioning fails.

---

## Documentation

Wiki PR: maziggy/bambuddy-wiki#13 — adds a new **Tailscale Certificate (Optional)** section to the Virtual Printer page covering:
- How the automatic cert provisioning works end-to-end
- The per-VP toggle and when to disable it (LAN slicers without Tailscale access)
- Docker setup: `/var/run/tailscale` socket bind-mount with annotated compose example
- Proxy-mode limitation explained
- Fallback behaviour table and two new troubleshooting entries

---

## Why `tailscale cert` instead of `tailscale serve`

The original design discussion considered `tailscale serve`. After deeper analysis, `tailscale serve` is **not viable** for this protocol stack:

- **MQTTS (port 8883)** — raw TLS over TCP; `tailscale serve` is an HTTPS reverse proxy and cannot terminate bare TLS sockets
- **FTPS (port 990)** — same reason; requires raw TLS termination, not HTTP proxying

`tailscale cert` provisions a cert that the VP TLS listener loads directly, which works for all three protocols. This is the only technically compatible path for the full VP stack.

---

### Major issues

| Issue | Fix |
|-------|-----|
| Proxy-mode TLS/SSDP mismatch | `_resolve_cert_and_advertise()` returns self-signed cert immediately in proxy mode; renewal loop not started for proxy VPs |
| LAN-slicer regression from auto-detect | Addressed in companion PR #1070 (`use_tailscale_cert` toggle) |
| Docker deployment gap | Docker socket hint logged; wiki PR #13 documents the volume-mount setup |

### Code-level fixes

| Issue | Fix |
|-------|-----|
| `_cancel_renewal_task` silences real bugs | `CancelledError` and `Exception` now split; unexpected errors logged as warnings |
| Missing `asyncio.TimeoutError` handling | `get_status()` and `provision_cert()` now catch `TimeoutError` and return safe defaults |
| Fire-and-forget restart task | Stored as `self._cert_restart_task`; cancelled in `stop_server()` + `stop_proxy()` |
| Restart failure kills renewal loop | `_restart_for_cert_renewal()` re-spawns the renewal loop in its except block |

---

### New: `TailscaleService` (`backend/app/services/virtual_printer/tailscale.py`)

- `get_status()` — runs `tailscale status --json` to get FQDN and Tailscale IPs; returns `TailscaleStatus(available=False)` on any error
- `provision_cert()` — runs `tailscale cert --cert-file … --key-file … <fqdn>`; validates FQDN, creates output directory, sets key permissions to `0o600`, verifies readability
- `cert_needs_renewal()` — checks expiry against a 30-day threshold and validates FQDN against cert SANs (RFC 4343 case-insensitive)
- `ensure_cert()` — combines renewal check + provisioning in one call
- Security: binary resolved to absolute path via `shutil.which()` to prevent PATH hijacking; `_SUBPROCESS_ENV` allowlist strips JWT keys, DB URLs, and SMTP passwords from subprocess environment; `asyncio.wait_for` timeouts (5 s for status, 60 s for cert provisioning)

### Changes to `CertificateService` (`certificate.py`)

- New `ts_cert_path` / `ts_key_path` properties pointing to per-VP Tailscale cert files
- New `use_tailscale_cert(fqdn, tailscale_svc)` method

### Changes to `VirtualPrinterInstance` (`manager.py`)

- New `_resolve_cert_and_advertise()` — proxy mode always uses self-signed; server mode tries Tailscale then falls back
- New `_cert_renewal_loop()` — daily background task; schedules restart via `_cert_restart_task` on renewal
- New `_cancel_renewal_task()` / `_cancel_restart_task()` helpers with proper exception logging
- `_restart_for_cert_renewal()` re-spawns the loop if restart fails

### New API endpoint

`GET /api/v1/virtual-printers/tailscale-status` — returns `TailscaleStatusResponse`; protected by `RequirePermissionIfAuthEnabled(SETTINGS_READ)`

### Frontend (separate PR #1070)

- Tailscale status card in `VirtualPrinterList`
- `ShieldCheck` badge in `VirtualPrinterCard` when active
- Per-VP `use_tailscale_cert` toggle in settings

---

## Fallback behaviour

| Scenario | Behaviour |
|----------|-----------| 
| `tailscale` binary not found | `get_status()` returns `available=False`; self-signed cert used |
| Daemon stopped / not logged in | `tailscale status` exits non-zero; self-signed cert used |
| HTTPS certs not enabled for tailnet | stderr parsed for disabled-certs pattern; actionable log points to admin DNS page; self-signed cert used |
| Rate-limited by Let's Encrypt | Warning logged with exit code + stderr; self-signed cert used |
| `tailscale cert` times out | `TimeoutError` caught; self-signed cert used |
| Cert files not readable (root-owned) | `os.access()` check after write; actionable `sudo chown` log; self-signed cert used |
| Any other exception | Caught at manager level; warning logged; self-signed cert used |
| Docker socket not mounted | Info log with volume-mount instruction |

---

## Cert auto-renewal

- Renewal check runs once per day while the VP is active
- Renews when fewer than 30 days remain or FQDN no longer matches cert SANs
- After successful renewal, the VP is restarted automatically — no manual intervention required
- On error the loop backs off 1 hour; `CancelledError` on VP stop exits cleanly
- Restart task tracked and cancelled cleanly if VP is deleted mid-restart

---

## Test plan

- [x] `pytest backend/tests/unit/services/test_tailscale.py -v` — 43 tests, all pass
- [x] `pytest backend/tests/unit/ -v` — no regressions
- [x] Frontend: `npm run build` — clean, no TypeScript errors
- [ ] Manual: start VP with Tailscale active → logs `"[VP …] Using Tailscale cert for <fqdn>"`
- [ ] Manual: `GET /api/v1/virtual-printers/tailscale-status` returns FQDN and `available: true`
- [ ] Manual: Tailscale disconnected → logs `"Tailscale not available, using self-signed cert"`
- [ ] Manual: LAN slicer + toggle Off → discovery works without Tailscale access
- [ ] Manual: Docker without socket mount → actionable log appears

Closes #701


legend813:
The toggle was implemented here: https://github.com/maziggy/bambuddy/pull/1070

maziggy:
Very nice PR!

maziggy (changes requested):
Thanks @legend813 — this is a really nice piece of work. The layering is clean (TailscaleService → CertificateService → VirtualPrinterInstance), every failure path falls
  back to self-signed with an explicit log line, and the security hygiene is tight: subprocess env allowlist, shutil.which for PATH-hijack defense, FQDN regex before passing to subprocess, 0o600 on the key, readability probe after write with an actionable chown log, and a specific stderr match for the HTTPS-disabled tailnet case that points at login.tailscale.com/admin/dns. 36 unit + 2 integration tests, and they cover the non-happy paths (OSError, ValueError, JSON parse failure, SAN mismatch, back-off, restart scheduling) — not just the sunny-day flow.

  A few things to discuss before we merge.

  Design questions

  1. Proxy-mode SSDP/TLS mismatch. In start_proxy() the third return from _resolve_cert_and_advertise() is discarded:

  cert_path, key_path, _ = await self._resolve_cert_and_advertise()

  SSDP then either forwards the real printer's announcements (preserving the target's IP in Location:) or falls back to _start_fallback_ssdp where advertise_ip=self.bind_ip.
  In both cases the slicer connects by IP, but TLS presents an LE cert whose SAN is only the Tailscale FQDN → name-validation failure. This only "works" if the slicer skips
  verification, in which case the LE cert provides no practical benefit over self-signed for proxy mode.

  I think the pragmatic choice is to scope Tailscale-cert usage to server mode only for this PR and leave proxy mode on self-signed. Otherwise the value prop is misleading in
  that path.

  2. LAN-slicer regression from auto-detect. In server mode SSDP now advertises the FQDN, which requires the slicer's machine to resolve it — i.e. Tailscale + MagicDNS on both ends. That's the intended case. But if Tailscale is running on the Bambuddy host and the slicer is on the plain LAN (no Tailscale), pre-PR discovery worked and post-PR it
  silently breaks. This argues for the per-VP use_tailscale_cert toggle to ship in this PR rather than as a follow-up — auto-detect as the only path makes the feature riskier
  than it needs to be.

  3. Docker. The "Bare-metal permissions note" covers the tailscale group on the host. Bambuddy is mostly deployed in Docker, and /var/run/tailscale/tailscaled.sock isn't
  inside the container by default — so tailscale status quietly returns "not found" and the feature never engages. Can we add a Docker section to the PR description / setup
  docs (socket bind-mount or --network=host) and maybe a one-shot log line at startup so Docker users see "Tailscale socket not reachable from container" rather than silence?

  Nits

  - _cancel_renewal_task catches (asyncio.CancelledError, Exception) silently (manager.py:380). CancelledError is expected; anything else is worth logging so real bugs don't disappear.
  - provision_cert doesn't catch asyncio.TimeoutError (tailscale.py:218) — only OSError. Timeouts propagate up and are ultimately caught by the blanket except Exception in
  _resolve_cert_and_advertise, but you lose the specific "provisioning timed out" signal in logs. A dedicated except asyncio.TimeoutError: return False with a targeted warning
   would be nicer.
  - _restart_for_cert_renewal is fire-and-forget (manager.py:425), not tracked anywhere. If the VP is deleted mid-restart the task continues against a stale instance. Unlikely
   in practice.
  - If stop_server()/start_server() fails inside the restart task, no new _cert_renewal_task is spawned → cert expiry won't be re-checked until the VP is manually restarted.
  Worth adding a retry or at least a loud warning.

  Test plan adds

  For manual testing I'd also verify:
  - Slicer on LAN without Tailscale while host has Tailscale: discovery still works (or is documented as "use self-signed mode"). This is the regression case from point 2.
  - Docker deployment: does the socket mount work end-to-end?
  - Cert renewal: if we bump TS_CERT_EXPIRY_THRESHOLD_DAYS high in a test build, does the daily check actually renew and restart services cleanly?

legend813:
Thanks for the thorough review! Here's a summary of what was addressed in the latest commit:

---

**Major Issue 1 — Proxy-mode TLS/SSDP mismatch**

Fixed. `_resolve_cert_and_advertise()` now has an early-exit guard for proxy mode that returns the self-signed cert immediately, skipping Tailscale entirely. The cert renewal loop is also no longer started in `start_proxy()` since `tailscale_fqdn` would always be `None` there.

**Major Issue 2 — LAN-slicer regression from auto-detect**

Addressed in companion PR #1070 (`feat(virtual-printer): add Tailscale opt-out toggle`) which adds the per-VP `use_tailscale_cert` toggle — already merged/open.

**Major Issue 3 — Docker deployment gap**

Two changes:
- `TailscaleService._log_docker_socket_hint()` now logs an actionable volume-mount instruction when `/.dockerenv` is present but the Tailscale socket is absent
- Wiki PR maziggy/bambuddy-wiki#13 documents the `/var/run/tailscale` bind-mount with a full annotated compose example

---

**Code-level fixes**

| Issue | Fix |
|-------|-----|
| `_cancel_renewal_task` catches all exceptions silently | Split into `except CancelledError` + `except Exception` — unexpected errors now logged as warnings |
| Missing `asyncio.TimeoutError` handling | `get_status()` and `provision_cert()` now each catch `TimeoutError` and return safe values instead of propagating |
| Fire-and-forget restart task | Stored as `self._cert_restart_task`; new `_cancel_restart_task()` helper called from both `stop_server()` and `stop_proxy()` |
| Restart failure kills renewal loop | `_restart_for_cert_renewal()` re-spawns `_cert_renewal_loop` in its `except` block so daily checks keep running after a failed restart |

7 new unit tests cover all of the above (43 total, all pass).

maziggy (changes requested):
Thanks @legend813 — second pass is very close. Every major and nit from the first review was addressed with code + matching tests, and the extras you added unprompted (HTTPS-disabled stderr match, actionable sudo chown log, /.dockerenv detection, OSError-wrapped chmods in CertificateService) are exactly the shape I wanted. The security hygiene is tight.

  Before I merge I need the following tidied up. I don't merge partially-done changes — everything should be in this PR, not deferred.

  1. _cancel_restart_task has the same silencing bug _cancel_renewal_task did

  In manager.py:

  async def _cancel_restart_task(self) -> None:
      if self._cert_restart_task and not self._cert_restart_task.done():
          self._cert_restart_task.cancel()
          try:
              await self._cert_restart_task
          except (asyncio.CancelledError, Exception):  # ← silences real bugs
              pass
      self._cert_restart_task = None

  Same split as you did on _cancel_renewal_task:

      pass
  except Exception as e:
      logger.warning("[VP %s] Unexpected error in cert restart task: %s", self.name, e)

  2. Fold the use_tailscale_cert toggle (PR #1070) into this PR

  I need #1067 to be self-consistent as a single merge. Auto-advertising the Tailscale FQDN without a per-VP opt-out breaks LAN-only slicers on hosts where Tailscale happens to be running — that's the regression I flagged in my first review, and I don't want it land-able in any intermediate state. Please close #1070 and bring its frontend +
  backend changes into this branch.

  3. Docker hint log spam

  TailscaleService._log_docker_socket_hint() fires on every get_status() call, which means every VP startup and every /tailscale-status poll from the UI card. Once the
  Tailscale status card is live the same "Docker environment detected…" info line will repeat indefinitely. A once-per-process guard (class-level _docker_hint_logged = False
  flag set on first log) fixes it cleanly.

  4. _restart_for_cert_renewal doesn't cover the CancelledError path

  Your except Exception: block re-spawns _cert_renewal_loop if stop_server()/start_server() raised, which covers the case from my earlier nit. But if an external stop_server() call fires _cancel_restart_task() while this is mid-flight, the restart task raises CancelledError — not a subclass of Exception — and the re-spawn doesn't run. A small finally: with a state check, or an explicit except asyncio.CancelledError: raise above the Exception branch, closes the gap.

Once those four are in I'll merge #1067 and we can close #1070 as superseded. Really nice piece of work overall.

legend813:
Can be closed as everything will be handled within #1070

Issue #1008, opened 2026-04-17
[Feature]: Purge out old files from file manager


cadtoolbox:
### Problem or Use Case

With lots of users and lots of prints the files can become plentiful making backup and restores take a long time.

### Proposed Solution

Add ability for admins to purge out older files that have not been re-printed in X number of days.

### Feature Category

File Manager

### Priority

Nice to have

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Before I build it, a few questions so it lands the way you actually need:

  1. Manual or scheduled (or both)?
    - Manual: an admin "Purge old files" button with a confirmation dialog showing count + disk space to be freed.
    - Scheduled: runs daily/weekly in the background, same way the new scheduled local-backup works.
    - Both is also fine but adds some UI.
  
2. What counts as "old"? Just a day threshold (e.g. "not re-printed in 90 days"), or would you also want per-file-size / total-library-size triggers (e.g. "keep library under 10 GB, drop oldest first")?

  3. Files never printed — some files get uploaded and never printed at all. Should those be included once they cross the age threshold, or left alone?

  4. External files (3MF files linked from a network path rather than stored locally) — I'd skip these since they don't take disk space. Any reason not to?

  5. Safety net — do you want purged files to go to a soft-delete / trash state first for a grace period, or direct hard-delete? Soft-delete is safer but uses the same diskspace until the grace period ends, so it doesn't fully solve the backup-size problem unless the grace period is short.

  6. Per-user vs global — should each user's files be purged independently based on their last-print time, or is this purely an admin-wide operation?

cadtoolbox:
1.) I would probably say manual would be sufficient and the safer route. 

2.) If it's manual, maybe allow the admin to pick "Older than XX number of days"

3.) Yes

4.) Agree

5.) A trash bin approach is fine, after X number of days it then deletes.

6.) I say an admin approach, but I don't try to use Bambuddy as a repository, it's a tool to print.

maziggy:
Available/Fixed in branch dev and available with the next release or daily build. Please let me know if it works for you.

Docs -> https://wiki.bambuddy.cool/features/file-manager/?h=purge#purge-old-files-admin

[...]

PR #1070, opened 2026-04-21
feat(virtual-printer): add Tailscale opt-out toggle (closes #701 point 3)

### LINKED ISSUE #701: [Feature]: Support Standardized Root Trust via Let's Encrypt / Tailscale

legend813:
## Summary

Closes the open point from #701 PR review — item 3: **opt-in vs auto-detect**.

Implements auto-detect with a user-facing toggle to disable Tailscale integration per virtual printer. The default behavior (auto-detect) is unchanged for all existing users.

- **Toggle OFF (yellow):** Tailscale is never queried; VP starts immediately with the self-signed cert. Useful for users on company/school tailnets where HTTPS certs are disabled, or anyone who simply prefers the self-signed flow.
- **Toggle ON (green, default):** Auto-detect as before — provisions a Let's Encrypt cert via `tailscale cert` when Tailscale is available, falls back to self-signed otherwise.

Toggling while the VP is running triggers an automatic restart via the existing `sync_from_db()` change detection.

## Changes

| Layer | What changed |
|---|---|
| DB | `tailscale_disabled BOOLEAN DEFAULT 0` column on `virtual_printers` + runtime migration |
| Backend | `VirtualPrinterInstance.__init__` accepts `tailscale_disabled`; `_resolve_cert_and_advertise()` skips Tailscale entirely when set; change-detection loop triggers restart on toggle flip |
| API | `VirtualPrinterUpdate` + `_vp_to_dict` expose `tailscale_disabled`; `/settings/virtual-printer` GET/PUT also handle it and propagate to the first VP row |
| Frontend | `VirtualPrinterSettings.tsx` toggle (green=auto/yellow=disabled); `VirtualPrinterSettings` interface updated; `updateSettings` accepts `tailscale_disabled` |
| i18n | `disableTitle`, `enabledHint`, `disabledHint` keys added to all 8 locales |
| Tests | 2 new unit tests: disabled skips Tailscale query; enabled default queries Tailscale — 38 total passing |

## Test plan

- [ ] With Tailscale installed and connected: toggle OFF → VP starts with self-signed cert; toggle ON → VP provisions LE cert
- [ ] With Tailscale absent: toggle has no visible effect on cert (still self-signed either way), but log line differs
- [ ] Toggle while VP is running → VP restarts automatically, picks up new cert path
- [ ] Existing users on upgrade: `tailscale_disabled` defaults to `false` — no behavior change
- [ ] Unit tests: `python -m pytest unit/services/test_tailscale.py -q` → 38 passed


maziggy (changes requested):
Thanks @legend813. I've reviewed this on top of my pending feedback on #1067 — since #1070 contains all of #1067's commits plus the toggle, the cleanest path forward is to continue in this PR and close #1067 as superseded. Please close #1067 and push the consolidated fixes here.

Six items to address before I merge. No deferred follow-ups — everything lands in this PR.

  From my second review of #1067

  1. _cancel_restart_task has the same silencing bug _cancel_renewal_task did. In manager.py:

  async def _cancel_restart_task(self) -> None:
      if self._cert_restart_task and not self._cert_restart_task.done():
          self._cert_restart_task.cancel()
          try:
              await self._cert_restart_task
          except (asyncio.CancelledError, Exception):  # ← silences real bugs
              pass
      self._cert_restart_task = None

  Same split as you did on _cancel_renewal_task:

      pass
  except Exception as e:
      logger.warning("[VP %s] Unexpected error in cert restart task: %s", self.name, e)

  2. Docker hint log spam. TailscaleService._log_docker_socket_hint() fires on every get_status() call — every VP startup plus every /tailscale-status poll from the UI card.
  Once the Tailscale status card is live the info line repeats indefinitely. Fix with a once-per-process class-level flag:

  class TailscaleService:
      _docker_hint_logged = False

      @classmethod
      def _log_docker_socket_hint(cls) -> None:
          if cls._docker_hint_logged:
              return
          if Path("/.dockerenv").exists() and not Path("/var/run/tailscale/tailscaled.sock").exists():
              logger.info(...)
              cls._docker_hint_logged = True

  3. _restart_for_cert_renewal doesn't cover the CancelledError path. Your except Exception: block re-spawns _cert_renewal_loop if stop_server()/start_server() raised — covers my earlier nit. But if an external stop_server() calls _cancel_restart_task() while this is mid-flight, the restart task raises CancelledError — not a subclass of Exception
   — and the re-spawn doesn't run. A small finally: with a state check, or an explicit except asyncio.CancelledError: raise above the Exception branch, closes the gap.

  New findings from #1070

  4. The tailscale_disabled toggle is not accessible to users. It's in VirtualPrinterSettings.tsx, which is not mounted anywhere in the active UI — SettingsPage.tsx renders <VirtualPrinterList />, which in turn renders <VirtualPrinterCard /> per VP. That's where the existing per-VP toggles live (enabled at line 107, auto_dispatch at line 302, remote_interface_ip at line 173). The toggle needs to move to VirtualPrinterCard.tsx and use the existing updateMutation / pendingAction plumbing. Remove the additions from VirtualPrinterSettings.tsx — that file is dead code and shouldn't be extended.

  5. No test for sync_from_db restart-on-toggle-flip. The whole point of the toggle is that flipping it at runtime triggers a VP restart. The change-detection line in
  sync_from_db():

  or instance.tailscale_disabled != vp.tailscale_disabled

  …has zero test coverage. The two new unit tests only exercise _resolve_cert_and_advertise on a freshly-constructed instance. A sync_from_db-level test (existing instance with tailscale_disabled=False, flip the DB row, call sync_from_db, assert the instance was restarted with the new value) would lock it in.

  6. No frontend test for the toggle. Once the toggle moves to VirtualPrinterCard.tsx (per item 4), add a test to VirtualPrinterCard.test.tsx matching the shape of the existing auto-dispatch toggle test: renders with correct initial state for tailscale_disabled true/false, click calls updateMutation with the right payload, pendingAction gates further clicks.

  ---
  Once those six are in I'll merge #1070 and we'll close #1067 as superseded. Really nice work overall — the backend hardening, security hygiene, and test coverage on the
  Tailscale service itself are exactly the shape I wanted. The gaps above are all tractable.

legend813:
All six items addressed in this push. Here's what changed:

**Items 1–3 (from #1067 review):**

1. `_cancel_renewal_task` — split `except (asyncio.CancelledError, Exception)` into two separate branches: `CancelledError` silently passes, `Exception` logs a warning. Same fix applied to the new `_cancel_restart_task` method.

2. `_cancel_restart_task` added — `_cert_restart_task` is now tracked on the instance (was fire-and-forget via `asyncio.create_task`). `_cancel_restart_task` is called from both `stop_server` and `stop_proxy` so the task is properly cancelled on shutdown.

3. `_restart_for_cert_renewal` — added `except asyncio.CancelledError: raise` above the `Exception` branch so external cancellation propagates correctly instead of being swallowed.

**Items 4–6 (from #1070 review):**

4. `tailscale_disabled` toggle moved from `VirtualPrinterSettings.tsx` (dead code path) to `VirtualPrinterCard.tsx`, following the existing `auto_dispatch` toggle pattern exactly — `localTailscaleDisabled` state, `useEffect` sync, `pendingAction='tailscaleDisabled'`, `updateMutation.mutate({ tailscale_disabled: newVal })`. Added `tailscale_disabled` to `VirtualPrinterConfig` and `multiVirtualPrinterApi.update()` in `client.ts`. `VirtualPrinterSettings.tsx` additions fully removed.

5. `sync_from_db` test added — creates an instance with `tailscale_disabled=False`, flips the DB row to `True`, calls `sync_from_db`, asserts `remove_instance` is called. Also added `tailscale_disabled: False` to `_make_db_vp` defaults.

6. `VirtualPrinterCard.test.tsx` — new describe block with 3 tests: renders green when `tailscale_disabled=false`, renders grey when `tailscale_disabled=true`, click triggers `multiVirtualPrinterApi.update(1, { tailscale_disabled: true })`.

Also added `_log_docker_socket_hint` as a classmethod with `_docker_hint_logged: bool = False` class-level flag — called from `get_status()` when the binary is not found.

All 139 backend tests and 8 frontend card tests pass.

maziggy (changes requested):
Thanks for the polish on this — overall really clean work. The subprocess env allow-list, FQDN validation, SAN-vs-FQDN renewal check, chmod fallbacks, and the 759-line
  tailscale test file are all well-considered. Two real issues to address before merge, plus a couple of smaller items.

  Blockers

  1. BOOLEAN DEFAULT 0 will silently fail on PostgreSQL

  backend/app/core/database.py:484:
  await _safe_execute(conn, "ALTER TABLE virtual_printers ADD COLUMN tailscale_disabled BOOLEAN DEFAULT 0")

  PostgreSQL rejects DEFAULT 0 for a BOOLEAN column (column is of type boolean but default expression is of type integer). _safe_execute swallows the error, so on a fresh PG
  install the column never gets created and any query over virtual_printers then 500s. We hit this exact pattern on the trash PR (#1008) and it broke the file manager in prod.
   The model's server_default="false" doesn't help because it only applies to CREATE TABLE, not ALTER.

  Please branch on the dialect, matching the rest of this file:
  if is_sqlite():
      await _safe_execute(conn, "ALTER TABLE virtual_printers ADD COLUMN tailscale_disabled BOOLEAN DEFAULT 0")
  else:
      await _safe_execute(conn, "ALTER TABLE virtual_printers ADD COLUMN tailscale_disabled BOOLEAN DEFAULT false")

  2. _cancel_restart_task deadlocks when called from inside the restart task

  Flow:
  - _cert_renewal_loop detects renewal is needed, creates _cert_restart_task = asyncio.create_task(_restart_for_cert_renewal()), then breaks.
  - _restart_for_cert_renewal runs, calls stop_server / stop_proxy.
  - stop_server calls await self._cancel_restart_task().
  - At that point self._cert_restart_task is the currently running task, so we cancel() and then await ourselves — Task cannot await on itself (RuntimeError, or in some loop
  states, hangs).

  CI won't catch this because it only fires on an actual renewal day (roughly every 60 days with a 30-day threshold). Suggested fix in _cancel_restart_task:

  task = self._cert_restart_task
  if task and not task.done() and task is not asyncio.current_task():
      task.cancel()
      try:
          await task
          pass
  self._cert_restart_task = None

  A direct async unit test that calls _restart_for_cert_renewal() and asserts completion would lock this down.

  Smaller items

  - Unused i18n keys. virtualPrinter.tailscale.disableTitle, enabledHint, and disabledHint were added to all 8 locales but the components only reference
  virtualPrinter.tailscaleDisabled.{title,description} and virtualPrinter.tailscale.{connected,notAvailable,trustedCert,notAvailableHint}. Please wire them up or drop them to avoid locale drift.
  - Toggle colour. The PR description says "Toggle OFF (yellow)" but the code uses bg-bambu-dark-tertiary (dark grey) when disabled. Not a blocker — just reconcile description vs. code.
  - use_tailscale_cert(..., tailscale_svc: object). Worth tightening to TailscaleService (or a small Protocol) so the signature documents the contract.
- The daily cert-renewal path does a full stop_server/start_server, which drops any in-flight slicer upload or MQTT session on that VP. Once per ~60 days so impact is low,
  but reloading the TLS context on the listeners instead would be a nicer follow-up.
  - The /settings/virtual-printer PUT route only propagates tailscale_disabled to the first VP row — consistent with how that legacy endpoint treats every other field, so
  fine, just worth a one-line note in the code.

Thanks again!

maziggy:
Lemme fix the remaining issue. Would like to merge it.

Thanks again for your great work! 💯

legend813:
All second-round review items addressed — here's what changed:

**Self-deadlock fix in `_cancel_restart_task`**
Added `task is not asyncio.current_task()` guard so `_restart_for_cert_renewal` (which runs *as* `_cert_restart_task`) can call `stop_server → _cancel_restart_task` without deadlocking on itself. Added a unit test (`test_restart_for_cert_renewal_no_self_deadlock`) that exercises this path.

**`_restart_for_cert_renewal` — explicit `CancelledError` re-raise**
Moved to `except asyncio.CancelledError: raise` before the `Exception` branch so external cancellation propagates correctly.

**PostgreSQL `BOOLEAN DEFAULT` fix in `database.py`**
Branched on `is_sqlite()`: SQLite gets `DEFAULT 0`, PostgreSQL gets `DEFAULT false`.

**`certificate.py` type annotation tightened**
`use_tailscale_cert` now uses the proper `TailscaleService` type (via `TYPE_CHECKING`) instead of `object`.

**`settings.py` comment expanded**
Explains the legacy first-VP propagation pattern inline.

**Orphaned i18n keys removed**
Dropped `tailscale.disableTitle`, `tailscale.enabledHint`, `tailscale.disabledHint` from all 8 locale files.

legend813:
Haha, already finished the work but didn't push it yet...

maziggy:
Oh, sorry!

PR #1160, opened 2026-04-28
feat(backup): Extend Backup to other Git providers

### LINKED ISSUE #576: [Feature]: Support self-hosted Git platforms (Gitea / GitLab) for repository backups
### Problem or Use Case

Currently, Bambuddy validates repository URLs strictly against the GitHub format: `https://github.com/owner/repo`

The validation rejects any other Git hosting platform URLs, including self-hosted Git services such as:
	•	Gitea
	•	GitLab (self-hosted)
	•	Forgejo
	•	Enterprise GitHub instances

This prevents users running local infrastructure from using Bambuddy with their internal Git platforms.
When attempting to use a self-hosted Gitea repository:
```
Failed to save: [{
  "type":"value_error",
  "loc":["body","repository_url"],
  "msg":"Value error, Invalid GitHub repository URL. Expected format: https://github.com/owner/repo",
  "input":"http://gitea.lan:3000/gitea/bambuddy",
  "ctx":{"error":{}}
}]
```

The validation logic appears to explicitly enforce github.com:
https://github.com/maziggy/bambuddy/blob/662cedef5d52608a1286806896126210a5b3ed69/backend/app/schemas/github_backup.py#L36-L46

### Proposed Solution

Instead of validating strictly against github.com, consider:
	•	Validating against a configurable base URL
	•	Or validating only the /owner/repo pattern
	•	Or adding a selectable provider type (GitHub / GitLab / Gitea)

### Alternatives Considered

_No response_

### Feature Category

Other

### Priority

Nice to have

### Mockups or Examples

_No response_

### Contribution

### Checklist

BurntOutHylian:
## Description

Extends the Git backup feature to support GitHub, Gitea, Forgejo, and GitLab as backup targets.
Previously the feature was hard-wired to github.com (URL validator rejected all non-github.com URLs
and all API calls used api.github.com). Users running self-hosted Gitea, Forgejo, or GitLab
instances could not use the feature.

## Key changes

- Provider dropdown in the backup settings UI (GitHub / GitLab / Gitea / Forgejo)
- An "Allow Insecure HTTP" option for users who have locally hostet Git repo
- Provider abstraction layer in the backend — each provider is its own class behind a factory
- URL validator updated to account for non-GitHUB URLs, self-hosted URL formatting, and insecure URLs (if allow insecure is checked)
- Database migration adds a provider column (default github) — safe to replay on existing installs
- Database migration adds a allow_insecure_http column (default false) — safe to replay on existing installs
- GitLab uses the GitLab REST API v4 (different from GitHub's Git Data API)
- All 8 locale files updated with provider labels and URL placeholders

## Related Issue

Fixes #576

## Documentation

Companion docs PRs:

- Wiki: maziggy/bambuddy-wiki#21

- Docs PR(s) linked above

## Type of Change

- New feature (non-breaking change that adds functionality)
 
## Changes Made

- backend/app/core/database.py — migration adds provider VARCHAR(30) DEFAULT 'github' and allow_insecure_http BOOL DEFAULT false
- backend/app/models/github_backup.py — provider mapped column
- backend/app/schemas/github_backup.py — ProviderType StrEnum, relaxed URL regex
- backend/app/services/git_providers/ — new package: base ABC, GitHub, Gitea, Forgejo, GitLab backends, factory
- backend/app/services/github_backup.py — delegates push/test to provider backend
- backend/app/api/routes/github_backup.py — reads/writes provider field
- frontend/src/api/client.ts — GitProviderType union type, updated interfaces
- frontend/src/components/GitHubBackupSettings.tsx — provider dropdown, provider-aware placeholders
- frontend/src/i18n/locales/*.ts — new keys in all 8 locale files
- backend/tests/unit/test_git_providers.py — provider factory and URL parsing tests
- backend/tests/unit/test_github_backup_schemas.py — schema validation tests
- frontend/src/__tests__/components/GitHubBackupSettings.provider.test.tsx — UI tests

### Before

Before

### After

After

### After - Selection List

After_DropDown

### After - Edge Case: Alert when using HTTP instead of HTTPS

After_EdgeCase

## Testing

- I have tested this on my local machine
- I have tested this as a user with an existing GitHub backup (to ensure no breaking change)
- I have tested this with GitHub
- I have tested this with GitLab
- I have tested this with a local Gitea instance
- I have tested this with a local Forgejo instance
- I have tested with my printer model: BambuLab P2S

Backend unit tests:

- pytest backend/tests/unit/test_git_providers.py — factory, URL parsing, API base derivation
- pytest backend/tests/unit/test_github_backup_schemas.py — schema validation, provider enum

Frontend tests:

- npm run test:run — provider dropdown renders, loads from config, Git options present
- UI interactive testing

## Checklist

- [x] ./test_frontend.sh passed
- [x] venv/bin/ruff check backend passed
- [x] venv/bin/ruff format --check passed for modified backend files
- [x] ../venv/bin/python3 -m pytest tests/ -v -n 30 passed with 3370 passed I have tested my changes thoroughly

## Additional Notes

The implementation drops api_base_url relative to the original issue discussion: Gitea and Forgejo are their own distinct providers in case Gitea and Forgejo diverge in their APIs over time.

With the abstraction layer, adding/removing hosts should allow us to minimize the impact of future add/remove changes


maziggy (changes requested):
Solid foundation -- the abstraction layer is well-shaped and the test coverage is good (38 backend + 9 frontend tests, all green; ruff clean; i18n parity holds). A few issues to fix before merge.

  Blocker

  B1 -- BOOLEAN DEFAULT 0 is SQLite-only syntax. backend/app/core/database.py:1789:

  await _safe_execute(conn, "ALTER TABLE github_backup_config ADD COLUMN allow_insecure_http BOOLEAN DEFAULT 0")

  PostgreSQL rejects this with column is of type boolean but default expression is of type integer, and _safe_execute only swallows idempotency errors -- it would re-raise this and abort startup for any Postgres user upgrading to this version. Match the pattern at database.py:632-635 (the Tailscale migration):

  if is_sqlite():
      await _safe_execute(conn, "ALTER TABLE github_backup_config ADD COLUMN allow_insecure_http BOOLEAN DEFAULT 0")
  else:
      await _safe_execute(conn, "ALTER TABLE github_backup_config ADD COLUMN allow_insecure_http BOOLEAN DEFAULT false")

  The provider VARCHAR(30) DEFAULT 'github' migration (line 1786) is fine on both dialects.

  High -- correctness bugs that surface in real usage

  H1 -- GitLab subgroup URLs are silently rejected. gitlab.py:33-38's regex matches only host/owner/repo, but GitLab's defining feature is nested groups:
  gitlab.com/group/subgroup/project, gitlab.com/myorg/team/api/backend, etc. The PR description says "tested with GitLab," but only flat-namespace projects work. Self-hosted GitLab without subgroups has the same constraint. Fix the regex to capture the full namespace path and URL-encode it for the /api/v4/projects/{encoded} calls (which is what GitLab expects).

  H2 -- Gitea/Forgejo create one commit per file. gitea.py:87-147 loops through each file and PUT/POSTs to the contents API, which auto-commits per call. A backup of N files =
   N commits in the user's repo. With 5 connected printers x 4 nozzle profiles + cloud profiles + spools + archives, that's easily 30+ commits per backup -- the repo's history
   becomes unreadable noise. Gitea exposes the same git/blobs, git/trees, git/commits Git Data API as GitHub, so the cleaner fix is having GiteaBackend inherit push_files from
   GitHubBackend instead of re-implementing it on the contents API.

  H3 -- GitLab tree pagination caps at 100 files. gitlab.py:147-151 requests per_page=100 and reads only that page. For repos with >100 files, paths beyond page 1 aren't in existing_blobs, so they're always queued as create (line 169) -- but GitLab's commits API will reject a create action for an already-existing path. Bambuddy backups easily exceed 100 files (each printer x each nozzle = 4 K-profile files alone, plus archives can be hundreds of rows). Either follow Link: next headers or paginate via page=N until
   empty.

  Medium -- defense-in-depth

  M1 -- provider query param on /test not validated. routes/github_backup.py:210 accepts provider: str = Query(...) -- an invalid value (e.g. provider=bitbucket) triggers
  ValueError from the factory which becomes HTTP 500. Use provider: ProviderType = Query(...) to get a proper 422 with a clean error message.

  M2 -- GitHub backend silently ignores non-github.com hostnames. github.py:19-20 hard-codes https://api.github.com regardless of the input URL, but parse_repo_url accepts any host. So a user who selects provider=github but pastes a self-hosted GitHub Enterprise URL will get API calls to api.github.com (which won't have their repo), with a confusing 404 instead of a clear "wrong provider" error. Either reject non-github.com hosts when provider=github, or compute api_base from the URL host (GHE uses <host>/api/v3).

  Low / nits

  - L1 github_backup.py:169 hardcodes "Pushing to GitHub..." -- should reflect the provider name now that it's not always GitHub.
  - L2 git_providers/__init__.py is empty -- fine functionally, but consider exporting get_provider_backend and GitProviderBackend for cleaner imports.
  - L3 _calculate_next_run is single-underscore but called externally from the route (routes/github_backup.py:104, 128, 174). Not a bug -- just rename to calculate_next_run
  since it's de facto public.
  - L4 Error messages embed full response.text from the upstream API (gitlab.py:186, github.py:179, 189, 199). API responses can be sizeable on tree-list endpoints; truncate
  to ~200 chars to keep error logs / UI sane.
  - L5 Belt-and-suspenders getattr(config, "provider", "github") or "github" (github_backup.py:524, routes:37, 233) -- once the migration runs, config.provider is always a non-null string. The fallback is harmless but makes the code read like the column might be missing.

  Test gaps

  - No test exercises subgroup-style GitLab URLs (would catch H1).
  - No test exercises >100 files in the GitLab tree response (would catch H3).
  - No test asserts that the GitHub backend is rejected for non-github.com hosts (would catch M2).
  - Schema Update validator is tested for happy paths; an end-to-end test for the PATCH route's "stored http URL + flip allow_insecure_http to false" interaction would lock   down that policy.

BurntOutHylian:
A few fixes and a few questions:

### B1 - QUESTION
I could go with a 1 line fix instead considering Bambuddy's requirement for Python is 3.10+.

SQLite 3.23.0+ supports the syntax BOOLEAN DEFAULT FALSE, so I could just switch it to:

await _safe_execute(conn, "ALTER TABLE github_backup_config ADD COLUMN allow_insecure_http BOOLEAN DEFAULT FALSE")

SQLite 3.23.0 was released on 2018-04-02, and the version bundled with Python 3.10 (if not using system SQLite) is 3.27+.

I'm happy to use the logic structure, but I wanted to at least point this out as it could reduce long-term complexity in the migrations. I also have an out-of-scope question below this might help with.

The current code in my feat branch has been changed to use the one line at the moment and tested with both SQLite 3.53 and Postgres 15.12

### H1 - FIXED (test added)
Right! Total blind spot there. Was able to replicate on Gitlab.com and the docker I set up for this.

### H2 - FIXED
Removed push_files, _create_branch, and _extract_commit_sha (and unused imports). Inherited push_files resolves all URLs through self.get_api_base and self.get_headers() so it can route correctly automatically.

Cleaned up test_git_providers to switch the Contents-API tests to Git Data API tests (asserting exactly 1 POST .../git/commits regardless of file count).

### H3 - FIXED (test added)
Single GET /repository/tree replaced with an incremental loop until response returns empty list. Non-200 response also breaks loop to avoid crashing from partial failures

Test added to exercise 120 file return

### M1 - FIXED
provider=bitbucket example now returns a 422 with standard enumeration validation message instead of a 500.

### M2 - FIXED (test added)
get_api_base no parses the host from the URL. github.com → https://api.github.com. anything else → https://{host}/api/v3 (the GHE API path).

SSH URLs are handled the same way (git@ghe.example.com:… → https://ghe.example.com/api/v3).

Gitea and Forgejo override get_api_base themselves, so they're unaffected.

### L1 - DONE
Modified to reflect selected provider name

### L2 - DONE
from backend.app.services.git_providers import get_provider_backend can now be called

### L3 - DONE
Sure thing

### L4 - DONE
Yikes - good call. I added a _truncated_response_text helper to the base.py and routed the upstream error messages through it. Since Gitea and jorgejo inherit push_files() and the failure handling from GitHubBackend, those will also be truncated.

### L5
Tightened up. Replaced the three getattr(config, "provider", "github") or "github" fallbacks with direct config.provider access in the routes (and github_backup.py)

### Test Gaps
- Test added to cover H1
- Test added to cover H3
- Test added to cover M2
- Added PATCH route regression. The new test creates a stored HTTP config with allow_insecure_http: true, then PATCHes only allow_insecure_http: false. It asserts the route returns 422 and confirms the stored config still has the HTTP URL and allow_insecure_http: true.

## Out of scope finding
This may be an irrelevant concern. It is out of scope for this feature, but I thought I'd share/ask.

While I was looking into B1, I did see a bunch of BOOLEAN DEFAULT 0 in lines 381-608 with no corresponding PostgreSQL syntax lines. Like line 387  await _safe_execute(conn, "ALTER TABLE smart_plugs ADD COLUMN auto_off_executed BOOLEAN DEFAULT 0")

maziggy (changes requested):
Round 2 — thanks for the thorough turnaround. Verified the changes locally on top of the latest dev merge: ruff clean, full backend suite (3618 / 3618), frontend build,
  vitest, and the i18n parity script all pass.

  What landed:

  - B1 — BOOLEAN DEFAULT FALSE works on both SQLite 3.23+ (Python 3.10's bundled sqlite is 3.27+) and Postgres. Different from the is_sqlite() branch I suggested, but
  functionally correct — accepted.
  - H1 — subgroup-friendly parse_repo_url + URL-encoded namespace. The wire-format test asserting group%2Fsubgroup%2Fproject lands in the GET URL is a clean pin.
  - H2 — GiteaBackend(GitHubBackend) inheriting push_files is exactly the right shape. The "exactly 1 POST to git/commits regardless of file count" assertion locks the
  contract.
  - H3 — pagination loop with empty-list and non-200 termination. 120-blob across-two-pages test covers it.
  - M1 — ProviderType on the Query returns a clean 422.
  - M2 — host-derived get_api_base, tests cover github.com / GHE / port / SSH.
  - L2 / L3 / L5 — all done as described.
  - L4 — the _truncated_response_text helper is a nicer factoring than truncating at each call site; consistently applied across both backends.
  - Bonus: the PATCH route HTTP-regression test on top of the schema-level integration coverage is a good defensive add.

  One thing still open:

  L1 is not actually fixed. backend/app/services/github_backup.py:169 still has:

  self._backup_progress = "Pushing to GitHub..."

  The provider dropdown / placeholders / token hints on the frontend are all provider-aware now, but the progress string the UI polls and renders during a running backup is
  still hardcoded "Pushing to GitHub..." regardless of provider — so a user running a GitLab backup will literally see "Pushing to GitHub..." in the progress indicator.

  Easiest fix is one line — derive the label from config.provider. Something like:

  self._backup_progress = f"Pushing to {config.provider.title()}..."

  Or a tiny dict if you want nicer casing ("GitHub" / "GitLab" / "Gitea" / "Forgejo"). Either is fine.

  Two minor issue — fold in or leave for a follow-up, your call:

  - N1 — _push_to_github at github_backup.py:522 is now provider-agnostic in behaviour but still named after GitHub. _push_to_provider would round out the rename pass.
  - N2 — There's no direct schema-level unit test for repository_url=http://... without allow_insecure_http=true raising ValidationError on GitHubBackupConfigCreate. The PATCH
   integration test pins the policy end-to-end so coverage is fine; a unit test on validate_repo_url would just localise any future regression to that file.

BurntOutHylian:
on it.

### L1 - Actually Fixed 
Added a dict (felt cleaner). Also put a fallback in place in the event a future provider is added to the factory, but not the dict

```
provider_name = _PROVIDER_DISPLAY_NAMES.get(config.provider, config.provider)
self._backup_progress = f"Pushing to {provider_name}..."
```

### N1
oops. More coffee next time. Renamed to **_push_to_provider**

### N2
Added 2 tests:
- one to check for an unknown provider so any accidental removal or rewording is caught, 
- one to check for HTTP without Allow insecure HTTP

BurntOutHylian:
Also - updated the initial PR comment to include the PR for the wiki updates. Not sure why I forgot that part. maziggy/bambuddy-wiki#21

PR #1176, opened 2026-04-30
fix(oidc): use preferred_username/name claim for auto-created username (#1173)

### LINKED ISSUE #1173: [Bug]: SSO/OIDC Account creation behavior
### Component

Bambuddy

### Bug Description

I was able to successfully integrate the OIDC connection via Authentik. However, I noticed that when Bambuddy creates an account for a user who does not yet exist, it assigns an ID as the name rather than username which comes from the IdP. The new user is automatically assigned to the “Viewers” group. Changing the permissions is only possible once an email address has also been added.

It might make sense to use the attributes from the IdP, such as preferred_username and email, and store the OIDC ID separately.

### Expected Behavior

New Account creation listed with username and email with groups synced to their group-permissions

### Steps to Reproduce

1. Go to oidc config
2. set Auto-create users to true
3. login with new user which isn't registered on bambuddy

### Printer Model

None

### Bambuddy Version

v0.2.3.2

### SpoolBuddy Version

_No response_

### Printer Firmware Version

_No response_

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

### Checklist

netscout2001:
## Description

Addresses two issues from #1173:

### Thread 1 — Readable usernames for OIDC auto-created users

When OIDC auto-creates a new user and no valid email claim is present (e.g. Fall C with a custom `email_claim`), the username was derived from the opaque `provider_sub[:30]` — producing unreadable usernames like `2f8a91bc04d7e3c0a512`. This PR fixes the derivation chain to prefer human-readable IdP claims.

**New derivation order (no email available):**
1. `preferred_username` claim → sanitized
2. `name` claim → sanitized (fallback if `preferred_username` is absent or sanitizes to empty)
3. `provider_sub[:30]` → sanitized (last resort)

`isinstance` guards protect against non-string claim values (e.g. a list), and sanitization is applied per candidate so that a value that strips to empty (e.g. `"!!!"`) correctly falls through to the next option instead of silently producing `"oidcuser"`.

### Thread 2 — Configurable default group for OIDC auto-created users

Each OIDC provider gains a `default_group_id` field. When set, auto-created users are placed in that group instead of always landing in Viewers. The resolution chain is: configured group → Viewers fallback → no group.

**Details:**
- Nullable FK column on `oidc_providers` with `ON DELETE SET NULL`; migration ordered after `_migrate_update_auto_link_constraint` to survive the SQLite table-recreation that function performs on stale-formula databases
- SQLite does not enforce FK constraints at the DB level here — a runtime lookup returning `None` falls through to Viewers, keeping the flow safe if the configured group is later deleted
- Provider create/update validate the FK exists (422 on a non-existent group ID)
- Exposed in the OIDC settings form as a group dropdown; detail view resolves the group name or shows the Viewers fallback label
- i18n: 3 new keys (`defaultGroup`, `defaultGroupDesc`, `defaultGroupViewersFallback`) in all 8 locales

**Limitation:** to clear a configured default group, delete the group or select a different one — explicit reset-to-null is not currently supported (consistent with how `icon_url` and other nullable fields behave).

### Thread 3 — Email-gating in UsersPage edit modal

Neither the maintainer nor the contributor can reproduce this on current `dev`. The UsersPage edit modal disables password-reset/email controls but not groups or permissions, and email is optional in the schema. **Dropped from this PR's scope** — reduced to "needs reporter repro" in #1173 for a separate follow-up if concrete steps surface.

## Related Issue

Fixes #1173 (Threads 1 & 2)

## Documentation

- [x] No docs update required for Thread 1 — internal behaviour change only
- [x] User-visible change for Thread 2 — new `default_group_id` field on OIDC provider (form dropdown + CHANGELOG entry added)

## Type of Change

- [x] Bug fix (Thread 1 — non-breaking, fixes opaque username derivation)
- [x] New feature (Thread 2 — configurable default group, backward-compatible)
- [x] Test addition or update

## Changes Made

**Thread 1**
- `backend/app/api/routes/mfa.py`: extend username derivation in the auto-create block (`preferred_username` → `name` → `provider_sub`) with per-candidate sanitization and `isinstance` guards
- `backend/tests/integration/test_mfa_api.py`: add `TestOIDCAutoCreateUsername` with 7 integration tests

**Thread 2**
- `backend/app/models/oidc_provider.py`: nullable `default_group_id` FK column
- `backend/app/schemas/auth.py`: `default_group_id` on `OIDCProviderCreate`, `OIDCProviderUpdate`, `OIDCProviderResponse`
- `backend/app/core/database.py`: idempotent `ALTER TABLE` migration
- `backend/app/api/routes/mfa.py`: FK validation on create/update; resolution chain in auto-create block
- `frontend/src/api/client.ts`: `default_group_id` on `OIDCProvider` and `OIDCProviderCreate` types
- `frontend/src/components/OIDCProviderSettings.tsx`: group dropdown + detail view
- `frontend/src/i18n/locales/*.ts`: 3 new keys in all 8 locales
- `backend/tests/integration/test_mfa_api.py`: 5 new `TestOIDCProviders` cases + `TestOIDCAutoCreateDefaultGroup` with 4 cases
- `frontend/src/__tests__/components/OIDCProviderSettings.test.tsx`: 6 new tests
- `CHANGELOG.md`: entry for both improvements including limitation note

## Screenshots

N/A — backend-only for Thread 1; Thread 2 adds a group dropdown to the existing OIDC provider form (admin-only settings page)

## Testing

**Thread 1 — 7 integration tests in `TestOIDCAutoCreateUsername`:**

| Test | Claims | Expected username |
|---|---|---|
| `test_preferred_username_used_when_no_email` | `preferred_username='johndoe'` | `johndoe` |
| `test_preferred_username_spaces_sanitized` | `preferred_username='John Doe'` | `JohnDoe` |
| `test_name_claim_used_when_no_preferred_username` | `name='Jane Smith'` | `JaneSmith` |
| `test_provider_sub_fallback_when_no_claims` | sub only | `abc123xyz` |
| `test_non_string_preferred_username_falls_through_to_name` | `preferred_username=["listval"]`, `name='BobJones'` | `BobJones` |
| `test_preferred_username_sanitizes_to_empty_falls_through_to_name` | `preferred_username='!!!'`, `name='bob'` | `bob` |
| `test_username_collision_appends_counter` | pre-existing `collider` user | `collider1` |

**Thread 2 — 9 backend + 6 frontend tests:**

| Test | Scenario |
|---|---|
| `test_create_provider_with_default_group_id` | valid group ID accepted, returned in response |
| `test_create_provider_invalid_default_group_id_returns_422` | non-existent group ID → 422 |
| `test_create_provider_omit_default_group_id_stores_null` | omitting field → `null` |
| `test_update_provider_default_group_id` | PATCH updates the field |
| `test_default_group_id_in_public_and_admin_list` | field present in both list endpoints |
| `test_configured_group_assigned_to_auto_created_user` | auto-created user lands in configured group |
| `test_null_default_group_id_falls_back_to_viewers` | `null` → Viewers fallback |
| `test_dangling_default_group_id_falls_back_to_viewers` | deleted group → Viewers fallback (SQLite FK not enforced) |
| `test_administrators_group_can_be_set_as_default` | Administrators group accepted |

```bash
python3 -m pytest tests/integration/test_mfa_api.py::TestOIDCAutoCreateUsername \
                  tests/integration/test_mfa_api.py::TestOIDCProviders \
                  tests/integration/test_mfa_api.py::TestOIDCAutoCreateDefaultGroup -v
# 16 passed
```

## Checklist

## Additional Notes

Thread 1 fix is limited to the auto-create path (`provider.auto_create_users = True`). The auto-link path (matching existing users by `provider_sub` or email) is unaffected.

Thread 2: `default_group_id` follows the existing `exclude_none=True` PATCH semantics — a PATCH that omits the field leaves the current value unchanged. This is consistent with `icon_url` and the other nullable provider fields.


maziggy (changes requested):
Solid work on the username derivation, and the test coverage is exactly the right shape — isinstance guard, sanitizes-to-empty fallthrough, and collision counter all locked
  down. Verified locally: ruff clean, full backend suite green (3574 passed) on the merged dev base.

  I don't merge PRs that don't fully resolve the linked issue, though, and #1173 has three threads:

  1. Opaque username derivation — fixed cleanly by this PR.
  2. Auto-assignment to "Viewers" group — not addressed. The group is hardcoded at mfa.py:1592. LDAP already has a per-config ldap_default_group setting; OIDC needs the
  equivalent. Concretely: a nullable default_group_id FK on OIDCProvider, a migration, exposure on the provider create/update schemas, the OIDC provider settings UI, and a
  Group.name == "Viewers" fallback when null. Probably 80–120 lines including UI + i18n + tests. Same code-path as your existing change so it should fold in cleanly.
  3. "Permissions only editable after adding email" — I tried to trace this and couldn't find the gate. The user email field is optional at schema level; the only email-disabled control I see in the user form is the password reset button (which is correct, since reset sends mail). May be a UX-level confusion or a bug in a different code path. Could you ask the reporter to share the exact UI step / screenshot in the #1173 thread? That way we can either fold a real fix into this PR or split it off as its own issue if it's outside the OIDC code path.

  If folding (2) into this PR would balloon scope past where you're comfortable, I'm fine splitting #1173 into two issues (one for default-group, one for the email-gating mystery) and merging this PR against the username-only issue once the split lands. Either way works — just don't want to merge against a single issue that this only resolves a third of.

netscout2001:
Thanks for the detailed review — really appreciate the structured feedback.

**Update: Thread 2 is now folded into this PR.**

I rebased on top of the latest `upstream/dev` (merge, history preserved) and added the configurable default group:

- **Backend**
  - `OIDCProvider.default_group_id` — nullable FK to `groups.id` with `ON DELETE SET NULL`.
  - Idempotent `ALTER TABLE` migration in `database.py`, placed **after** `_migrate_update_auto_link_constraint` so it survives the SQLite table-recreation that function does on stale-formula DBs.
  - Provider create/update endpoints validate the FK exists (returns `422` on dangling references).
  - Auto-create resolution chain at `mfa.py`: configured group → `Viewers` fallback → no group. SQLite doesn't enforce `ON DELETE SET NULL` (no `PRAGMA foreign_keys=ON`), so a dangling `default_group_id` returns `None` from the lookup and falls through to `Viewers` — verified by an explicit test.
- **Frontend**
  - `OIDCProviderSettings` adds a `<select>` populated from `/api/v1/groups/`. First option (`""`) maps to `null` and shows the “Viewers (default)” fallback label; detail view shows the resolved group name.
  
image

  - 3 new i18n keys (`defaultGroup`, `defaultGroupDesc`, `defaultGroupViewersFallback`) added to **all 8 locales**.
- **Tests**
  - 5 new `TestOIDCProviders` cases: happy path, invalid FK → 422, omit → null, update, response field in both list endpoints (incl. the public one).
  - 4 new `TestOIDCAutoCreateDefaultGroup` cases (DB-agnostic, asserting via the OIDC exchange response): configured group, `null` → Viewers, dangling → Viewers, `Administrators`-as-default.
  - 6 new frontend cases (label, fallback, resolved name, dropdown render + population).
  - Local run after merge: ruff clean, backend `2345 passed`, frontend `1623 passed`, `pip-audit`/`bandit`/`npm-audit` clean.

**One known limitation** (consistent with `icon_url` and other nullable fields): the update route uses `body.model_dump(exclude_none=True)`, so resetting `default_group_id` back to `null` via PATCH/PUT isn't possible — operators have to either delete the group or pick a different one. Happy to switch to a sentinel-based update if you'd prefer that semantics.

**On Thread 3 (email-gating mystery):** I can't reproduce it on `dev` either. The `UsersPage` edit modal disables only the password-reset / mail-related controls, never permissions/groups. Email is optional at the schema level. My read matches yours — likely either UX confusion or a bug on an older release. **Recommendation:** split #1173 — keep this PR against the username + default-group portion, open a separate issue for the email-gate report once the reporter shares an exact reproduction step or a screenshot of the disabled control. Happy to either way; let me know which you prefer and I'll update the PR description / link the new issue accordingly.

maziggy (changes requested):
Beautiful turnaround — thanks for folding Thread 2 in cleanly.

  Reviewed the full diff:

  - Username derivation — preferred_username → name → provider_sub with per-candidate sanitization is exactly right; the "!!!" fall-through case is the kind of thing that
  usually slips. isinstance guards on the claim values are appreciated.
  - Default group feature — column nullable + ON DELETE SET NULL, migration ordered after _migrate_update_auto_link_constraint (correct — that function recreates the SQLite table and would otherwise drop the new column), runtime fallthrough handles dangling refs on SQLite (which doesn't enforce PRAGMA foreign_keys=ON here). Validation returns 422 on bad FK in both create and update. Tests exercise happy path, invalid FK, omit-defaults, update, and both list endpoints with real async_client + db_session. Solid.
  - Frontend — dropdown + detail view + i18n in all 8 locales. getGroups() requires GROUPS_READ but the OIDC page is admin-only anyway, so non-issue.

  On the reset-to-null limitation: accept it. It's consistent with how icon_url and the other nullable fields behave today, and a sentinel/clear pattern would sprawl across every nullable field for marginal benefit. Just add one line to the CHANGELOG entry: "To clear a configured default group, delete the group or select a different one — explicit reset-to-null is not currently supported."

  On Thread 3 (email-gating): agreed, split it. I can't reproduce on dev either — the UsersPage edit modal only disables password-reset / mail controls, never
  groups/permissions, and email is optional in the schema. Please drop Thread 3 from this PR's scope and either close that part of #1173 or reduce it to "needs reporter repro"
   status. Happy to take a separate issue if it ever resurfaces with concrete steps.

netscout2001:
Splitting Thread 3 out as requested. Neither of us can reproduce the email-gating behavior on current dev — the UsersPage edit modal disables password-reset/email controls but not groups or permissions, and email is optional in the schema. Reducing this to needs reporter repro: if the behavior resurfaces with concrete steps to reproduce (Bambuddy version, browser, user role), happy to take it as a separate issue. Closing this thread.

netscout2001:
> Beautiful turnaround — thanks for folding Thread 2 in cleanly.
> 
> Reviewed the full diff:
> 
> * Username derivation — preferred_username → name → provider_sub with per-candidate sanitization is exactly right; the "!!!" fall-through case is the kind of thing that
>   usually slips. isinstance guards on the claim values are appreciated.
> * Default group feature — column nullable + ON DELETE SET NULL, migration ordered after _migrate_update_auto_link_constraint (correct — that function recreates the SQLite table and would otherwise drop the new column), runtime fallthrough handles dangling refs on SQLite (which doesn't enforce PRAGMA foreign_keys=ON here). Validation returns 422 on bad FK in both create and update. Tests exercise happy path, invalid FK, omit-defaults, update, and both list endpoints with real async_client + db_session. Solid.
> * Frontend — dropdown + detail view + i18n in all 8 locales. getGroups() requires GROUPS_READ but the OIDC page is admin-only anyway, so non-issue.
> 
> On the reset-to-null limitation: accept it. It's consistent with how icon_url and the other nullable fields behave today, and a sentinel/clear pattern would sprawl across every nullable field for marginal benefit. Just add one line to the CHANGELOG entry: "To clear a configured default group, delete the group or select a different one — explicit reset-to-null is not currently supported."
> 
> On Thread 3 (email-gating): agreed, split it. I can't reproduce on dev either — the UsersPage edit modal only disables password-reset / mail controls, never groups/permissions, and email is optional in the schema. Please drop Thread 3 from this PR's scope and either close that part of #1173 or reduce it to "needs reporter repro" status. Happy to take a separate issue if it ever resurfaces with concrete steps.

All done.

PR #1114, opened 2026-04-24
feat(inventory): unified Spoolman inventory UI + AMS slot assignments + Storage Location + NFC write support + Spoolman Filament Catalog Picker

netscout2001:
## Description

This PR integrates Spoolman more deeply into Bambuddy so users never have to leave the app to manage their filament, regardless of which inventory backend is active.

**Unified Inventory UI for Spoolman mode**
When Spoolman is enabled, the Bambuddy inventory page proxies all spool data through a new backend API (`/spoolman/inventory/*`). Create, edit, archive, delete and bulk-create spools — all from the same table/card UI as the internal inventory. The frontend is fully inventory-backend-agnostic.

**AMS Slot Assignments (Spoolman)**
A new `spoolman_slot_assignments` table maps `(printer, ams_id, tray_id) → spoolman_spool_id` locally, without touching Spoolman's `spool.location` field (which remains user-managed). A full CRUD API (`POST/DELETE/GET /spoolman/inventory/slot-assignments`) handles assignment, unassignment, and lookup. The sync routes use the slot table as a no-RFID fallback (`spoolman_spool_id_hint`): if a tray has no readable RFID tag the previously assigned spool ID is used to keep weight data flowing. Trays with neither RFID nor a hint produce a `SkippedSpool` entry in the sync response instead of silently returning null. When a printer is deleted, all its slot assignments are explicitly removed (SQLite does not enforce FK cascades without `PRAGMA foreign_keys`).

**Storage Location field**
A new "Storage Location" text field on every spool (e.g. "Shelf A, Box 3"). In internal inventory mode it is stored locally. In Spoolman mode it maps bidirectionally to Spoolman's native `location` field — reads on load, writes back on save, and can be explicitly cleared.

**"Open in Inventory" deep-link from AMS slot hover card**
The hover card shown when hovering over an AMS slot now contains an "Open in Inventory" button (for both Spoolman-linked and internally-assigned spools). Clicking navigates to `/inventory?spool=<id>`, which immediately opens the edit modal for that exact spool. Falls back to a targeted single-spool fetch if the spool list is not yet cached.

**SpoolBuddy NFC write support for Spoolman spools** *(bonus — not part of the original FR)*
[...]

PR #1200, opened 2026-05-03
[Fix] Dual-UID spool identification, use tray_uuid end-to-end

### LINKED ISSUE #984: [Bug]: Respect both BL spool tags
### Bug Description

BL spool have 2 tags each, with unique tag UIDs. The actual identifier for the spool, tray UID, has the same value stored in both tags (usually it's one of the two tags UIDs with 8 chars of padding). Currently, Spoolbuddy doesnt even read the tray UID block 9, at all.

So for example, let's say that the AMS reports the tray UID A6EA1BC600000100 for a spool, and it gets saved in inventory as it's the first time you use it. What Spoolbuddy actually reads for that spool is the material info (block 4+5), gets a bunch of gibberish, and falls back to tag_uid for matching. So if you flip it one side and read on tag UID, that is A6EA1BC6 (just 8 chars), you're lucky, and the spool gets correctly matched. But if you read the other side, the tag UID is different (), and the spool is not resolved in inventory, and therefore marked as new. You could also get lucky and have a fuzzy match of blocks 4+5 against the tray UUID.
My solution so far was to read block 9 and use that to match in inventory.

### Expected Behavior

.

### Steps to Reproduce

.

### Printer Model

Not printer-related

### Bambuddy Version

latest

### Printer Firmware Version

latest

### Installation Method

Manual (git clone)

### Operating System

Linux (Ubuntu/Debian)

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

### Checklist

Keybored02:
## Description
Adds a secondary `tag_uid_2` field so a spool can be identified by two different hardware UIDs (different physical tags). Fixes tray_uuid being silently dropped when adding a new spool through SpoolBuddy — it's now saved and used as the primary match identifier end-to-end. Also fixes the spool card staying visible on the SpoolBuddy dashboard after removing a freshly-added spool.

## Related Issue

Fixes #984

## Documentation

**Pick one**:
- [] Docs PR(s) linked above
- [x] No docs update required — reason: backend update and behavior fix

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

1. Dual tag UID support (tag_uid_2)

Adds an optional secondary hardware UID column per spool, to handle the real-world case where the same physical tag is reported with a different first byte by different NFC reader hardware. The second slot is strictly opt-in — it is never auto-assigned; it can only be set explicitly via the PATCH /spools/{id}/link-tag API or the spool form.

New tag_uid_2 VARCHAR(16) column on the spool table with an inline _safe_execute migration
All matching logic (get_spool_by_tag, fuzzy fallback) searches both columns transparently via a shared _fuzzy_uid_match() helper
Conflict checks in the link-tag route cover both columns on active and archived spools
Frontend displays the primary UID with both in the tooltip; "tagged/untagged" filters and the write-tag page check both columns

2. tray_uuid saved end-to-end through SpoolBuddy

Previously, when a Bambu Lab spool was scanned by SpoolBuddy and the user created a new inventory entry or linked it to an existing spool, the tray_uuid (read from MIFARE Classic block 9) was silently dropped. This meant subsequent scans matched by hardware UID only, losing the more-reliable UUID-based matching.

spoolbuddy_unknown_tag and spoolbuddy_tag_matched WebSocket broadcasts now include tray_uuid
useSpoolBuddyState carries tray_uuid through state (unknownTrayUuid, currentTrayUuid)
SpoolBuddyDashboard tracks displayedTrayUuid alongside displayedTagId, and passes it to createSpool and linkTagToSpool calls
tag_type is set to 'bambulab' (not 'generic') when a tray_uuid is present

3. Dashboard card cleared on tag removal

After a quick-add, removing the spool from the scale/reader left the card visible. The tag-removal useEffect only cleared state when hiddenTagId was set (i.e. the user had closed the card), so a freshly-added spool — where the card was never manually closed — was never cleared on physical removal.

The else-branch now clears displayedTagId/displayedTrayUuid whenever either is non-null, unconditionally

## Testing

## Checklist


maziggy:
Not sure if this is the right moment to add it. Please align with @netscout2001

Keybored02:
Not sure it's possible to defer it. Spoolbuddy is getting traction and the current tag assignment has a duplication issue and does not track the correct NFC block. Later changes will have a much broader impact.

maziggy:
Talking to each other helps ;)

Keybored02:
Will do.

netscout2001:
Hey, thanks a lot for taking the time to put this together — I really appreciate the effort! 🙏

That said, I don't want to pull this into my current PR at this point. It can be revisited and implemented once I'm done with the Spoolman integration on my side — I'd rather not introduce a second UUID in there right now.

I only had a quick look, so this is a fly-by review, but from what I can see the test coverage isn't complete yet — a few tests are in there, but not enough to fully cover the changes. On top of that, this is more of a new feature than only a bug fix.

Let's revisit this later once Spoolman is wrapped up. Thanks again for contributing! 🚀

Keybored02:
Thanks for taking the time to review this. 

This is not a DB feature but rather a fix for an issue in Spoolbuddy. A Bambulab spool has two tags, each with one unique `tag_uid`, but both sharing the same `tray_uuid` values (the correct identifier). The NFC currently reads and saves only `tag_uid `, which means that a spool will be duplicated into two entities based on the side on which is scanned. This effectively nullifies the usefulness of SB, and creates a DB state that is complex to reconcile in the future.

If you think that covering the full behavior is excessive (2 tags and 1 tray IDs per spool), at least add the switch to tray_uuid and the hw config for that. 

Given that the project is getting quite a bit more users recently, I do not see how this can be postponed.  I can extendTesting coverage.

maziggy:
Unfortunately we have to postpone it until PR #1114 is finished. The PR is too complex and I would like to get it finished.

I also don't see any pressure with the duplicate tags. Don't experienced any issues with it unti now and also don't received any issues which are related.

maziggy:
BTW: we are talking about BL spools - third party spools don't have the issue. So we have a nice workaround until it's fixed: don't weigh your BL spools, just put them into the AMS.

Keybored02:
> Don't experienced any issues with it unti now and also don't received any issues which are related.

I constantly run into this issue. The lack of reporting would probably be for the fact that it's rather hard to catch.

> don't weigh your BL spools, just put them into the AMS.

That's not how people are using it though. 

Anyway, will wait.

maziggy:
> That's not how people are using it though.

Sure, it's a workaround until the fix lands.

maziggy:
I've just merged the large unified inventory PR. Now it's the perfect time to merge your changes.

maziggy:
Lemme check....

Keybored02:
I'll fix the merging issues, add test coverage, and test the while thing again. Will take a while.

maziggy:
Wait....I'm nearly finished with resolving conflicts.

maziggy (changes requested):
Three things worth changing before merge plus one nit:

1. tag_uid_2 length should be VARCHAR(32), not VARCHAR(16).

  PR #1241 widened tag_uid from String(16) to String(32) to support the full Bambu tray UUID width. Adding tag_uid_2 as String(16) reintroduces the old limit on a brand-new column, so the secondary slot can't hold the same kinds of UIDs the primary slot now accepts. I bumped this to String(32) (and the migration ALTER TABLE to VARCHAR(32)) during the merge — please confirm that's what you wanted.

2. InventorySpool.tag_uid_2 should be optional, not required.

  Making it tag_uid_2: string | null (required) breaks every existing literal that constructs an InventorySpool shape — tsc -b flagged three on dev that the PR didn't touch:
  - frontend/src/components/ForecastPanel.tsx:1189 (bulkCreateSpools payload from "received" shopping-list items)
  - frontend/src/pages/spoolbuddy/SpoolBuddyDashboard.tsx — the setJustLinkedSpool literal in the Spoolman link branch
  - frontend/src/pages/spoolbuddy/SpoolBuddyDashboard.tsx — the createSpoolmanInventorySpool payload in the Spoolman quick-add branch

  I added tag_uid_2: null in those three spots so the build passes, but the cleaner long-term fix is tag_uid_2?: string | null on the InventorySpool interface — matches how every other additive column in this codebase has been introduced and avoids the next dev who adds an inventory literal silently breaking the build.

3. Spoolman backend has no tag_uid_2 column.

  The dual-UID feature is local-DB only — Spoolman's REST API doesn't expose a second tag field, and _map_spoolman_spool won't have anywhere to read one from. So under
  spoolman_enabled=true, the secondary UID column is unreachable: setting it via the form persists only locally (and only in local mode), and no Spoolman spool will ever match against it. That's fine as a scoping decision, but worth saying so explicitly in the PR description so users on Spoolman don't expect dual-reader matching to work.

4. Nit: setShowLinkModal(false) on the success path of handleLinkTagToSpool is redundant with the existing finally { setShowLinkModal(false) }. Removed during the merge.

netscout2001:
For 3. it is possible to add a tag2 field as extra field to spoolman.
So it is 100% the same as for local inventory.

Keybored02:
1. Confirm, it's fine

> the cleaner long-term fix is tag_uid_2?: string | null on the InventorySpool interface

2. Agree
3. As @netscout2001 mentioned, it can be added as an extra field.
4. Fine.

maziggy:
Question is, if this PR is still good after merging the huge "Unified inventory" code.

Keybored02:
I've not run into any particular issues with my flow, but I've not tested extensively on Spoolman. I'm also unclear if you still need me to add anything before a review? Seems like you've already fixed the remaining points.

maziggy:

tooldforgames:
@Keybored02 .   i think you are making this more complicated than it needs to be (speaking for bambu labs spools only)   the Only key you need is tray_uuid  key 9  if this is saved in tag in spoolman then the scale and the AMS will both see the same record,  tag_uid nor tag_uid2 is really needed for bambu spools,  if you must have one to display on the scale then use the first 8 char or last 8 char of key 9  this is the same both sides of the spool ID.
I have added some of the changes keybored02 made to add key 9 and i dont have the knowledge to save it so spoolman tag, i believe this uses tag_uid and it cant hold the full key 9.  this means the record wont save with the full key, but when the scale scans key 9 and you already have a ams record the scale will see it.

@maziggy   the scale is basically useless as it stands for bambu spools.
put spool in ams and it creates a record
put spool on scale,  it prompts to create a new record (blank)  you cant select the ams record as it has a rag id in spoolman so is not shown to select so the record is created and cant be seen by the ams.
put the spool on the scale again and the scale has a 50-50 of being a new record (second side tag) or pull the other blank record.

The key to all this is key 9,  sorry for the pun there, as i say i have edited some of the code and have the scale seeing the ams record and updating it,  but cant create records :(

regards
tooldforgames

Sorry if its not allowed to post these sort of comments here

maziggy:
> @maziggy the scale is basically useless as it stands for bambu spools.
put spool in ams and it creates a record
put spool on scale, it prompts to create a new record (blank) you cant select the ams record as it has a rag id in spoolman so is not shown to select so the record is created and cant be seen by the ams.
put the spool on the scale again and the scale has a 50-50 of being a new record (second side tag) or pull the other blank record.

Use the latest version and disable auto-create in settings.

tooldforgames:
Hi Martin

Thank you for you reply.

Im hoping keybored02 understands what i am trying to say with the key 9 issue.

Im fine with the AMS creating a record when a new spool is inserted.
The main issue is the Scale relies on the tag ID to run, the Mifare tag ID is created when the tag is created and every single tag has a different ID and has nothing to do with anything Bambu.
Bambu use block 9 to store their spool ID and its the same both sides of the spool.

I did some messing around with the files and got the scale to read block 9 and when it found the AMS created record it would show all the spools details (the AMS populates the spoolman tag field with block 9 key)
It would also allow you to store the weight off the scale directly to the AMS record.  Im not a programmer and used A1 to help get that working and when trying to get the scale to store the block 9 key as it created the record in spoolman would not work becase the tag_uid the scale stores the value in to set the record up cannot hold the block 9 key as its 32 something and tag_uid is 8 ?

If the block 9 key is stored in the spoolman tag  field then both systems would see the same record and it would mean the tag_uid or proposed tag_uid2 would not be needed in bambu's case.

Thanks
Gavin


-------- Original Message --------

Keybored02:
@tooldforgames We need to keep tag_UID support both for non-BL spools and for backwards compatibility with inventory records using them. I'm working a migration path for the older BL spools.

apoapoapo:
> > @maziggy the scale is basically useless as it stands for bambu spools.
> > put spool in ams and it creates a record
> > put spool on scale, it prompts to create a new record (blank) you cant select the ams record as it has a rag id in spoolman so is not shown to select so the record is created and cant be seen by the ams.
> > put the spool on the scale again and the scale has a 50-50 of being a new record (second side tag) or pull the other blank record.
> 
> Use the latest version and disable auto-create in settings.

i tested that in that issue here but it was closed:
https://github.com/maziggy/bambuddy/issues/1810

Sawtaytoes:
@Keybored02 @maziggy — this had gone stale and would no longer merge, so I rebased it onto current `dev` and applied the review: #3022. The original work is yours, @Keybored02; I kept your commits' authorship and I'm happy to close mine if you'd rather update this branch.

Only the import in `backend/app/api/routes/inventory.py` conflicted. Review item 2 (`tag_uid_2` optional on `InventorySpool`) is applied; items 1 and 4 were already done here; item 3 is now stated explicitly in the description. Gates on that tree: ruff clean, pytest 7254 passed, `tsc -b` clean, vitest 2432 passed, i18n parity clean.

I also confirmed block 9 independently on a USB PN532 — two tag UIDs on one spool, one block-9 value, byte-for-byte the `tray_uuid` already stored from the AMS. Details in #3022.

Thomansky:
Coming from the reader side — I implemented the block 9 switch in ESPoolBuddy (CSchlipp/espoolbuddy#13) and measured it against real hardware. Posting the data here because it settles the question @tooldforgames raised on Jun 30, which I think is the right one.

**Block 9 is byte-for-byte the `tray_uuid` the printer reports over MQTT.** Two spools hand-scanned with a PN532, then loaded into an AMS, then read back from Bambuddy's own printer status:

```
ams0/tray1   printer     tag_uid 2C0273DD00000100   tray_uuid 9E0B0717BEE94D7887EB1D8DFD1A14F3
             hand scan   tag_uid 2C0273DD           tray_uuid 9E0B0717BEE94D7887EB1D8DFD1A14F3

ams0/tray2   printer     tag_uid 1E783DA000000100   tray_uuid 5E5498918CBF4B94A25EF669C24DECC3
             hand scan   tag_uid 8E3A00A2           tray_uuid 5E5498918CBF4B94A25EF669C24DECC3
```

Tray 2 is the interesting row: the hand scanner read card UID `8E3A00A2`, the printer read `1E783DA0` — the spool's two different tags. They cannot match on `tag_uid` by any rule. Block 9 identified the spool across both readers exactly, and both matched their existing inventory row without creating a duplicate. Full write-up in #984.

So for Bambu Lab spools, block 9 alone is sufficient, and `get_spool_by_tag()` already prefers `tray_uuid` over `tag_uid` — the backend can do this today, it just never receives a `tray_uuid`, because `daemon/pn5180.py` reads `BAMBU_BLOCKS = [1, 2, 4, 5]` and never touches block 9.

**On `tag_uid_2`:** I don't think the measurement makes it pointless, it just narrows its purpose. It buys nothing for BL spools. It does still matter for third-party spools that ship two read-only tags with no shared identifier — see #1740. Framing it that way might make the review smaller: block 9 as the BL path, `tag_uid_2` as the foreign-spool path, rather than one mechanism doing both.

**On the migration path** (@Keybored02, your Jun 30 note): the value pre-fix installs hold is worse than random, but that also makes it detectable. With block 9 unread, the fallback ends up at the raw hex of block 4 — the filament type. Three separate spools of mine carried:

```
tray_uuid = 504C41204D6174746500000000000000     <- ASCII: "PLA Matte"
```

Identical for every spool of that type, which is why they collapse onto one row. It fails both RFC 4122 checks (version nibble `7`, variant `6`), whereas every genuine value I have — 13 from the printer, 6 from block 9 — passes both. So `WHERE tray_uuid IS NOT NULL AND NOT <valid v4>` selects exactly the rows to clear. They re-link on the next scan; nothing beyond the stale identity is lost.

Happy to run further checks — I have a PN532 hand scanner, an AMS and a few hundred spools to test against. No PN5180, so I can't verify the daemon side.

Keybored02:
My issue with just going forward with this is that A) people will need to rescan their spools; B) We need a full UI-driven workflow for this; C) There is still no agreement with @maziggy on non-BL spool tagging. 

The point is that matching a tray_uuid to an existing, in-db spool that has none has to be done with a guided two-scan sequence in Spoolbuddy. There will be a reconciliation feature that will be mandatory for those spools that passed through Spoolbuddy but not yet through an AMS. This is not a small deal for all the users that have stockpiles of refills and spools. 

As it stands, this PR is kept open but it will not be merged standalone. What needs to be done:

1. Backfill needs a guided flow: Matching a tray_uuid to an existing NULL row needs an explicit two-scan sequence in SpoolBuddy and reconciliation for anything the AMS hasn't covered. This is bot backend,as well as an interactive guide. 
2. No automatic write off a fuzzy match: A %LIKE% suffix hit must never trigger a tray_uuid backfill
3. Third-party spools  dual-UID: I agree. What's missing is the Ui flow to guide user through it.

I also cannot reproduce the block 4 issue on a PN5180. Whatever the fix ends up being, it needs to go into the write path rather than the backend.

Thomansky:
@Keybored02 — agreed on all three points. And the reproduction for the fallback doesn't need a PN5180, which I think explains why you don't see it.

**Where the bad value comes from.** It isn't reader-specific, it's in the daemon. `spoolbuddy/daemon/nfc_reader.py` (around lines 256-264):

```python
# Fallback for partially decoded payloads: use first 16 raw bytes as hex.
# This preserves compatibility with older decoding behavior.
uuid_str = raw[:16].hex().upper()
```

With block 9 unread, `raw` begins at the detailed-type field, so the first 16 bytes are the filament type string. Feeding synthetic blocks through that same line:

```
PLA Matte            -> 504C41204D6174746500000000000000
PLA Basic            -> 504C4120426173696300000000000000
PETG HF              -> 50455447204846000000000000000000
Support for PA/PET   -> 537570706F727420666F722050412F50
```

The first is bit-identical to what my database actually holds. A reader that decodes the payload fully never reaches that branch — which is consistent with your PN5180 result. It bites on any path where the decode is partial, and the collision is by construction: every spool of a given type gets the same "UUID".

**Correction to my previous comment.** I wrote that three spools of mine "carried" that value. Those bytes came out of my own inventory DB, written through my ESPoolBuddy scanner — not from a PN5180. I should have made the argument above instead: the identical fallback line lives in the daemon, so the same collision is reachable there regardless of my hardware.

**On your three requirements** — I agree with all of them, and I don't think anything I proposed conflicts with them:

1. **Guided two-scan backfill.** Yes. Writing a `tray_uuid` onto an existing NULL row is an identity assertion and needs the user in the loop, plus reconciliation for stock that never passed an AMS.
2. **No automatic write off a fuzzy match.** Yes, unconditionally. A `%LIKE%` suffix hit is a display convenience, never identity proof.
3. **Fix belongs in the write path.** Yes.

One distinction worth keeping separate: clearing a provably-invalid value is not the same operation as backfilling a correct one. `WHERE tray_uuid IS NOT NULL AND NOT <valid RFC 4122 v4>` deletes rows that cannot be any spool's identity — it does not decide *which* spool they were, so it doesn't pre-empt your guided flow. If anything it makes that flow unambiguous, because the colliding values are what currently make several spools indistinguishable from each other. Every genuine value I have (13 from printer MQTT, 6 from block 9) passes the version and variant checks; every fallback value fails both.

Also worth linking: @Sawtaytoes rebased this onto current `dev` as #3022 with the review items applied and green gates, and confirmed block 9 independently on a USB PN532 — two card UIDs on one spool, a single block-9 value, byte-identical to the `tray_uuid` already stored from the AMS. That's a second reader, independent of mine.

Keybored02:
Please, stop copy pasting the LLM response. Contributions to the discussion should be human-made. 

This is mostly an issue of persistence, as in Spoolbuddy the corrupt value is computed, sent, used for a lookup that misses and discarded. It should go into the PR.  I've to check if there are any read path bugs that would cause the match to fail on tray_uuid before it reaches the tag_uid stage, which I suspect.  

Testing for a Spoolbuddy update should be done on a Spoolbuddy, even though the code is shared across other projects. This is especially valid as PN532 and PN5180 access and decrypt sectors differently, and the latter is less susceptible to partial dumps. The current issue is HW-independent, but still.

tooldforgames:
hi keyboard02 / maziggy
i am on my way back from holiday currently, if i get time i will update tonight or it will be in the morning. i have the scale reading blk 09 and matching the AMS code in spoolbuddy this recalls the correct record and displays the info on screen,  i am 90% sure the first 8 char from key 9 and the last 8 char are the bambu spool ID's for each side, but i will need to check that at home.   But from the blk 09 data you can also pull out all the information to read get the spool data , colour ,type etc.  i can upload that code if needed (its just python),  this will also solve another PR with having the scale create a full spool Bambu record from just that blk 09 read

tooldforgames

Keybored02:
The issue with material type extraction is that there are some types that are not documented in the community repos, and the fact that they change. This can be done, but needs a fallback path and a quick way to tell the user (e.g. "This value is not associated in our DB. Consider submitting a feature request to have it added"). Regarding block 9, as far as I can tell there is no connection to tag_uid.

Issue #1060, opened 2026-04-20
[Feature]: Structured Printer Card UX Refresh


EdwardChamberlain:
### Problem or Use Case

I appreciate that changes to the printer card UI may come down to design direction and personal preference, and that some of these suggestions may be contentious. With that in mind, I would like to propose a refresh of the Printers page UI.

The goal is to improve clarity and consistency across printer cards by introducing a stronger visual hierarchy, making interactive elements more obvious, and enforcing more consistent layout and behaviour between cards and view modes.

Working from top to bottom of the page, the current issues I have identified are:

* Error indicators are unclear, as it is not immediately obvious what the dots relates to or where they are sourced.
* Printer images sit on very dark backgrounds, which reduces separation and divider lines hard to distinguish which reduces visual clarity.
* Status pills are numerous and can wrap onto multiple lines, which increases visual noise, makes overall status harder to read at a glance, and causes inconsistent card heights
* The Controls section includes some items that behave as indicators rather than controls, such as fan speeds
* In medium card layout, ~~some control elements overlap~~ (fixed in latest release)
* AMS cards have alignment problems, along with several overflow issues
* Plate clear, light, and camera actions are placed in the footer instead of within the main controls area, which makes the interaction model feel less consistent

<img width="1511" height="855" alt="Image" src="https://github.com/user-attachments/assets/3ce6ec23-8ebd-486c-bd32-7ea18fc24be7" />

### Proposed Solution

Redesign the printer cards to improve structure, consistency, and usability across both normal and compact views.

<img width="1512" height="857" alt="Image" src="https://github.com/user-attachments/assets/533501bc-f6df-43ec-8c24-021de341dcf6" />

Proposed changes:

- Add printer icon and title to match the queue page
- Introduce a new top-line health indicator system, with each indicator clickable to filter printers by health state:
  - **Green** = Healthy
  - **Amber** = Requires user attention
  - **Red** = Error state
  - **Blue** = Eco, where the printer is unreachable but linked to a powered-off smart socket

<img width="677" height="796" alt="Image" src="https://github.com/user-attachments/assets/399e16c5-5b19-4adc-bf3f-a1c1b0748f38" />

- Tighten the visual presentation of the printer title area
- Condense the current health pills into a single, clearer health indicator with colour matching the above health state
- Clicking the health indicator opens a dedicated health overview view

[...]

PR #1203, opened 2026-05-03
[FEAT] Printer page header update

EdwardChamberlain:
## Description

Update to the printers page header to add element consistency and layout tightening. 

## Related Issue

Partial implementation as discussed in #1060

## Documentation

**Pick one**:
- [ ] Docs PR(s) linked above
- [ ] No docs update required — reason: No functional change - only Element clean up

## Type of Change

- [ ] Documentation update
- [X] Code refactoring (UI Refactor)
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Add printer icon to header
- Align sort / filter controls with search
- Enforce consistent UI element height and style
- Use "Bambuddy styled" UI elements for drop down and check box instead of relying on variable system elements.
- Reorder header elements into clear groups: Filter, View, Action
- Match "subtitle" formatting to other pages
- Implement collapsing for better consistency across UI widths

## Screenshots

Full width Desktop:
image

Mobile:
image

## Testing

- [X] I have tested with my printer model: P1S / H2S

## Checklist

## Additional Notes

Tested against multiple screen widths.


maziggy (changes requested):
Thanks for picking this up — the layout looks clean and the regrouping into Filter / View / Action makes the toolbar a lot easier to scan.

  One blocker before merge: replacing the native `<select>` with the new   `ToolbarDropdown` breaks 5 existing tests in PrintersPage.test.tsx that use `getByDisplayValue('All statuses')` + `fireEvent.change`. They need to be rewritten to click the new button trigger and then click the 
  option (e.g. `userEvent.click(getByRole('button', { name: /all statuses/i }))`
  followed by `userEvent.click(getByRole('button', { name: /^offline$/i }))`).
  Test names below.

  A few smaller things worth confirming while you're in there:
  - The `<Plus>` icon was dropped from the "Add Printer" button — was that    intentional or an oversight in the consolidation?
  - `text-sm` was removed from `StatusSummaryBar`; assuming that's the
    "match subtitle formatting to other pages" change.
  - New i18n keys (`printers.toolbar.filters/view/actions`) rely on the inline `t('…', 'Filters')` defaultValue — they should also be added to `en.ts` for consistency with the rest of the printers section.
  - In the compact action menu, the sort-direction arrow button sits next to a `w-full` ToolbarDropdown without shrink-0 — worth a quick check on a ~360 px viewport that the arrow doesn't get squished.

  Failing tests (all in PrintersPage.test.tsx):
  - filters by status (offline) via dropdown
  - shows empty state when status filter matches nothing
  - combines search and status filter
  - filters by location via dropdown
  - hides location filter when no printers have a location

EdwardChamberlain:
Test cases updated. Now showing all pass. Added i18n keys. 

> The <Plus> icon was dropped from the "Add Printer" button — was that intentional or an oversight in the consolidation?

This was semi-intentional. I dropped it to reduce toolbar width but happy to restore it as I think the width is under control now.

> text-sm was removed from StatusSummaryBar; assuming that's the
"match subtitle formatting to other pages" change.

Yes, thats right. It's helping enforce closer look to the print queue page. 

> In the compact action menu, the sort-direction arrow button sits next to a w-full ToolbarDropdown without shrink-0 — worth a quick check on a ~360 px viewport that the arrow doesn't get squished.

I think it's ok but I added a guard for it just in case.

EdwardChamberlain:
Having checked the other pages I have restored the "+" to add printers as that is more consistent.

maziggy (changes requested):
Thanks for the rework — tests pass, the + is back, sort-arrow has shrink-0, and the dropdown→ToolbarDropdown test rewrites look clean. One thing left before merge.

  i18n parity: The new printers.toolbar.{filters,view,actions} keys only landed in en.ts and de.ts. Bambuddy ships 8 locales — fr, it, ja, pt-BR, zh-CN, zh-TW are missing them. The
  runtime falls back to the inline defaultValue so nothing visibly breaks, but we keep all 8 files structurally in sync so translators have a single place to land translations later
  (and so the locale-parity check passes).

  Could you add the same block to the other 6 locales? English fallback strings are fine — translators refine in a follow-up. Block to paste, right after the existing filter: { ... }
  object in each file:

  toolbar: {
    filters: 'Filters',
    view: 'View',
    actions: 'Actions',
  },

  Files to update:
  - frontend/src/i18n/locales/fr.ts
  - frontend/src/i18n/locales/it.ts
  - frontend/src/i18n/locales/ja.ts
  - frontend/src/i18n/locales/pt-BR.ts
  - frontend/src/i18n/locales/zh-CN.ts
  - frontend/src/i18n/locales/zh-TW.ts

  Once those are in, this is good to merge.

EdwardChamberlain:
Added the printers.toolbar block to the remaining six locales 😊

maziggy (approved):

maziggy:
> No docs update required — reason: No functional change - only Element clean up

Since the PR introduces new features (multi printer actions) the docs need to be updated. Please add to Wiki.

EdwardChamberlain:
> > No docs update required — reason: No functional change - only Element clean up
> 
> Since the PR introduces new features (multi printer actions) the docs need to be updated. Please add to Wiki.

I might be misunderstanding what you mean but this PR doesn't introduce any new features. The Multi-printer actions were already present before this and are documented here: https://wiki.bambuddy.cool/features/printer-control/#bulk-actions and implementation remains unchanged.

maziggy:
Sorry, my fault.

Issue #1154, opened 2026-04-28
[Feature]: Enhanced filament colour handling


maugsburger:
### Problem or Use Case

It would be great to have a visual representation of filaments with transparency and/or multiple colours and/or effects. 

It can even be misleading, getting just a single colour shown on the filament overview, when it is in fact a gradient or other type of effect.

### Proposed Solution

I would love to set for a filament

1. multiple colours, could be just comma seperated in the current field
2. set the mode of combination (gradient, multi-color) as a seperate property
3. add transparency (probably the most universal would be aarrggbb for the colour value(s))
4. set effects (sparkle, wood) as a seperate property

This affects how the filament colour is shown:

- 2 in combination with 1 decides how the colours are combined, I like the way it's done on https://3dfilamentprofiles.com/
- 3 would be visible when having a checkerboard pattern behind the colour(s).
- 4 could be some sort of overlay with sparkling dots / wood pattern / … and transparent background overlaying the colour.

### Feature Category

Spool Inventory

### Priority

Would improve my workflow

### Mockups or Examples

https://3dfilamentprofiles.com/filament/details/159
<img width="501" height="237" alt="Image" src="https://github.com/user-attachments/assets/d1f660b1-2edf-44af-97ee-129f4e3154a2" />

https://3dfilamentprofiles.com/filament/details/124
<img width="501" height="225" alt="Image" src="https://github.com/user-attachments/assets/929f3282-79a8-4562-aac8-4551c22d5bc7" />

https://3dfilamentprofiles.com/filament/details/20797
<img width="501" height="266" alt="Image" src="https://github.com/user-attachments/assets/fa369678-5a5a-4721-8bfd-45ec2bcb1275" />

Transparency could look like here:
<img width="491" height="380" alt="Image" src="https://github.com/user-attachments/assets/c4996ebe-d0c0-48a5-8c46-903f8452ee9b" />

### Contribution

- [x] I would be willing to help implement this feature

Regarding the development help: Leave the documentation part for me :>

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
> multiple colours, could be just comma seperated in the current field

What page/function you are talking about? Not sure....

maugsburger:
> > multiple colours, could be just comma seperated in the current field
> 
> What page/function you are talking about? Not sure....

[...]

PR #1205, opened 2026-05-03
UI/multi color handling

maugsburger:
## Description

Rework multi color swatch handling

## Related Issue

https://github.com/maziggy/bambuddy/issues/1154

## Documentation

Not done so far. I will update the documentation as soon as I know this PR is considererd mergeable.

**Pick one**:
- [ ] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

Only UI rendering changes:
- Fix that all swatch patterns were tiled.
- Create unique but constant sparkle patterns
- Different density presets per swatch type / header / …
- More prominent checkerboard, works better on light and dark filaments
- show multi-color in <color-count> segments.

## Screenshots

image

## Testing

Looked at the changes in a browser, updated the automated tests.

## Checklist


maziggy (changes requested):
Thanks Martha — direction is good (size-aware sparkle density, more prominent checkerboard, hard-split multicolor) and the call-site fanout is clean. Two correctness items to address before merge, then a few performance / packaging concerns.

  Blockers

  1. filamentSwatchHelpers.ts:114 — off-by-one in the sparkle loop. for (let i = 1; i < preset.dotCount; i++) produces dotCount - 1 sparks. The tests at lines 235-237 of
  FilamentSwatch.test.tsx bake the bug in (preset 5 → 4 radials, 40 → 39, 20 → 19). Either start at i = 0 or rename the field to something like dotCountBound so the variable name matches its semantics. Right now the preset table reads as "card has 40 sparks" but it actually paints 39.
  2. filamentSwatchHelpers.ts:108-109 — comment says "positions seeded from spoolId so each spool has its own unique, but always stable, sparkle pattern", but the actual seed (in FilamentSwatch.tsx:54 and buildFilamentBackground :248) is ${rgba}|${extraColors}|${subtype}|${effectType}. Two PLA-Sparkle white spools render with identical sparkle. Either thread spool.id through into the seed (touches FilamentSwatch props) or update the comment to match what you're actually doing — colour-keyed sparkle is fine, the comment just needs to say so.

  Significant

  3. Default effectSize is undefined → falls back to 'table' → 4 sparks. Previously every swatch got 13 sparks regardless. Any call site I missed in this PR (or any future caller)
  silently regresses to 4 dots. Suggest defaulting to 'card' or making the prop required.
  4. The previous #1154 follow-up's sparkle deliberately used mixed dot sizes (1px / 1.5px / 2px) and varied opacity (0.65 → 1.0) to give a "metal flake" depth feel — that's now
  flattened to a single rgba(255,248,220,0.9) at one radius per dot. The size variation comes back via dotScale, but please re-introduce the opacity variation per dot. Easy as
  rand.floatBetween(0.65, 1.0) for the alpha.
  5. random-seed (with transitive json-stringify-safe) + @types/random-seed for 5-80 dot positions is heavier than warranted. A 5-line mulberry32 inline:
  function seededRand(seed: string) {
    let h = 2166136261 >>> 0;
    for (let i = 0; i < seed.length; i++) {
      h ^= seed.charCodeAt(i);
      h = Math.imul(h, 16777619) >>> 0;
    }
    return () => {
      h = (h + 0x6D2B79F5) >>> 0;
      let t = h;
      t = Math.imul(t ^ (t >>> 15), t | 1);
      t ^= t + Math.imul(t ^ (t >>> 7), t | 61);
      return ((t ^ (t >>> 14)) >>> 0) / 4294967296;
    };
  }
  5. gives the same deterministic result for the same seed without the npm surface area. Happy to leave the dep if you'd rather keep random-seed, but if so move @types/random-seed to devDependencies (it's currently in dependencies — type-only packages shouldn't ship at runtime).
  6. FilamentSwatch.tsx:52-55 builds effectLayer outside useMemo. For a group-header swatch (preset 80) rendered across 100 spools, that's 8000 PRNG rolls per render. Wrap it:
  const effectLayer = useMemo(
    () => typeof effectType === 'string'
      ? resolveEffectOverlay(effectType, `${rgba ?? ''}|...`, effectSize)
      : null,
    [effectType, rgba, extraColors, subtype, effectSize],
  );

  Documentation / tests

  7. The "deterministic" test only verifies same-input → same-output. Once item 2 is resolved, please add expect(spool1Bg).not.toBe(spool2Bg) so a regression that collapses different inputs to identical sparkle is caught.

  Nits

  - FilamentSwatch.tsx:45 — effectSize: effectSize rename is a no-op, just effectSize.
  - helpers.ts:111 — spoolSeed+effectSize without separator can collide in theory ("ab"+"cd" vs "abc"+"d"); use ${spoolSeed}|${effectSize}.
  - helpers.ts:97 — stray space in 'repeating-conic-gradient( #979797... after the opening paren.

maugsburger:
Thx for the feedback so far, will work through it, update the spool inventory documentation in the wiki and then update the PR.

Is frontend/src/utils/random.ts the correct location for the prng+seed implementation?

maziggy:
Yes, frontend/src/utils/random.ts is a good spot — it matches the convention of small single-purpose helpers (colors.ts, weight.ts, currency.ts, etc.) and keeps the PRNG reusable if any other UI later wants stable-per-key randomness (e.g. avatar tints, chart jitter).

  Just two minor things while you're at it:

  1. Export both the seed-string hash and the float generator so callers can pick (createSeededRand(seed: string): () => number plus a floatBetween(rand, lo, hi) convenience). Keeps filamentSwatchHelpers.ts from re-implementing the hi/lo math.
  2. Add a one-liner JSDoc noting it is not cryptographically secure — purely for deterministic visual variation. Future-proofs against someone reaching for it for token generation.

  If you'd rather keep filamentSwatchHelpers.ts self-contained (since it is the only caller today) inlining is also fine — but a utils/random.ts is the better long-term home and I would not push back on that choice.

maugsburger:
1. Off by one fixed in code and test
2. spoolId was a leftover from the beginning, now its hash based on these atttributes `hash_fnv1a32(rgba, extraColors, subtype, effectType` and commented accordingly
3. made the prop required, adjusted the automated tests as well
4. opacity variation is back
5. added own implementation into `frontend/src/utils/random.ts` together with helpers for ranges and float/int
6. see 5
7. added Memo and use buildFilamentBackground (cbbc1ad817ca76c42a4963b5b223413d2007a3f1)

maziggy:
Please never ever modify and push internal files!!!

Deleted test_frontend.sh and test_backend.sh

PR #1231, opened 2026-05-07
feat(encryption): MFA at-rest encryption auto-bootstrap with status UI (#1219)

### LINKED ISSUE #1219: [Bug]: On a default Docker install, MFA_ENCRYPTION_KEY is not set
### Component

Bambuddy

### Bug Description

  On a default Docker install, MFA_ENCRYPTION_KEY is not set, so:

  - OIDC client_secret values are stored as plaintext in the database
  - TOTP secrets (2FA enrollments) are stored as plaintext in the database

  The encryption module (backend/app/core/encryption.py) emits a single WARNING log line at startup pointing this out, but the warning is not surfaced anywhere a user is likely to see it. Setup wizards, the Settings UI, and the docker-compose template do not mention MFA_ENCRYPTION_KEY at all, so an admin who wires up OIDC and asks users to enroll TOTP can reasonably believe those secrets are protected at rest when in fact they are not.

This is a meaningful security degradation for any install that uses either feature. Users actively using OIDC SSO are the most likely to be affected because the OIDC client_secret is directly disclosable from the SQLite database without any decryption step.

### Expected Behavior

  In a current Docker deployment (v0.2.4b3, support bundle attached on a separate ticket), the warning fires on every startup:

  WARNING [backend.app.core.encryption] MFA_ENCRYPTION_KEY is not set — TOTP secrets and OIDC client_secrets are stored in plaintext.

The same install has active POST /api/v1/auth/oidc/exchange traffic in the access log, confirming OIDC is in use against an unencrypted client_secret column.

### Steps to Reproduce

.

### Printer Model

None

### Bambuddy Version

0.2.3.2 and up

### SpoolBuddy Version

_No response_

### Printer Firmware Version

_No response_

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

_No response_

### Screenshots

_No response_

### Additional Context

  Migration considerations

  Legacy plaintext rows must keep working when a key is later configured — the current mfa_decrypt path already handles this (no fernet: prefix → returned as-is). So auto-bootstrap is non-breaking for existing installs: the key gets generated, new writes are encrypted, old reads still work, and the UI can prompt for re-enrollment to migrate.

  Key rotation is a separate (harder) problem and is out of scope here — the current code raises on rotation, which is fine until someone asks for it.

  Severity / impact

  - Affects: any install using OIDC SSO or TOTP 2FA, which is the entire population of users actually relying on the auth features
  - Likelihood: 100% on default Docker installs unless the operator has read the warning in the logs
  - Impact: at-rest secret exposure; database backup or restore-from-bad-source leaks usable credentials directly

### Checklist

netscout2001:
## Description

Default Docker installs stored TOTP secrets and OIDC `client_secret` values in plaintext because `MFA_ENCRYPTION_KEY` was never set and there was no fallback. This PR adds a self-bootstrapping Fernet encryption key in `DATA_DIR/.mfa_encryption_key`, re-encrypts legacy plaintext rows on startup, and surfaces the encryption state in a new Settings → Authentication → Security sub-tab.

Additionally fixes several security bugs discovered during review: a ZipSlip bypass in backup restore, TOTP recovery being blocked when the encryption key was lost, and an OSError in the key-file read path silently falling through to key regeneration (destroying all encrypted rows).

## Related Issue

Fixes #1219

**Pick one**:
- [x] Docs PR(s): https://github.com/maziggy/bambuddy-wiki/pull/25

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition or update

## Changes Made

**Backend — encryption core**
- Add `backend/app/core/paths.py` with `resolve_data_dir()` helper that reads `DATA_DIR` fresh on every call (enables per-test override via `monkeypatch`)
- Rewrite `_load_or_generate_key()` in `encryption.py`: ENV validation (rejects invalid Fernet keys), file read with refusal to regenerate on OSError (previously fell through and would overwrite the key, destroying all encrypted rows), auto-generate with `chmod 0600`, `get_key_source()` getter
- Add `_migrate_encrypt_legacy_secrets()` in `database.py`: re-encrypts existing plaintext rows after `run_migrations()` in `init_db()`; idempotent via `fernet:` prefix check
- Add `EncryptionStatusResponse` schema and `GET /api/v1/auth/encryption-status` endpoint: reports `key_source`, encrypted/legacy row counts per table, and `decryption_broken` flag
- Add `extra = "ignore"` to pydantic `Settings.Config` so unknown env vars (e.g. `MFA_ENCRYPTION_KEY`) in `.env` do not cause a validation error at startup

**Backend — security bug fixes**
- **ZipSlip** (`settings.py`): replace `str.startswith()` prefix check with `Path.is_relative_to()` — the old check was bypassable via a directory whose name shares a prefix with the temp path
- **TOTP recovery** (`mfa.py`): `disable_totp` and `regenerate_backup_codes` previously raised 500 immediately on `RuntimeError` from TOTP decryption, blocking backup-code recovery; now falls through to the backup-code branch
- **`setup_totp`** (`mfa.py`): wrap `pyotp.TOTP(existing.secret)` calls in `try/except RuntimeError` so a broken decryption returns a clean 500 instead of an unhandled exception
- **Backup key copy** (`settings.py`): wrap `shutil.copy2` for the MFA key in `try/except OSError` with explicit logging
- **Restore error handling** (`settings.py`): add `except HTTPException: raise` before `except Exception` so key-write failures surface as 500 instead of being swallowed into a generic response; add `logger.error` with the target path

**Frontend**
- New `SecurityStatusCard` component: five severity levels (green/yellow/orange/red/grey) with 30 s polling, registered in settings search
- Extend `SettingsSearchSubTab` and `UsersSubTab` with `'security'`; render card under Settings → Authentication → Security (admin only)

**i18n**
- Add `settings.encryption.*` keys (14 keys) to all 8 locale files; full translations for `en`/`de`, English seeds for `fr`/`it`/`ja`/`pt-BR`/`zh-CN`/`zh-TW`

**Deployment**
- `.env.example`: `MFA_ENCRYPTION_KEY` block with generation command and backup note
- `docker-compose.yml`: commented `MFA_ENCRYPTION_KEY` entry
- `.gitignore`: `.mfa_encryption_key` entry (mirrors `.jwt_secret` pattern)

## Screenshots

image
image

| State | Trigger |
|---|---|
| 🟢 Green | Key active, all rows encrypted |
| 🟡 Yellow | Key active, legacy plaintext rows remain |
| 🟠 Orange | Key is auto-generated — back up `DATA_DIR/.mfa_encryption_key` |
| 🔴 Red | Encrypted rows exist but no key is loadable (`decryption_broken`) |
| ⚫ Grey | Encryption not configured, no encrypted rows |

## Testing

**Test coverage added:**
- `mfa_encryption_isolation` autouse fixture: per-test tmp `DATA_DIR`, ENV clear, module-global reset
- `TestEncryption`: +2 tests (read-OSError path; renamed existing write-OSError test for clarity)
- `TestEncryptLegacyMigration`: 6 tests covering idempotency, mixed state, no-op without key
- `TestEncryptionStatusEndpoint`: 8 tests (key_source values, row counts, `decryption_broken`, 403 without permission)
- `TestBackupKeyFiles`: 6 tests (backup includes/skips key, restore writes key with `chmod 0600`, restore without key in ZIP, path-traversal rejection including absolute paths)
- `TestTOTPDecryptionBroken`: +3 tests (setup_totp 500, disable/regenerate backup-code recovery 200, previous 500 tests updated to 400)
- `SecurityStatusCard.test.tsx`: 10 tests covering all 5 severity levels, loading, and error states

## Checklist

## Additional Notes

**Backward compatibility:** Existing plaintext rows are read back via the `fernet:`-prefix check in `mfa_decrypt` — no data loss on upgrade. The migration is idempotent and a no-op if encryption is not configured.

**Key priority:** `MFA_ENCRYPTION_KEY` env var → `DATA_DIR/.mfa_encryption_key` file → auto-generate (writes file with `0o600`). OSError during file read refuses to regenerate (would destroy all encrypted rows) and falls back to plaintext instead.

**Backup self-consistency:** `create_backup_zip` includes `DATA_DIR/.mfa_encryption_key` as a top-level ZIP entry. Restoring from such a ZIP on a keyless host fully restores the key and decryption capability. Legacy backups without the key file restore cleanly via the plaintext fallback path.

**Key rotation:** explicitly out of scope per maintainer guidance.


maziggy:

netscout2001:
@maziggy 
ready for check

maziggy:
> No docs update required — reason: All operator-facing documentation is covered inline: .env.example has a full block with the generation command and backup note; docker-compose.yml has a commented entry; CHANGELOG.md is updated.

Also needs Wiki docs.

maziggy (changes requested):
**Blockers (data-corruption / lying status):**

1. Restore order bug (settings.py:812 then :867-872). DB is replaced via src_conn.backup(dst_conn) first, then the MFA key is copied. If shutil.copy2 on the key throws OSError (which you explicitly wrap), the DB now holds the backup's encrypted rows but the old key is still on disk — every encrypted row becomes unrecoverable. The 500 message gives no hint of partial restore. Suggest copying the key before the DB swap, or staging everything and atomic-renaming. At minimum, the error message must explicitly say "DB was overwritten but key restore failed".
2. Legacy migration rolls back all on any single row failure (database.py:286-291). One poison row blocks every successful re-encryption on every startup forever. Suggest committing per-row (or per-savepoint) and surfacing the failure count via /encryption-status.
3. Outer except Exception swallows migration failures (database.py:301-302). Startup-fatal errors (session creation, model imports, query timeouts) leave rows plaintext with no operator-visible banner. Re-raise after logging, or persist a last_migration_failed flag the status endpoint can surface.
4. decryption_broken is false-negative when the loaded key is wrong (schemas/auth.py:509-512). The flag only checks "key configured + encrypted rows present" — never attempts an actual decrypt. If an operator pastes a different valid Fernet key (rotation, cross-deployment restore, env override), the status card shows GREEN while every encrypted row is unrecoverable. Sample-decrypt one row in a try/except RuntimeError and surface a distinct "wrong-key" state.

  **Should-fix before merge:**

- TOCTOU on key file write (encryption.py:84-86, same pattern in settings.py:872): write_text honors umask (typically 0o644) and chmod(0o600) only narrows after the file exists. Co-tenant processes have a window to read the Fernet key. Use os.open(path, os.O_WRONLY | os.O_CREAT | os.O_EXCL, 0o600) + os.fdopen to write atomically with restrictive mode from the start.
- Endpoint permission contradicts CHANGELOG (auth.py:1484): /auth/encryption-status is gated on Permission.SETTINGS_READ, which Viewers have. PR description and CHANGELOG advertise it as "admin-only". Pick one — either gate on SETTINGS_UPDATE / a new SECURITY_READ, or update the docs.
- disable_totp debits failed-attempt counter on key-loss (mfa.py:660-683). When the key is gone, RuntimeError falls through to backup-code path, but record_failed_attempt still runs on every attempt — locks the account out for a server-side issue the user can't fix. Surface a distinct error and skip the failed-attempt debit when the cause is decryption failure rather than wrong code.
- setup_totp try-block too broad (mfa.py:558-576). The try/except RuntimeError wraps record_failed_attempt, clear_failed_attempts, and _assert_totp_not_replayed in addition to pyotp.TOTP(existing.secret). Future RuntimeError from any of those gets misreported as "TOTP secret unavailable". Narrow the try to just the .secret access.
- SecurityStatusCard polling stops permanently after one error (SecurityStatusCard.tsx:40). refetchInterval: error ? false : 30_000 — a single 5xx/blip kills the live status indicator until the user reloads the page. Use bounded retry with backoff or add a manual "Retry" button.
- extra = "ignore" swallows typo'd env vars silently (config.py:97). Necessary to fix the original Pydantic-rejects-MFA_ENCRYPTION_KEY issue but now  MFA_ENCYPTION_KEY=... (typo) is silently ignored — operator sees plaintext storage with no warning. Worth logging unknown MFA_* / BAMBUDDY_* keys at INFO on startup.
- mfa_decrypt "fernet: prefix missing" log demoted from WARNING to DEBUG (encryption.py:170) — hides half-migrated state from operators
- "DATA_DIR not writable" warning (encryption.py:131-138) misfires when actual cause was a corrupted key file
- chmod(0o600) silent no-op on Windows / SMB / some FUSE — key ends up world-readable; consider stat + warn
- conftest mfa_encryption_isolation docstring says "default test path is plaintext fallback" but DATA_DIR=tmp_path actually triggers auto-generation; tests have to manually patch  _load_or_generate_key to get plaintext

  **Test gaps to fill:**

- No regression test for the prefix-collision ZipSlip case — the literal reason the code was rewritten. Existing tests cover ../etc/passwd and /etc/passwd, both of which the old
  startswith check also rejected. Add a ZIP entry like tmp_<basename>_evil/file extracted into temp_path = /tmp/<basename> to pin that is_relative_to rejects what startswith would have allowed.
- No wrong-backup-code-with-hashes-seeded test for disable_totp / regenerate_backup_codes. Both 400 tests today hit the "no codes configured" branch first; if a future refactor swallows compare_digest mismatches, those tests still pass.
- No backup-and-restore round-trip on a keyless host. The user-facing promise of bundling .mfa_encryption_key into the ZIP has zero E2E coverage.

netscout2001:
Thanks for the thorough review — this round addresses every blocker, should-fix, and test gap you flagged. All changes pushed in three commits:

- `3c191a9` fix(mfa-encryption): backend hardening per review
- `56bb702` feat(ui): SecurityStatusCard resilience and migration warning + i18n parity
- `b040e0b` build(frontend): rebuild static assets after PR review fixes

---

### Blockers (data-corruption / lying status)

| # | Your finding | Fix |
|---|---|---|
| **B1** | *"Restore order bug … `src_conn.backup(dst_conn)` first, then the MFA key is copied. If `shutil.copy2` … throws OSError, the DB now holds the backup's encrypted rows but the old key is still on disk — every encrypted row becomes unrecoverable."* | The `.mfa_encryption_key` block is now executed **before** the database swap (`settings.py`). Atomic write via `os.open(O_WRONLY\|O_CREAT\|O_TRUNC, 0o600)` + `os.replace`, so the key file is never partially written. On `OSError` the handler raises `HTTPException(500, "Restore aborted: MFA key write failed. Database is unchanged. Check server logs.")` — DB is provably untouched. After the swap the encryption singleton (`_fernet_instance`, `_key_source`, `_warn_shown`) is reset so `init_db`'s migration picks up the restored key. |
| **B2** | *"Legacy migration rolls back all on any single row failure (database.py:286-291). One poison row blocks every successful re-encryption on every startup forever."* | Rewrote `_migrate_encrypt_legacy_secrets` as a two-phase, per-row strategy: **phase 1** read-only collects `(id, value)` candidates; **phase 2** opens a separate `async_session()` per row and commits independently. Failure of one row no longer blocks the others. The skipped-row count is exposed via `get_migration_error_count()` and surfaced on `/encryption-status` as `migration_error_count`. The UI renders a yellow secondary banner when `> 0` (i18n `settings.encryption.migrationErrorWarning` in all 8 locales). |
| **B3** | *"Outer `except Exception` swallows migration failures (database.py:301-302). … Re-raise after logging, or persist a `last_migration_failed` flag the status endpoint can surface."* | Both — the inner per-row catch handles individual row failures via the `migration_error_count` surface (B2), and the outer `except Exception:` now re-raises so startup-fatal errors (session creation, model imports, query timeouts) propagate to lifespan / CLI / restore handler. New test `test_init_db_propagates_unexpected_migration_error` pins the contract. |
| **B4** | *"`decryption_broken` is false-negative when the loaded key is wrong (schemas/auth.py:509-512). The flag only checks 'key configured + encrypted rows present' — never attempts an actual decrypt. … Sample-decrypt one row in a try/except RuntimeError and surface a distinct 'wrong-key' state."* | `decryption_broken` is no longer a `@computed_field` — it is a plain field filled by `/encryption-status`. When `key_configured` and `encrypted_rows > 0`, the endpoint sample-decrypts one encrypted row (`LIMIT 1`, OIDC preferred, TOTP fallback). On `RuntimeError` (wrong key or `InvalidToken`) → `decryption_broken=True`. Defensive: a `SQLAlchemyError` during the sample query also flips it true (over-alert > silent corruption). |

---

### Should-fix before merge

| Your finding | Fix |
|---|---|
| *"TOCTOU on key file write (encryption.py:84-86, same pattern in settings.py:872): `write_text` honors umask … `chmod(0o600)` only narrows after the file exists. Use `os.open(path, os.O_WRONLY \| os.O_CREAT \| os.O_EXCL, 0o600) + os.fdopen`."* | Done in both locations. `encryption.py` auto-generate uses `O_EXCL` (race-aware: `FileExistsError` returns `none_corrupted`). `settings.py` restore uses `O_TRUNC` + `os.replace` for atomic overwrite. The mode is `0o600` from creation — no TOCTOU window. |
| *"Endpoint permission contradicts CHANGELOG (auth.py:1484): `/auth/encryption-status` is gated on `Permission.SETTINGS_READ`, which Viewers have. PR description and CHANGELOG advertise it as 'admin-only'."* | Changed to `Permission.SETTINGS_UPDATE` (Admins + Operators, not Viewers). New test `test_status_returns_403_for_viewer_in_viewers_group` puts a real user into the `Viewers` group and asserts 403 — the existing `test_status_requires_settings_read_permission` (which only used a no-group user) was insufficient to discriminate the two permissions. |
| *"`disable_totp` debits failed-attempt counter on key-loss (mfa.py:660-683). … locks the account out for a server-side issue the user can't fix. Surface a distinct error and skip the failed-attempt debit when the cause is decryption failure rather than wrong code."* | Both `disable_totp` and `regenerate_backup_codes` now track a `decryption_failed` flag in the `RuntimeError` branch and skip `record_failed_attempt` when set. Existing 400-tests in `TestTOTPDecryptionBroken` were extended with `AuthRateLimitEvent` count assertions to pin the contract; new T2 tests cover the dual paths (counter increments on real wrong-code, counter stays at 0 on key-loss). |
| *"`setup_totp` try-block too broad (mfa.py:558-576). The try/except RuntimeError wraps `record_failed_attempt`, `clear_failed_attempts`, and `_assert_totp_not_replayed` in addition to `pyotp.TOTP(existing.secret)`."* | Narrowed: the `try` now only wraps the `existing.secret` property access (the single line that raises on key-loss). `secret_plain` is cached and reused, so `pyotp.TOTP(...)` and the replay-guard helpers are outside the catch. Future RuntimeErrors from those helpers will now propagate correctly instead of being misreported as "TOTP secret unavailable". |
| *"SecurityStatusCard polling stops permanently after one error … Use bounded retry with backoff or add a manual 'Retry' button."* | Both. `refetchInterval` returns `30_000` on success, then `5s/10s/15s` on consecutive `fetchFailureCount` (1/2/3), then `false` to stop hammering. The error branch now renders a manual `Retry` button (`encryption-retry-button`) that calls `refetch()`. Existing i18n bug fixed at the same time: `common.errorLoading` was referenced but undefined in any locale — added to all 8. New i18n key `common.retry` also added in all 8 locales. |
| *"`extra = 'ignore'` swallows typo'd env vars silently (config.py:97). Worth logging unknown `MFA_*` / `BAMBUDDY_*` keys at INFO on startup."* | After `Settings()` is constructed, `os.environ` is scanned for `MFA_*` / `BAMBUDDY_*` keys. Unknown keys (not in `settings.model_fields` and not in `_INTENTIONAL_UNSETTINGS = {MFA_ENCRYPTION_KEY, DATA_DIR, DATABASE_URL, LOG_DIR, LOG_LEVEL, BUG_REPORT_RELAY_URL}`) are logged at INFO so typos like `MFA_ENCYPTION_KEY` are surfaced. New unit suite `test_config_env_warnings.py` covers all three branches (typo / unknown bambuddy / intentional silent). |
| *"`mfa_decrypt` 'fernet: prefix missing' log demoted from WARNING to DEBUG (encryption.py:170) — hides half-migrated state from operators."* | Restored to `logger.warning(...)`. After the B2 migration, normal operation has no plaintext rows so this is non-spammy in steady state; if half-migrated rows persist, the `migration_error_count` surface complements the per-decrypt warning. |
| *"'DATA_DIR not writable' warning misfires when actual cause was a corrupted key file."* | Internal `_load_or_generate_key` now returns one of `none_write_failed`, `none_corrupted` (or `none` for legacy stubs). `_get_fernet` only emits the "DATA_DIR not writable" warning when source is `none_write_failed`. Corrupted-file errors are already logged with their own descriptive message in `_load_or_generate_key` — no duplicate misleading warning. Public `get_key_source()` continues to return `"env"\|"file"\|"generated"\|"none"`. |
| *"`chmod(0o600)` silent no-op on Windows / SMB / some FUSE."* | After every `chmod(0o600)` (auto-generate path and restore path), the file's actual `stat().st_mode` is checked. If `(actual & 0o777) != 0o600` a WARNING is logged so operators on Windows / SMB / FUSE without uid mapping notice that the OS did not enforce the requested permissions. |
| *"conftest `mfa_encryption_isolation` docstring says 'default test path is plaintext fallback' but `DATA_DIR=tmp_path` actually triggers auto-generation."* | Docstring corrected — explicitly notes that the default path is `key_source='generated'`, and that tests needing the plaintext fallback must monkeypatch `_load_or_generate_key` to return `(None, 'none')` (or `'none_write_failed'` / `'none_corrupted'`). |

---

### Test gaps

| Your finding | Coverage added |
|---|---|
| *"No regression test for the prefix-collision ZipSlip case — the literal reason the code was rewritten. Existing tests cover `../etc/passwd` and `/etc/passwd`, both of which the old `startswith` check also rejected."* | New test `test_restore_rejects_prefix_collision_zipslip` in `TestBackupKeyFiles`: ZIP entry whose resolved path would have been startswith-positive but `is_relative_to`-negative. Asserts 400 + `unsafe path` detail. |
| *"No wrong-backup-code-with-hashes-seeded test for `disable_totp` / `regenerate_backup_codes`. Both 400 tests today hit the 'no codes configured' branch first; if a future refactor swallows `compare_digest` mismatches, those tests still pass."* | Three new tests in `TestTOTPDecryptionBroken`: `test_disable_totp_wrong_code_with_seeded_hashes_returns_400_and_debits_counter`, `test_regenerate_backup_codes_wrong_code_with_seeded_hashes_returns_400_and_debits_counter`, and `test_disable_totp_wrong_code_with_seeded_hashes_at_keyloss_no_counter_debit`. Each seeds 10 real backup-code hashes via `_generate_backup_codes()` and exercises the actual `pwd_context.verify` mismatch path. The two paired tests pin the S3 contract: counter increments on real wrong-code, counter stays at 0 on key-loss. |
| *"No backup-and-restore round-trip on a keyless host. The user-facing promise of bundling `.mfa_encryption_key` into the ZIP has zero E2E coverage."* | New test `test_backup_restore_roundtrip_preserves_encrypted_oidc_secret` in `TestBackupKeyFiles`: pin a key, encrypt an OIDC secret via the property setter, run the real `create_backup_zip`, verify the ZIP contains `.mfa_encryption_key`, simulate full key loss (delete file + drop env var + reset singleton), POST `/api/v1/settings/restore`, then read the OIDC client_secret and assert it decrypts to the original plaintext. |

---

### Plan-level guarantees pinned by tests

In addition to the gap-driven tests above, three new tests were added to anchor the new contracts so future refactors cannot silently regress:

- **B1 singleton-reset** — `test_restore_resets_encryption_singleton_after_key_replace`
- **B1 abort semantics** — `test_restore_aborts_db_swap_when_key_write_fails`
- **B3 re-raise** — `test_init_db_propagates_unexpected_migration_error`
- **B4 wrong-key detection** — `test_status_decryption_broken_when_wrong_key_active`
- **B2 counter surface** — `test_status_surfaces_migration_error_count`
- **S5 polling backoff** — `polling stops after 3 consecutive errors` (Vitest)
- **S5 manual recovery** — `renders Retry button in error state and recovers when clicked` (Vitest)
- **S5 migration UI banner** — `renders migration error warning when migration_error_count > 0` (Vitest)
- **S6 typo detection** — 3 unit tests in `test_config_env_warnings.py`
- **S8 differentiated source** — three existing source-asserting tests adapted to `none_write_failed` / `none_corrupted`
- **B2 per-row commit semantics** — `test_migration_continues_on_row_error` inverted: previously asserted "rollback all", now asserts good-row-encrypted + bad-row-unchanged + `get_migration_error_count() == 1`

---

### Test suites — final state

| Suite | Result |
|---|---|
| `./test_frontend.sh` | **1700 / 1700 passed**, i18n parity verified across all 8 locales (4690 leaf keys each) |
| `./test_backend.sh` | **2607 passed**, **0 failed**, 30 errors in `tests/unit/services` (pre-existing `TLS_FTPHandler` import — unrelated to this PR) |
| `./test_security.sh` | bandit ✅, npm-audit ✅, pip-audit ✅ |

Happy to address anything I missed.

maziggy (approved):

PR #1246, opened 2026-05-08
[Feature] Copy filament

### LINKED ISSUE #1234: [Feature]: Clone filament
### Problem or Use Case

We typically have many filaments of the same brand and specifications, differing only in color.

It would be useful to add the option to clone a filament or use it as a template to create a new one, where we only need to specify the color and weight, with the rest of the data (brand, profile, spool type, etc.) already filled in.

### Proposed Solution

A new option on the filaments to allow create clone that show a create filament form filled

### Alternatives Considered

_No response_

### Feature Category

Spool Inventory

### Priority

Nice to have

### Mockups or Examples

_No response_

### Contribution

### Checklist

miguelangellv:
## Description

Allow copy a filament to create new one with similiar options quickly

## Related Issue

Fixes #1234 

## Documentation

- Wiki: maziggy/bambuddy-wiki#27

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition or update

## Changes Made

- Added a copy button on filaments table
- Update Spool Modal to create a copy
-

## Screenshots

image

image

## Testing

## Checklist


maziggy (changes requested):
Thanks for the contribution, @MiguelAngelLV — useful UX.

  A few things to fix before merge:

1. copy is typed as required, not optional — this breaks every existing caller that doesn't pass it. 

SpoolFormModalProps adds copy: boolean | null | undefined; (no ?). Same pattern on SpoolCard.onCopy, SpoolTableRow.onCopy, SpoolTableGroup.onCopy — all required. The 9 existing test renders in frontend/src/__tests__/components/SpoolFormModal.test.tsx don't pass copy, so they'll fail to typecheck. Same problem will hit any future call site that just wants to edit. Fix:

  copy?: boolean;   // SpoolFormModalProps
  onCopy?: () => void;   // SpoolCard / SpoolTableRow / SpoolTableGroup

  When onCopy is undefined on the card/row/group, conditionally render the copy button — same pattern onPrintLabel? already follows in SpoolTableRow.

2. Typo: isCoping → isCopying.

  The present participle of "copy" is "copying", not "coping" (which means "managing / dealing with"). Three sites in SpoolFormModal.tsx. User-visible-adjacent so worth fixing.

3. No tests added.

  The "Test addition or update" box isn't checked, and there are no test files in changed_files. The new code branches in non-trivial ways (isEditing vs isCopying vs neither), so please add at least:

- A SpoolFormModal test asserting that when spool and copy=true are passed, the title reads "Copy Spool" and Save calls api.createSpool (not api.updateSpool).
- A test asserting weight_used is 0 in the create payload when copying a spool whose original weight_used was non-zero (this is the actual user-facing benefit — verifies the form doesn't carry stale usage data into a fresh spool).
- An InventoryPage test asserting the copy button on a card / row dispatches with copy: true and the original spool object — guards against future regressions where someone refactors the modal-state object shape.

The contributor's "tested locally" claim is fine for sanity but doesn't guard against regressions in CI.

4. (Style) Consider mode instead of copy.

  Holding two booleans (isEditing, isCopying) that derive from spool + copy is fine but error-prone — a future change that adds e.g. a "duplicate to Spoolman" mode adds a third boolean. Cleaner shape:

  mode?: 'create' | 'edit' | 'copy';   // defaults to spool ? 'edit' : 'create'

  Same idea, fewer states to reason about. Not a merge blocker — happy with the current shape if you'd rather keep the diff small.

5. Save button JSX has stray indentation in the rebuild.

  The diff shows the <Save /> block under the loading-spinner ternary's else branch lost a space of indent (+                 <Save instead of +                <Save). Probably your
  editor's auto-format. Pure cosmetic — Prettier will likely fix it on save.

6. Spoolman-mode coverage looks correct.

  createMutation correctly routes to api.createSpoolmanInventorySpool when spoolmanMode is true (line 399-400 unchanged). The copy flow inherits this naturally because it goes through the same submit path. Worth a manual smoke test to confirm — Spoolman doesn't have a weight_used column the same way local-DB does, so the weight_used: 0 reset is a no-op there but harmless.

miguelangellv:
>mode?: 'create' | 'edit' | 'copy'; // defaults to spool ? 'edit' : 'create'

I've chosen to make the mode mandatory and update all calls. Since TypeScript will check if the value isn't passed in any call, this should prevent future problems.



I have also added the tests and the rest of the requested changes.

maziggy:
Thanks!

Issue #1182, opened 2026-04-30
[Feature]: Allow API Keys to access cloud presets


turulix:
### Problem or Use Case

I am currently working on building a fully automated, headless pipeline to manage my 3D prints. A crucial step in this workflow is automatically slicing the 3D models on my server before sending them to the printer.

However, as discussed in #665, the current APIs are unable to access cloud data. Because the API cannot interact with the cloud to retrieve user presets.

### Proposed Solution

The most straight forward way would be to allow API keys to access cloud data of the user that has created them

### Feature Category

API

### Priority

Critical for my use case

### Contribution

- [x] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
You didn't fill out the bug report completely. I'm missing important details.

turulix:
What details do you still need?

maziggy:
The bug report form has all required infos.

turulix:
I made this using the feature request form, and these are all the infos it asked for, this is not a bug and was done on purpose in #665

maziggy:
I'm sorry, mixed up the issues. Need to grab a coffee first.

maziggy:
Available/Fixed in branch dev and available with the next release or daily build. Please let me know if it works for you.

Docs -> https://wiki.bambuddy.cool/features/api-keys/

-----
If you find Bambuddy useful, please consider giving it a ⭐ on [GitHub](https://github.com/maziggy/bambuddy) — it helps others discover the project!

turulix:
Got around to testing it thanks for the fast implementation, its not quite working as i need it, i can now successfully get the needed ids from:
`/api/v1/cloud/settings`

Sending a request to  `/v1/library/files/<file_id>/slice` with the payload:
```json
{
    "printer_preset": {
        "source": "cloud",
        "id": "PMUS3f3e..."
    },
    "process_preset": {
        "source": "cloud",
        "id": "PPUS17c03..."
    },
    "filament_presets": [
        {
            "source": "cloud",
            "id": "PFUS594..."
        }
    ]
}
```
successfully starts a slicing task, but checking the status of the slicing task under `/api/v1/slice-jobs/<job_id>`
results in:

```json
{
    "job_id": 1,
    "status": "failed",
    "kind": "library_file",
    "source_id": 13,
    "source_name": "test.3mf",
    "created_at": "2026-05-01T19:03:30.049578+00:00",
[...]

PR #933, opened 2026-04-10
feat: Two-Factor Authentication (TOTP, Email OTP) and OIDC/SSO – full implementation with admin UI

netscout2001:
### Description

Adds a complete multi-factor authentication and single sign-on system to BamBuddy, accompanied by an extensive security hardening pass (~50 findings across four review rounds).

### Related Issue
Features <https://github.com/maziggy/bambuddy/issues/902>

## Documentation

**Companion docs PRs** (delete lines that don't apply):
- Wiki: maziggy/bambuddy-wiki#___
- Website: maziggy/bambuddy-website#___

**Pick one**:
- [ ] Docs PR(s) linked above
- [ ] No docs update required — reason: bundling with PR initially; will add wiki docs if maintainers want to proceed

## Type of Change

- [x]  New feature
- [ ]  Bug fix
- [x]  Security improvement
- [x]  Test addition or update

---

#### Two-Factor Authentication

| Feature | Details |
|---|---|
| TOTP | Compatible with Google Authenticator, Authy, 2FAS and any standard app |
| Email OTP | Alternative second factor via email |
| Backup codes | 10 single-use recovery codes per user |
| Opt-in | Per-user; each user can enable 2FA independently |
| Brute-force protection | Rate limiting on all verification and enable endpoints |

#### OIDC / Single Sign-On

| Feature | Details |
|---|---|
| Providers | PocketID, Authentik, Keycloak and any standards-compliant OIDC provider |
| PKCE (S256) | Works with public clients that have no client secret |
| Account linking | Existing users linked automatically via verified email on first SSO login (opt-in, disabled by default) |
| Auto-provisioning | Optional: create new BamBuddy accounts on first SSO login |
| Login page | SSO buttons rendered dynamically per configured provider |

---

#### Security Hardening

**Input validation (pbkdf2 / pyotp DoS guard)**

| Finding | Fix |
|---|---|
| Multiple request schemas had no `max_length` on string fields — unbounded input reached pbkdf2 or pyotp | `max_length` enforced on all user-facing fields before any hashing or OTP call: `LoginRequest`, `SetupRequest`, `UserCreate`, `UserUpdate`, `ChangePasswordRequest`, `ForgotPasswordRequest`, `TOTPSetupRequest`, `EmailOTPEnableConfirmRequest`, `OIDCProviderCreate/Update` |
| Email OTP setup code accepted any string | `code` now validated as exactly 6 digits before reaching any verification logic |
[...]

PR #1295, opened 2026-05-12
fix(auth): cleanup orphan OIDC/MFA rows when user is deleted (#1285)

### LINKED ISSUE #1285: [Bug]: SSO/OIDC deleted user recreation not working
### Component

Bambuddy

### Bug Description

Hi there,
i've tried to test the new version of OIDC integration- but i got a new error on recreating deleted users which relogins with SSO to bambuddy.

### Expected Behavior

Deleted users should reappear/should get recreated in userlist of Bambuddy after deletion and relogin with IdP as long they're enabled in IdP

### Steps to Reproduce

1. login as a new user with SSO to bambuddy 
2. logout 
3. as admin: delete new created account from userlist in Authentication>Users in Bambuddy 
4. relogin as user before 
5. user doesn't reappear in userlist, 
6. user is not able to login anymore 

### Printer Model

None

### Bambuddy Version

v0.2.4

### SpoolBuddy Version

_No response_

### Printer Firmware Version

_No response_

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

relevant container logs regarding login process:
DELETE /api/v1/users/14?delete_items=false HTTP/1.1" 204 No Content
"POST /api/v1/printers/camera/stream-token HTTP/1.1" 401 Unauthorized

### Screenshots

_No response_

### Additional Context

_No response_

### Checklist

netscout2001:
## Description

Fixes a bug where users created via OIDC SSO cannot be re-created after an admin deletes them. On SQLite (default deployment) the `delete_user` endpoint left orphan rows in `user_oidc_links`, `user_totp`, and `user_otp_codes` because the project intentionally runs with `PRAGMA foreign_keys=OFF`, so the `ON DELETE CASCADE` declared on these tables never fires. The subsequent OIDC callback then found the orphan link, failed to resolve the deleted user, and redirected to `account_inactive` instead of triggering `auto_create_users`.

The fix mirrors the existing APIKey-cleanup pattern in `delete_user` (added in PR #1182, three weeks after these tables were introduced in PR #933 — which is how the gap arose), plus an idempotent startup migration so installations carrying orphans from before the fix are healed automatically without manual DB intervention. A defensive null-check in `list_oidc_links` covers the symmetric edge case where a `UserOIDCLink` could reference an orphaned provider.

PostgreSQL deployments are unaffected — cascade was already firing there. The new migration is a no-op on PG.

## Related Issue

Fixes #1285

## Documentation

**Pick one**:
- [x] No docs update required — reason: Bug fix only. Restores the documented behavior of `auto_create_users` (and other documented OIDC re-login flows) on SQLite. No new config keys, endpoints, UI, installation steps, or user-visible feature surfaces.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition or update

## Changes Made

**Backend code (3 files):**
- `backend/app/api/routes/users.py` — `delete_user` now explicitly deletes `UserOIDCLink`, `UserTOTP`, and `UserOTPCode` rows owned by the user, mirroring the established `APIKey` cleanup block. Comment explains the SQLite/`foreign_keys=OFF` rationale.
- `backend/app/core/database.py` — `run_migrations()` gains an idempotent orphan-cleanup block that sweeps rows in those three tables where `user_id` no longer exists in `users`. Wrapped in `begin_nested()`, logged via `logger.info` only when rows are actually removed. No-op on PostgreSQL.
- `backend/app/api/routes/mfa.py` — `list_oidc_links` returns `"<deleted>"` for `provider_name` when `link.provider` is null instead of raising `AttributeError`.

**Tests (5 files, 14 new/extended tests):**
- `backend/tests/integration/test_users_auth_cleanup.py` *(new)* — 4 tests verifying `delete_user` removes OIDC/TOTP/OTP rows individually and combined.
- `backend/tests/integration/test_oidc_relogin.py` *(new)* — full end-to-end test reproducing the #1285 symptom: mocked IdP, first OIDC login, admin delete, second OIDC login proves `auto_create_users` fires again.
- `backend/tests/unit/test_orphan_auth_cleanup_migration.py` *(new)* — 6 tests for per-table cleanup, idempotency, no-op on fresh install, and survival of rows belonging to real users.
- `backend/tests/integration/test_mfa_api.py` — adds `TestListOidcLinksDefensiveProviderNull` for the null-check.
- `backend/tests/integration/test_auth_api.py` — extends existing `test_delete_user` with explicit assertions on all four auth-table side effects (`UserOIDCLink`, `UserTOTP`, `UserOTPCode`, `APIKey`).

## Screenshots

N/A — backend-only fix, no UI changes.

## Testing

- [ ] I have tested with my printer model: N/A — auth/DB-only change, no printer interaction

**Test execution per CONTRIBUTING.md:**

`./test_backend.sh` equivalent — `ruff check` + `ruff format --check` on the full `backend/` tree pass cleanly. `pytest -v -n 30` runs the full backend suite: **4585 passed**, plus the 14 new tests added by this PR all green. The remaining test failures observed on Windows (`test_bambu_ftp.py`, `test_spoolbuddy_*`, file-permission tests) are pre-existing environment-specific issues unrelated to this change — they touch areas this PR does not modify.

`./test_frontend.sh` — `tsc --noEmit` clean, ESLint clean, **1891/1891 Vitest tests passed**, i18n parity check passes for all 8 locales (4752 keys each).

End-to-end verification: the new `test_oidc_relogin.py` integration test reproduces exactly the steps from the issue report (login via SSO → delete → re-login) and asserts a new user row is created, with a fresh `UserOIDCLink` pointing at it. Verified that without the `delete_user` change, the test fails at the second callback with the `account_inactive` redirect — confirming the test covers the regression boundary.

## Checklist


netscout2001:
@maziggy 
## Additional Notes

**Why three tables in one PR, not just `user_oidc_links`:** All three were introduced in the same PR (#933) before the `APIKey` cleanup pattern existed (PR #1182). They share the exact same root cause and one-line fix in `delete_user`. Splitting them would mean two follow-up issues for what is structurally a single pattern-drift problem. `UserTOTP` in particular is security-relevant (orphan MFA secrets persisting in the DB), so addressing it together with the reported symptom seems strictly better than leaving it open.

**Scope discipline:** The PR does *not* touch `long_lived_tokens`, `user_email_pref`, or `user_groups`, which share the same root cause but were authored by other contributors. Those are intentionally left for a separate audit-PR / discussion.

**Followup considerations (not in this PR):**
- A broader architectural decision around whether to enable `PRAGMA foreign_keys=ON` for SQLite globally would eliminate the need for these manual cleanups across all FK-bearing tables. Worth a separate discussion thread.

---

maziggy (changes requested):
## Blocker

  `long_lived_tokens` has the same problem and is not addressed in this PR. Look at backend/app/models/long_lived_token.py:50-55:

      user_id: Mapped[int] = mapped_column(
          Integer,
          ForeignKey("users.id", ondelete="CASCADE"),
          nullable=False,
          index=True,
      )

  Identical declaration to UserOIDCLink / UserTOTP / UserOTPCode. Identical SQLite-FK-off behaviour: deleting a user leaves orphan camera-stream tokens whose secret_hash
  is still valid. The PR's stated scope ("remove orphan auth rows on user delete") should cover it — and the description's own framing of why the gap arose ("PR #1182 added APIKey cleanup three weeks after these tables were introduced in PR #933") applies just as cleanly to long_lived_tokens, which post-dates both.

Mechanical fix:

  - Add `await db.execute(delete(LongLivedToken).where(LongLivedToken.user_id == user_id))` in `delete_user` alongside the three new DELETEs.
  - Add a fourth `DELETE FROM long_lived_tokens WHERE user_id NOT IN (SELECT id FROM users)` block in the migration.
  - One more `test_delete_user_removes_long_lived_tokens` integration test + one more migration table test.

Worth closing the whole class in this PR rather than as a follow-up.

## Nits

  **1. Migration test's hand-rolled model registry has already drifted.** `test_orphan_auth_cleanup_migration.py:_register_all_models()` imports models by hand and is missing six models that exist in `backend/app/models/__init__.py` (github_backup, local_preset, notification, notification_template, orca_base_cache, pending_upload). One-line replacement keeps it in sync forever:

      def _register_all_models():
          import backend.app.models  # noqa: F401  # __init__.py registers all tables in Base.metadata

**2. OIDC test fixtures bypass the encryption setter.** `test_users_auth_cleanup.py` constructs `OIDCProvider(_client_secret_enc="cleanup_secret", ...)` directly,
  writing the raw column instead of going through the `client_secret` property setter (which calls `mfa_encrypt`). Harmless for these tests since nothing decrypts, but
  brittle if encryption-dependent assertions land later. Use:

      provider = OIDCProvider(name="CleanupProv", issuer_url=..., client_id=..., scopes=..., is_enabled=True)
      provider.client_secret = "cleanup_secret"
      db_session.add(provider)

**3. Adjacent same-class bug.** `print_batches.created_by_id` (print_batch.py:33) is `ondelete="SET NULL"` but isn't in the existing SET-NULL block at users.py:393-407.
   Same SQLite-FK-off root cause as the columns already handled there — admin-deleted users leave dangling `created_by_id` in `print_batches`. Out of scope for #1285 (not auth), but happy for it to be folded in here since the SET-NULL block is right there; otherwise a tracking issue is fine.

netscout2001:
Thanks for the thorough review — all four points addressed in `4918c207`:

| # | Item | Status | Where |
|---|---|---|---|
| Blocker | Extend cleanup to `long_lived_tokens` | ✅ Done | `users.py:delete_user`, `database.py:run_migrations`, plus `test_delete_user_removes_long_lived_tokens` (integration) + `test_migration_deletes_orphan_long_lived_tokens` (unit) + the combined-cleanup test + the keeps-rows-for-existing-users migration test all extended |
| Nit 3 | `print_batches.created_by_id` in SET NULL block | ✅ Folded in | `users.py:393-407` — added to both `delete_items=True` and the SET NULL branch |
| Nit 1 | Module-import registry instead of hand-rolled list | ✅ Done with caveat | Replaced with `import backend.app.models` — but `__init__.py` doesn't re-export 7 submodules (`virtual_printer`, `slot_preset`, `project_bom`, `print_queue`, `spoolman_k_profile`, `spoolman_slot_assignment`, `external_link`) whose tables `run_migrations` touches. Imported those by submodule with a comment so the test engine still gets the full schema. Reduces maintenance from ~35 imports to 7. Happy to open a follow-up to re-export those from `__init__.py` so the test can be the pure one-liner you suggested. |
| Nit 2 | `OIDCProvider` fixtures use encryption setter | ✅ Done | Both fixtures in `test_users_auth_cleanup.py` now do `provider.client_secret = "..."` instead of `_client_secret_enc=...` |

**Verification:**
- 15/15 auth-related tests pass (4 in `test_users_auth_cleanup.py` + 1 E2E in `test_oidc_relogin.py` + 7 in `test_orphan_auth_cleanup_migration.py` + extended `test_delete_user` + `test_list_oidc_links_returns_deleted_marker_for_orphan_provider`)
- `ruff check` + `ruff format --check` on `backend/` clean
- `bandit --severity-level medium`: 0 medium, 0 high

**On the history:** I had a local merge of `upstream/dev` ready when you used the "Update branch" button on the PR — to keep the graph clean I reset to your merge commit (`33a07a14`) and cherry-picked the review-feedback commit on top with `--force-with-lease`. So the branch now reads as your merge + one follow-up commit, no double-merge.

maziggy (approved):

PR #1330, opened 2026-05-14
fix(spoolman): filter external library lookup by Bambu Lab manufacturer

### LINKED ISSUE #1309: [Bug]: Spoolman filament auto-match picks competitor names (e.g. "3DXTECH™ Black") for Bambu Lab RFID spools
### Bug Description

When Bambuddy auto-creates a filament entry in Spoolman for a Bambu Lab RFID spool, the second-stage lookup against Spoolman's external library (`GET /api/v1/external/filament`, served from [SpoolmanDB](https://donkie.github.io/SpoolmanDB/filaments.json)) matches on `material + color_hex` only — there is no `manufacturer` / `vendor` filter. SpoolmanDB is a multi-vendor catalog (~6,957 entries, 272 of them Bambu Lab) and entries are roughly ID-sorted, so the first hit for a common combination is almost always some other manufacturer.

Concrete example: for PLA + `000000` (black), SpoolmanDB contains 64 entries; the first hit is `3djake_pla_black_1000_175_n` (`3DJAKE`), the third is `3dxtech_pla_carbonxcarbonfiberblack_500_175_p` (`3DXTECH`, name `CarbonX™ Carbon Fiber Black`), and the actual `bambulab_pla_black_1000_175_n` (`Bambu Lab`, name `Black`) is the 15th. Bambuddy therefore creates a filament under the Bambu Lab vendor but labels it with a competitor's name.

Real-world observations in production:

- Bambu Lab ABS Black → created as `3DXTECH™ Black` under the Bambu Lab vendor (should be `Black` from `bambulab_abs_black_1000_175_n`).
- Bambu Lab PLA Support → adjacent / wrong variant picked instead of `bambulab_pla_supportforpla/petgblack_500_175_n`.
- Bambu Lab PLA Basic Black → created as `PLA` (material, not `PLA Basic`) under the Bambu Lab vendor, with the PLA-default 1.24 density fallback rather than the value carried by the matched external entry.

A secondary issue compounds this: `_create_filament_from_external` drops the `density` field from the external entry, so even when the right entry is picked the density gets overwritten by `create_filament`'s built-in PLA default.

### Expected Behavior

For an RFID-identified Bambu Lab spool, the auto-created filament name should come from a Bambu-Lab-only candidate set:

1. First, an existing Bambu Lab filament in Spoolman matching `material + color_hex` (already correct in the code).
2. Otherwise, an entry from the external library whose `manufacturer == "Bambu Lab"` (or whose `id` starts with `bambulab_`) matching `material + color_hex`. If multiple Bambu entries match, prefer the one whose `name` equals the AMS `tray_sub_brands` (so `PLA Basic`, `Support for PLA/PETG Black`, etc. are preserved when possible).
3. Otherwise, fall back to `tray_sub_brands or tray_type` and create a new filament under the Bambu Lab vendor.

Density / other physical properties from the chosen external entry should also be carried over, not silently replaced with the PLA-default 1.24.

### Steps to Reproduce

1. In Spoolman, delete (or do not yet have) any filament whose vendor is Bambu Lab, material `PLA`, and color `#000000`.
2. Make sure Spoolman's external DB has synced (default behaviour; `https://donkie.github.io/SpoolmanDB/filaments.json`).
3. Insert a genuine Bambu Lab PLA Basic Black spool into the AMS so its RFID is read.
4. Trigger Bambuddy → Spoolman sync.

**Actual result:** a new filament is created under the Bambu Lab vendor with `name="3DJAKE Black"` (or `"3DXTECH™ Black"` for ABS black; the exact "winner" depends on dict ordering in `filaments.json`), instead of `"Black"` from `bambulab_pla_black_1000_175_n`. The density is also the PLA-default 1.24 rather than whatever the external entry carries.

#### Root cause

`backend/app/services/spoolman.py`, method `_find_or_create_filament` (around L1025-1072 on `main`). The internal-filament loop (L1031-1043) correctly filters by `bambu_vendor_id`, but the external-library loop (L1045-1052) does not:

```python
# Search external filaments (Bambu library)
external = await self.get_external_filaments()
for filament in external:
    fil_material = filament.get("material") or ""
    fil_color = filament.get("color_hex") or ""
    if fil_material.upper() == tray.tray_type.upper() and fil_color.upper() == color_hex.upper():
        # Found in external library - need to create internal copy
        return await self._create_filament_from_external(filament, tray)
```

The comment claims "(Bambu library)" but the endpoint actually returns the full multi-vendor SpoolmanDB. The Spoolman endpoint (`spoolman/api/v1/externaldb.py`) is a plain `FileResponse` of `filaments.json` with no query-parameter filtering, so the only fix is client-side filtering on the `manufacturer` field (which is required and always present in the `ExternalFilament` schema).

`_create_filament_from_external` additionally drops `density` (and `external_id`), causing the typical-density fallback in `create_filament` to be used even when the external entry has a more accurate value.

#### Suggested fix

```suggestion
    async def _find_or_create_filament(self, tray: AMSTray) -> dict | None:
        """Return a Bambu Lab filament matching the tray's material/color, creating it if absent."""
        bambu_vendor_id = await self.ensure_bambu_vendor()
        color_hex = tray.tray_color[:6]
        material_upper = tray.tray_type.upper()
        color_upper = color_hex.upper()

        # Search internal filaments - only match Bambu Lab vendor
        filaments = await self.get_filaments()
        for filament in filaments:
            fil_vendor_id = filament.get("vendor_id") or filament.get("vendor", {}).get("id")
            if fil_vendor_id != bambu_vendor_id:
                continue
            fil_material = filament.get("material") or ""
            fil_color = filament.get("color_hex") or ""
            if fil_material.upper() == material_upper and fil_color.upper() == color_upper:
                return filament

        # Search external filaments (SpoolmanDB) — restrict to Bambu Lab only.
        # The endpoint returns the full multi-vendor catalog with no server-side
        # filter, so without a manufacturer check the first PLA/black hit is
        # typically 3DJAKE or 3DXTECH, not Bambu Lab.
        external = await self.get_external_filaments()
        sub_brand = (tray.tray_sub_brands or "").strip().lower()
        bambu_candidates = []
        for filament in external:
            manufacturer = (filament.get("manufacturer") or "").strip().lower()
            ext_id = (filament.get("id") or "").strip().lower()
            if manufacturer != "bambu lab" and not ext_id.startswith("bambulab_"):
                continue
            fil_material = filament.get("material") or ""
            fil_color = filament.get("color_hex") or ""
            if fil_material.upper() == material_upper and fil_color.upper() == color_upper:
                bambu_candidates.append(filament)

        if bambu_candidates:
            # Prefer entries whose `name` matches the AMS sub-brand (e.g. "PLA Basic",
            # "Support for PLA/PETG Black") so the more specific variant wins.
            chosen = next(
                (f for f in bambu_candidates if (f.get("name") or "").strip().lower() == sub_brand),
                bambu_candidates[0],
            )
            return await self._create_filament_from_external(chosen, tray)

        return await self.create_filament(
            name=tray.tray_sub_brands or tray.tray_type,
            vendor_id=bambu_vendor_id,
            material=tray.tray_type,
            color_hex=color_hex,
            weight=tray.tray_weight,
        )
```

And in `_create_filament_from_external`, propagate `density` (and optionally `external_id` if `create_filament` is extended to accept it):

```suggestion
    async def _create_filament_from_external(self, external: dict, tray: AMSTray) -> dict | None:
        """Create an internal Spoolman filament from an external library entry."""
        vendor_id = await self.ensure_bambu_vendor()
        return await self.create_filament(
            name=external.get("name", tray.tray_sub_brands),
            vendor_id=vendor_id,
            material=external.get("material", tray.tray_type),
            color_hex=external.get("color_hex", tray.tray_color[:6]),
            weight=external.get("weight", tray.tray_weight),
            density=external.get("density"),
        )
```

#### Scope of impact

- Any user running Spoolman with the default external DB (SpoolmanDB) enabled.
- Any Bambu Lab RFID spool whose `material + color_hex` is also offered by another manufacturer in SpoolmanDB — i.e. essentially every common color (PLA black, white, gray; ABS black; PETG black/white; etc.).
- Affects the auto-create path only; manual links and re-uses of existing Bambu-Lab-vendor filaments are unaffected. Previously-created mislabeled filaments will keep being reused (because step 1 short-circuits on `vendor + material + color_hex`), so users have to delete the wrongly-named filament once for the corrected name to be created.
- Frontend "Filament Catalog Picker" (PR #1114) is unrelated and not regressed by this change.

#### Environment

- **Component:** Bambuddy
- **Bambuddy version:** v0.2.4 (also present on `main` at the time of writing; see source references above)
- **Installation:** Docker (`ghcr.io/maziggy/bambuddy:latest`)
- **OS:** Linux (Docker Desktop on Windows 11 + WSL2 backend)
- **Printer / firmware:** Bambu Lab H2S, firmware `01.02.00.00` (RFID-related, not printer-side specific)
- **Spoolman:** `ghcr.io/donkie/spoolman:latest`, with default external DB enabled (`https://donkie.github.io/SpoolmanDB/filaments.json`)

#### Search for duplicates

I searched the existing issues and PRs (`filament name`, `spoolman vendor manufacturer`, `external library`, `3DXTECH`, `filament wrong name`, `filament match`, etc.) and could not find a prior report of this specific bug. #149 is adjacent (linking pre-registered spools) but does not touch the naming logic. PR #1114 is unrelated.

Happy to be assigned and open a PR with the suggested change if you agree with the direction.

ojimpo:
## Description

Fix `_find_or_create_filament` so Bambu Lab RFID spools no longer get labeled with competitor names (e.g. `3DXTECH™ Black`, `3DJAKE Black`) when the external SpoolmanDB lookup is consulted.

Bambuddy's external-library loop matched on `material + color_hex` only, with no manufacturer filter. SpoolmanDB is a multi-vendor catalog sorted roughly by ID, so the first hit for any common combination is almost always a competitor — `bambulab_pla_black_1000_175_n` is the 15th entry for PLA + `#000000`. The Bambu Lab vendor entry was being assigned a competitor's product name (and density).

The fix:

1. Restrict the external-library loop to entries whose `manufacturer == "Bambu Lab"` (or whose `id` starts with `bambulab_` as a defensive fallback).
2. When multiple Bambu Lab candidates exist, prefer the entry whose `name` equals the AMS `tray_sub_brands` (e.g. `"PLA Basic"`) so the more specific variant wins over a generic one.
3. Forward `density` from the chosen external entry to `create_filament` instead of letting it get overwritten by the PLA-default 1.24 fallback inside `create_filament`.

## Related Issue

Fixes #1309

## Documentation

**Pick one**:
- [x] No docs update required — reason: silent bug fix from the user's perspective. The auto-create path still triggers, only the chosen `name` / `density` change. No config keys, ports, URLs, or installation steps affected. The one operationally relevant note (existing mis-named filaments are not auto-renamed) is in *Additional Notes* below, per maintainer request on #1309.

## Type of Change

## Changes Made

- `backend/app/services/spoolman.py::_find_or_create_filament`: filter external-library candidates by `manufacturer == "Bambu Lab"` (with `id.startswith("bambulab_")` fallback for schema drift); when multiple Bambu candidates match, prefer one whose `name` equals `tray_sub_brands`.
- `backend/app/services/spoolman.py::_create_filament_from_external`: propagate `density` from the external entry to `create_filament`.
- `backend/tests/unit/services/test_spoolman_service.py::TestFindOrCreateFilament`: six new unit tests covering the new behavior — internal short-circuit preserved, non-Bambu external entries skipped, `PLA Basic > generic PLA` tiebreaker (per maintainer request), no-match fallback, `id` prefix defensive fallback, density propagation.

## Screenshots

N/A (backend logic; no UI change).

## Testing

- [x] I have tested this on my local machine — `./test_backend.sh` clean (ruff + format + full pytest -n 30).
- [x] I have tested with my printer model: Bambu Lab H2S (the spool that triggered #1309 was a PLA Basic Black on this printer; locally the misnamed `filament_id` "Bambu Lab Black PLA / 1.24" entry that Bambuddy had auto-created was the same bug — once deleted, the next AMS read with this fix re-creates it as `Bambu Lab Black PLA Basic / 1.26`).

## Checklist

## Additional Notes

**Existing mis-named filaments are NOT auto-renamed.** Step 1 of `_find_or_create_filament` is the internal-Spoolman-filaments loop that short-circuits on `(vendor = Bambu Lab, material, color_hex)`. Any Bambu Lab filament created by an older Bambuddy build (or hand-edited by the user) will continue to be matched and reused on subsequent AMS reads, regardless of how wrong its name is. To pick up the corrected name, the user has to **delete the mis-named filament once** — then the next AMS read for the same material+color falls through to the external-library step and creates a new entry with the correct Bambu Lab name.

This is deliberate. Some users may have intentionally renamed Bambu Lab filaments (e.g. to follow their own naming convention or to merge variants) and a silent auto-rename would undo that. The one-time manual delete is a small cost; auto-renaming user-edited records is a much bigger one.

One follow-up worth flagging but not in scope of this PR: when an AMS read arrives for material+color X and a wrongly-named filament for the same X already exists internally, Bambuddy doesn't surface anywhere that the new code path *would* have picked a different name. If that becomes a frequent support question, a startup-time scan that flags such candidates without modifying them could be a small future PR.


maziggy (approved):

maziggy:
Very clean work! Thanks :)

Issue #1289, opened 2026-05-12
[Feature]: Allow Self Signed Certificates for access to Home Assistant


WizBangCrash:
### Problem or Use Case

I have an internal LAN that uses self signed certificates for SSL access to my internal endpoints.
I would like the option to upload my CA to Bambuddy so that I can access my local instance of Home Assistant., which only supports SSL connections.
I have tried manually installing the CA on the docker instance but this does not appear to work. Testing the HA connection always seems to fail.

### Proposed Solution

I would like the option to upload my self signed CA to Bambuddy via the Network Settings Panel.
I would also like to upload an SSL certificate for Bambuddy so that I can access it directly via https and not use a reverse proxy.

### Alternatives Considered

I've worked around this by putting the remote URL for Home Assistant in the settings, but would prefer all bambuddy to access my HA instance locally.

### Feature Category

Other

### Priority

Would improve my workflow

### Contribution

- [x] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Please note our rules in https://github.com/maziggy/bambuddy/blob/main/CONTRIBUTING.md. Thanks.

WizBangCrash:
@maziggy I've been looking up the behaviour of the httpx library and it looks like adding support for this feature could be quite easy.
The httpx library can use [environmental variables](https://www.python-httpx.org/environment_variables/) to stop it using the default `certifi` trust store.

I propose that the docker container be modified to recognise if the SSL_CERT_DIR environment variable is set and if so run `update-ca-certificates` at boot time.
Then, if the User's docker compose file contains the following:
```
volumes:
    - /path/to/my/certs:/usr/local/share/ca-certificates
environment:
    - SSL_CERT_DIR=/etc/ssl/certs
 ```
and the User has their own self signed CA certificate in /path/to/my/certs, the httpx library with use the container's system trust store.

I've tested this by adding the above and then manually running `update-ca-certificates` in a shell within the container.

Doing it this way ensures that everything else is still getting a valid trust store for access e.g. the Bambu Cloud profiles page etc.

I have also checked the virtual printer & the mqtt modules to see if this will be affected, but I do not believe it will because the modules use `SSLContexts` which will only make use of the environment variables if explicitly set in their definition.

I wanted to run this past you and get your thoughts before i fork the repo and set up a test environment to create a PR.

WizBangCrash:
[...]

PR #1384, opened 2026-05-16
Add support for self signed CA certs

WizBangCrash:
## Description

Updated the docker entrypoint script and the example docker compose file to include support for self signed CA certificates.

## Related Issue

Features #
https://github.com/maziggy/bambuddy/issues/1289

Fixes #

## Documentation

- Wiki: https://github.com/maziggy/bambuddy-wiki/pull/31

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Added code to detect is environment variable USE_SYSTEM_TRUST_STORE is defined and if so switch to using the system trust store and update with any certificates added in /usr/local/share/ca-certificates
- Updated docker-compose.yaml to include example
- Docker logs will be updated with status and error messages

## Screenshots

N/A

## Testing

## Checklist


WizBangCrash:
I also checked that `ca-certificates` is present in the python:3.13-slim-trixie base image

maziggy (changes requested):
Direction is right — env-var-gated opt-in, merging the user's CA into the system bundle instead of replacing it, no application code change. Matches what we agreed on in #1289. A few things to address before merge, plus some nice-to-haves.

== Should fix before merge ==

1. Silent-failure path leaves the user worse off than the disabled case.

In the "update-ca-certificates not found" branch, the script logs "error: ... cannot update trust store" and continues. But because USE_SYSTEM_TRUST_STORE is set, the user explicitly asked Bambuddy to trust their CA. If the binary is missing, SSL_CERT_DIR never gets exported, Python falls back to the default trust store, and cert validation against HA still fails — but the user sees the container start cleanly and assumes the feature is working.

I just verified `update-ca-certificates` ships in python:3.13-slim-trixie (via the ca-certificates package, pre-installed) so in practice this branch never fires today. But the      safe failure shape is `exit 1` here — if the user opted in and we can't deliver, refuse to start so the failure is visible. Same for the "update-ca-certificates failed" warning branch — surface it loudly, don't silently continue.

2. No detection of empty mount.

If the user sets USE_SYSTEM_TRUST_STORE=true but forgets to mount anything at /usr/local/share/ca-certificates (or mounts an empty dir), update-ca-certificates runs successfully — but their CA isn't added. Same UX failure as #1: user thinks it works, TLS      still fails. Could either:

- List the dir before and after update-ca-certificates and warn if the count didn't change.
- Or just `ls -1 /usr/local/share/ca-certificates/*.crt 2>/dev/null` before running and refuse to proceed if no files match.

3. Interaction with the non-root early-exit at line 37.

The new block lands AFTER the `if [ "$(id -u)" -ne 0 ]; then exec "$@"; fi` guard. If a user runs the container with an explicit `user: 1000:1000` in their compose (which the v0.2.4 entrypoint docs still mention for backwards compat), USE_SYSTEM_TRUST_STORE silently does nothing. Either move the new block above that early exec, or document the constraint in the docker-compose.yml comment. I'd lean toward moving it — `update-ca-certificates` doesn't need root for the read+merge, only for writing into /etc/ssl/certs, which is fine because the early-exit path skips that anyway.

Actually scratch the "move it above" idea — update-ca-certificates does need root to write into /etc/ssl/certs. Just document the constraint: "USE_SYSTEM_TRUST_STORE only takes effect when the container starts as root; remove any explicit `user:` directive      if you set this flag."

== Smaller stuff ==

4. Pin ca-certificates explicitly in the Dockerfile apt-install line.

It's in the base image today as a transitive of curl etc., but that's a fragile dependency — a future base-image rebuild could drop it. One-word addition next to the existing apt-get install line makes the contract explicit.

5. update-ca-certificates is noisy.

It prints every cert it finds (often 130+ lines on Debian). `update-ca-certificates 2>&1 | tail -3` or `update-ca-certificates --fresh 2>/dev/null` cuts that to the summary line.

6. Typo in docker-compose.yml: "certifictae" -> "certificate".

7. Compose example uses `/certs` as the host path.

Most users have certs under their compose project (e.g. `./certs/my-ca.crt`). The literal `/certs` works on Linux but reads as unusual — could go either:
 - `./certs:/usr/local/share/ca-certificates` (relative, matches other examples in the same file)
- Keep `/certs` but add a one-line comment that it's a literal absolute path; replace with your own location.

== Tested ==

I haven't reproduced end-to-end against an HA install, but the entrypoint changes match the pattern from #1289 and the env-var/ SSL_CERT_DIR plumbing is the standard Python TLS path. The runtime behaviour (Python httpx + httpcore + urllib3 all honour   SSL_CERT_DIR) is unchanged from your testing.

WizBangCrash:
Updated as per your requests.
I changed the example in the compose file to `/path/to/certs` to make it explicitly clear the user needs to specify their path to the certs directory.

I now check there are certificates in `/usr/local/share/ca-certificates`. I decided not to check that the directory exists as this is always present on the base container and if the user has mounted the volume incorrectly they will either get an error from docker or from my script.

I have tested all the paths:
- incorrect volume mount
- no .crt file
- bad .crt file

I also reduced the verbosity of update-ca-certificates.

maziggy (changes requested):
Round 2 — most of round 1 is addressed cleanly:

- exit 1 on missing/failed update-ca-certificates (was silent)
- exit 1 on empty mount directory (was silent — now fail-loud, good UX)
- ca-certificates pinned explicitly in the Dockerfile
- update-ca-certificates --fresh to keep the log quiet
- "certifictae" typo gone from docker-compose.yml
- /path/to/certs comment makes the path-substitution intent clear

Three things left, ranked.

1. Non-root user interaction still undocumented (round 1 item 3).

The new cert block sits BELOW the `if [ "$(id -u)" -ne 0 ]; then exec "$@"; fi` guard at line 37. If a user pins `user: 1000:1000` in their compose (which works fine for everything else in this entrypoint — chown_if_needed is a no-op when not root), the cert      block is skipped silently. They get no log line either, because the whole `if [ -n "${USE_SYSTEM_TRUST_STORE:-}" ]` branch is never reached. They'll set USE_SYSTEM_TRUST_STORE=true, restart, and wonder why nothing changed.

2. Diagnostic loss on failure (small but worth fixing).

`update-ca-certificates --fresh 2>/dev/null` swallows stderr even in the failure branch. When the command DOES fail (which is exactly when you want to know why), the only log line is "warning: update-ca-certificates failed" with no detail.

3. Tiny comment typo.

`deploy/docker-entrypoint.sh` line in the new block:
 Check if we have any certificates to process. Error is directory is empty → "Error if directory is empty"

WizBangCrash:
Hopefully this is the last round 😀
1. Non-root
I have moved the certs block before the check for non-root and exit with error if non-root. I wanted to keep all the certs block together rather than having a check before and the actual operation after the root check.

2. Diagnostics
I spent a bit of time on this one checking the output of `update-ca-certificates` under error conditions.
It is not the best of commands and can silently ignore certificates if their content is corrupt 🤔
After consideration, I think that the best approach is to use the command `update-ca-certificates --fresh`. This has a terse output and only lists the certificates it cannot process (it doesn't even list the new ones it has successfully processed).
So, if all is well, the user gets:
```
Updating certificates in /etc/ssl/certs...
155 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
```
and, if certificates cannot be processed the output looks like:
```
Updating certificates in /etc/ssl/certs...
rehash: warning: skipping bad.pem, it does not contain exactly one certificate or CRL
rehash: warning: skipping bad2.pem, it does not contain exactly one certificate or CRL
rehash: warning: skipping bad2_copy.pem, it does not contain exactly one certificate or CRL
rehash: warning: skipping ca-certificates.crt, it does not contain exactly one certificate or CRL
155 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
```

3. Typo fixed

maziggy:
Thanks @WizBangCrash, going to close this one — not because of your changes, but because the branch picked up a CHANGELOG  and APP_VERSION rollback from a main → branch merge that predates dev's 0.2.5b1 bump (main is at 0.2.4.1 and was edited after dev moved on, so the three-way merge kept main's older values). The PR diff against dev now shows the version going backwards, which isn't safe to merge as-is.

WizBangCrash:
Do you need me to raise a new PR against `dev` ?

maziggy:
To be clear on why this got closed rather than asked to be reworked: the branch state was the dealbreaker, not the   feature. The merge from main on top of dev pulled APP_VERSION and CHANGELOG.md backwards from 0.2.5b1 to   0.2.4.1 alongside your cert changes. If I'd merged the PR as-is it would have rolled the version back and clobbered the whole 0.2.5b1 changelog section — every unreleased fix on dev would have disappeared from the file. That's not  something I can accept on a merge, regardless of how good the feature itself is.

Thanks for the work.

WizBangCrash:
Okay, so if i understand you correctly, a new PR based on the latest `dev` would be acceptable?

PR #1440, opened 2026-05-19
fix: compute AMS mapping from force-colour overrides when 3MF reqs unavailable

### LINKED ISSUE #1437: [Bug]: filament_overrides with force_color_match ignored at dispatch — AMS slot not selected
### Component

Bambuddy

### Bug Description

When a print queue item has `filament_overrides` set (with `force_color_match: true`) and `ams_mapping` is `null`, bambuddy dispatches the print without translating the overrides into an AMS slot mapping. The printer falls back to default AMS behaviour and uses the wrong filament slot.

#### Actual Behaviour

Bambuddy dispatches the job with `ams_mapping: null`. The printer ignores the `filament_overrides` and picks filament using its default AMS logic, resulting in the wrong colour being used.

#### Queue Item Payload (reproduction data)

```json
{
  "id": 3,
  "printer_id": 5,
  "target_model": "P1S",
  "use_ams": true,
  "ams_mapping": null,
  "filament_overrides": [
    {
      "slot_id": 1,
      "type": "PLA",
      "color": "#CBC6B8",
      "color_name": "PLA Matte",
      "force_color_match": true
    }
  ],
  "manual_start": false,
  "status": "cancelled"
}
```

The `filament_overrides` are stored correctly (confirmed via the PATCH `/queue/{id}` endpoint) and are visible in the queue item response, but they are not actioned at dispatch.

### Expected Behavior

Bambuddy should inspect `filament_overrides` at dispatch time. For each override with `force_color_match: true`, it should match the `slot_id`, `type`, and `color` against the printer's loaded AMS slots and build an `ams_mapping` before sending the print command to the printer.

### Steps to Reproduce

1. Add a print job to the queue via the API with `filament_overrides` containing at least one entry where `force_color_match: true` and `ams_mapping` is omitted (left `null`).
2. Allow bambuddy to auto-dispatch the job when a suitable printer becomes available.
3. Observe the filament slot the printer uses.

### Printer Model

P1S

### Bambuddy Version

0.2.4.1

### SpoolBuddy Version

_No response_

### Printer Firmware Version

01.09.00.00

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

_No response_

### Screenshots

_No response_

#### Suggested Fix

At dispatch time (before sending the print command to the printer), check whether `ams_mapping` is `null` and `filament_overrides` is non-empty. If so:

1. Fetch the current AMS slot state for the assigned printer.
2. For each override with `force_color_match: true`, find the AMS slot whose `type` and `color` best match the override's `type` and `color`.
3. Build and apply an `ams_mapping` from that match before dispatching.

This ensures the `filament_overrides` actually influence which AMS slot the printer draws from, not just which filament metadata is recorded.

### Checklist

Person2099:
## Description

When a print queue item has `force_color_match` filament overrides but no specific plate is targeted (`plate_id=None`), the scheduler was dispatching jobs with `ams_mapping: null`. The printer then picked an AMS slot by filament type only, ignoring the required colour entirely.

Two-layer fix:

1. **`filament_requirements.py`** — `plate_id=None` now walks `<plate>` elements first (modern BambuStudio format). The previous code fell through to `_collect_filaments(root, …)` which uses `./filament` XPath and finds only direct children of `<config>` — returning `[]` for every modern 3MF whose filaments are nested under `<plate>`. Slots shared across multiple plates are deduplicated; the entry with the highest `used_grams` wins.

2. **`print_scheduler.py`** — Added `_build_override_direct_mapping` helper and a fallback in `_compute_ams_mapping_for_printer`. When `_get_filament_requirements` still returns `None` (e.g. 3MF is missing `slice_info.config` entirely) but `force_color_match` overrides are present, the scheduler now builds the AMS mapping directly from the overrides instead of returning `None`.

## Related Issue

Fixes #1437 

## Documentation

**Pick one**:
- [x] No docs update required — reason: internal bug fix, no user-visible config, CLI flags, or API changes

## Type of Change

- [x] Test addition or update

## Changes Made

- `backend/app/services/filament_requirements.py`: Fixed `plate_id=None` path to iterate `<plate>` elements in modern BambuStudio 3MF format, with slot deduplication by highest `used_grams`
- `backend/app/services/print_scheduler.py`: Added `_build_override_direct_mapping` method; added fallback in `_compute_ams_mapping_for_printer` to build AMS mapping from `force_color_match` overrides when 3MF filament requirements are unavailable
- `backend/tests/unit/services/test_filament_requirements.py`: 4 new tests covering `plate_id=None` with modern/flat 3MF formats and deduplication behaviour
- `backend/tests/unit/test_scheduler_force_color_ams_fallback.py`: 11 new tests covering `_build_override_direct_mapping` and the fallback path in `_compute_ams_mapping_for_printer`

## Screenshots

N/A — backend scheduler logic, no UI change.

## Testing

- 107 unit tests pass (22 new, 85 existing — no regressions)
- Ruff lint and format clean

## Checklist

## Additional Notes

The fix is backwards-compatible. The `plate_id=None` flat-filament fallback (older BambuStudio format) is preserved — the new code only takes the `<plate>` path when `<plate>` elements are present. The `_build_override_direct_mapping` fallback only activates when both conditions hold: filament requirements are unavailable **and** at least one override has `force_color_match: true`. Normal dispatch paths are unaffected.


maziggy:
Wish all PR's would have this quality ;) Thanks!!!

PR #1529, opened 2026-05-25
Feature/windows installer

### LINKED ISSUE #1472: [Feature]:Windows Installer
### Problem or Use Case

No native windows installation and easy way to do it. 

### Proposed Solution

Make either a powershelgl script and or a web gui exe 

### Alternatives Considered

_No response_

### Feature Category

Other

### Priority

Would improve my workflow

### Mockups or Examples

_No response_

### Contribution

### Checklist

vmhomelab:
## Description

Added a powershell script to install bambuddy easily via powershell
## Related Issue

Fixes #1472 

## Documentation

- Wiki: maziggy/bambuddy-wiki#36

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Added the powershell to the /install folder
- edited the README.md and /install/README.md accordingly

## Screenshots

Added the script and updated the /install/README.md file
grafik
grafik

Updated the main README.md
grafik

## Testing

- [x] I have tested this on another virtual machine (Windows 11)

## Checklist


maziggy (changes requested):
## Blocker — data lives inside the repo

The script never sets DATA_DIR or LOG_DIR. With them unset, Bambuddy falls back to base_dir = the cloned repo root (see backend/app/core/config.py — _data_dir = _app_dir if no env var). So bambuddy.db, archive/, data/plate_calibration/, and logs/ all land   inside C:\Bambuddy\bambuddy\ (the git working tree).

Two ways that hurts:

1. The "Remove this directory and clone again?" prompt at the broken-repo branch defaults to Yes and does Remove-Item -Recurse -Force on the whole repo dir — which now contains the user's archives and database. Real data loss on a re-run.
2. Any future git pull update is at the mercy of whether runtime files collide with newly-tracked files.

install.sh handles this by setting DATA_DIR / LOG_DIR via the systemd Environment= lines (install.sh:538-539). Mirror that on Windows:

- Create $installDir\data and $installDir\logs as siblings of the repo
- For NSSM:
    & $nssmExe set $ServiceName AppEnvironmentExtra `
        "+DATA_DIR=$dataDir" "+LOG_DIR=$logDir"
- Prepend to Start-Bambuddy.ps1:
    $env:DATA_DIR = "<path>"
    $env:LOG_DIR  = "<path>"

After this fix, user data survives every re-run / re-clone path the script has.

## Important (please address in the same revision)

1. Set-ExecutionPolicy AllSigned -Scope Process -Force (around line 290) — process-scoped, the script is already running, doesn't sign anything, and is at odds with the -ExecutionPolicy Bypass used to invoke the generated Start-Bambuddy.ps1 later. Either remove the line or set it to Bypass with a comment explaining why.

2. NSSM download has no integrity check. The script runs as Administrator and writes nssm.exe into the install dir then registers it as a service. Pin the SHA256 of nssm-2.24.zip and verify with Get-FileHash before extracting — nssm.cc has been      stable since 2014, but the cost of a one-line check is tiny vs. the consequence of a hijacked download.

3. No Python version assertion. Bambuddy requires Python >= 3.10 (pyproject.toml). Get-PythonCommand only checks for "Python 3" via --version; a system with Python 3.9 in PATH proceeds and then pip install fails opaquely. Parse the minor version and reject < 3.10.

4. No unattended mode. install.sh has -y; docker-install.ps1 accepts named params. This script asks 8+ Y/n questions with no skip path. That blocks IT-managed deploys and the maker-space use case people are asking about. Add [switch]$Yes and [switch]$Silent that take the defaults for every Read-YesNo.

5. --host 0.0.0.0 is hardcoded in the generated start script — binds on every interface with no warning. Most LAN users want this, but ask (default Yes for "expose on LAN?", fall back to 127.0.0.1 if No).

6. Read-YesNo accepts j / ja / nein but no other UI string is localized — confusing partial localization. Please add locales for all languages - as per CONTRIBUTING.md.

7. NSSM AppExit Default Restart will respin on any non-zero exit, including signals during Stop-Service. Verify Bambuddy returns 0 on clean shutdown; if not, the service loops on stop. If you can't verify, set AppExit Default Exit and AppExit 0 Exit and only      restart on actual crashes.

## Minor polish

- Relaunch-AsAdmin and Fix-FolderPermissions use unapproved PowerShell verbs — Restart-AsAdmin and Repair-FolderPermissions (or Set-FolderPermissions) are the conformant names. PowerShell warns on import in some host configurations.
- Stop-InstallerLogging is an empty function (body just says "No transcript is used anymore"). Remove.
- Three log destinations (install.log, bambuddy-runtime.log, the repo's own logs/bambuddy.log) — call out which is which in the final summary, or consolidate.
- Set-Content -Encoding UTF8 writes UTF-8 *with BOM* on PS 5.1. The generated Start-Bambuddy.ps1 carries that BOM. Works on 5.1 but is finicky cross-version — docker-install.ps1:223 uses [System.IO.File]::WriteAllText to write without BOM, same trick applies here.
- Default install dir C:\Bambuddy is unusual on Windows; convention is C:\Program Files\Bambuddy or %LOCALAPPDATA%\Bambuddy. Not a blocker, but worth a comment in the wiki PR explaining the choice. 
- git clone --progress could be git clone --depth=1 for the fresh-clone path — Bambuddy's history is multi-GB, shallow clone halves the install time for new users.
- Probe Get-NetTCPConnection -LocalPort $port before accepting the port — otherwise a conflict shows up as an NSSM 5-second-restart loop and the user has no idea why.

vmhomelab:
@maziggy thank you for the feedback. I will review the problems and make changes accordingly.

maziggy (changes requested):
## Round 2 review

Thanks for the rework — the round-1 blocker (data outside the repo) is solidly addressed, and the SHA256-pinned NSSM, Python 3.10 gate, AppExit policy, UTF-8 without BOM, port pre-check, and the -Yes / -Silent / -NoService / -NoStart / -LocalOnly switches are all done well. One real blocker left plus a few smaller items.

## Blocker — Move-LegacyRuntimeData corrupts auth state

The third mapping moves <repo>\data wholesale to $DATA_DIR\data:

      @{ Source = (Join-Path $BambuddyDir "data"); Destination = (Join-Path $DataDir "data") }

But the runtime expects the children directly under DATA_DIR:

- backend/app/core/config.py:24 — plate_calibration_dir = DATA_DIR / "plate_calibration"
- backend/app/core/paths.py:23-26 — .jwt_secret and .mfa_encryption_key live at DATA_DIR / "..."

So after the migration the files land at $DATA_DIR\data\plate_calibration, $DATA_DIR\data\.jwt_secret, $DATA_DIR\data\.mfa_encryption_key. The code keeps looking one level up, so on next start it:

- treats plate calibration as missing — every plate needs re-calibrating
- generates a fresh .jwt_secret — every user is logged out
- generates a fresh .mfa_encryption_key — every user with 2FA enrolled is locked out (their stored MFA secrets become undecryptable)

Only hits users who tested a previous build of this PR and now upgrade, but that is everyone who beta-tested. Suggested shape:

      $legacyDataRoot = Join-Path $BambuddyDir "data"
      if (Test-Path $legacyDataRoot) {
          Get-ChildItem -Path $legacyDataRoot -Force | ForEach-Object {
              $dest = Join-Path $DataDir $_.Name
              if (-not (Test-Path $dest)) {
                  Write-Log "Migrating $($_.FullName) -> $dest" "INFO" Cyan
                  Move-Item -Path $_.FullName -Destination $dest -Force
              }
              else {
                  Write-Log "Keeping $($_.FullName); $dest already exists." "WARN" Yellow
              }
          }
          if (-not (Get-ChildItem -Path $legacyDataRoot -Force)) {
              Remove-Item $legacyDataRoot -Force
          }
      }

## Important

1. Elevated session swallows errors. Restart-AsAdmin spawns a new elevated PowerShell with Start-Process -Verb RunAs and no -NoExit. The catch block at line 1036 writes    the error and exit 1, and Windows closes the elevated window before the user can read it. Either pass -NoExit in $arguments, or end the catch with Read-Host "Press Enter to close" gated on Test-IsAdmin so the non-elevated parent does not double-pause.

2. Set-FolderPermissions runs four times on overlapping paths (lines 678, 760, 761, 762). icacls /T /C is recursive, and once $bambuddyDir contains venv\ that is ~50k files per pass. Drop line 760 and let inheritance from the line-678 pass cover the data\ and logs\ siblings created right after.

3. Half-localization replaced with English-only. Round 1 flagged de/j/ja/nein in Read-YesNo as confusing partial localization and asked for full locales per   CONTRIBUTING.md. This round removes the German answers, making the script consistent English. Defensible (the installer is not part of the React i18n stack), but does   not strictly match the round-1 ask — confirm English-only is acceptable.

4. Broken-repo prompt (line 789) defaults to No now, which is safer, but if the user says Yes and the broken dir holds a DB or .jwt_secret, Remove-Item nukes it. Call Move-LegacyRuntimeData on $bambuddyDir before Remove-Item runs.

## Minor

- Line 605: AppExit 0 Exit is covered by AppExit Default Exit on line 604. Drop it.
- Service args (line 579) embed quotes in a single string. Works for C:\Bambuddy; worth verifying with nssm get Bambuddy AppParameters after an install into a path with    spaces (e.g. C:\Program Files\Bambuddy).
- Default C:\Bambuddy still does not match Windows convention (C:\Program Files\Bambuddy or %LOCALAPPDATA%\Bambuddy). Wiki PR could note the choice.
- catch block at line 1036 could log $_.ScriptStackTrace into install.log — without it, post-mortem from a user-submitted install.log is painful.

maziggy (changes requested):
Round 3 review

Two Important items left plus a few smaller things; aiming to wrap this in the same revision.

Important — NSSM restart policy diverges from systemd Restart=on-failure

Current AppExit policy (lines 625-628):

AppExit Default Exit
AppExit 0 Exit
AppExit 1 Restart

Only exit code 1 triggers a restart. Anything else — segfaults (0xC0000005), Windows OOM kill, Python sys.exit(2), unhandled signals — stops the service with no retry.
The Linux installer at install/install.sh:543 uses Restart=on-failure, which restarts on any non-zero exit code and on signals/watchdog timeouts. Same Bambuddy, very different production resilience.

Round 1's "if you can't verify, set Default Exit + 0 Exit + only restart on real crashes" was conservative guidance; "real crashes" really means "any non-zero exit",  not "exit code 1 only". Recommended:

AppExit Default Restart
AppExit 0 Exit
AppRestartDelay 5000

That matches Linux systemd behaviour and removes the round-2 minor 1 redundancy (AppExit 0 Exit previously duplicated AppExit Default Exit) in the same edit. Bambuddy   returns 0 on Stop-Service / SIGTERM (uvicorn handles both cleanly), so the safety condition is met.

Important — LAN-expose prompt should mention auth

Line 715:

"Expose Bambuddy on the LAN? Choose No to bind only to this computer. Exposing Bambuddy on the LAN binds to all network interfaces."

Bambuddy ships with auth disabled by default; binding to 0.0.0.0 exposes an unauthenticated UI on every interface — including any VPN tunnels (Tailscale etc.) that   happen to be up. Worth one extra line in the prompt:

"Bambuddy is unauthenticated by default. After installation, open Settings -> Security and enable auth before relying on LAN access."

Cheap nudge; addresses the security posture this project keeps having to talk about.

Minor

1. Round-2 minor 4 (log $_.ScriptStackTrace to install.log) not yet in. Catch block at line 1063 still only logs $_.Exception.Message. One extra Add-Content line covers it:

Add-Content -Path $script:LogFile -Value "[ERROR] $($_.Exception.Message)`n$($_.ScriptStackTrace)" -Encoding UTF8

1. Without it, post-mortem from a user-submitted install.log still tells you what failed but not where.
2. Round-2 minor 2 (NSSM service args with spaces-in-path). Did you run nssm get Bambuddy AppParameters after installing to a path with spaces (e.g. C:\Program
  Files\Bambuddy)? The current $serviceArguments string at line 600 should preserve the inner quotes around the start-script path, but worth confirming once before merge.
3. Restart-AsAdmin elevation-cancelled branch (lines 158-161): if the user dismisses UAC, you Write-Host then exit 1 from a window that may close immediately
  (double-clicked .ps1, IE prompt, etc.). Same Read-Host pause you added to the main catch block at line 1070 belongs here too:

Read-Host "Press Enter to close"
exit 1
4. Show-BambuddyBanner runs even in -Silent mode (line 660 → unconditional). Trivial; gate it on -not $script:Silent.

maziggy (changes requested):
Round 4(!!!) review

All six round-3 items landed correctly. The NSSM restart policy now matches systemd Restart=on-failure, the LAN-bind prompt carries the auth caveat, stack traces hit   install.log, nssm get AppParameters verifies the args after install, and the elevation-cancelled + banner paths respect -Silent cleanly. The legacy data migration also   lines up with where the backend reads its files: $DATA_DIR/plate_calibration, $DATA_DIR/.jwt_secret, $DATA_DIR/.mfa_encryption_key, $DATA_DIR/archive — all verified against backend/app/core/config.py and paths.py.

One important and a few minors, then we should be done.

Important — Firewall rule covers Public profile

Line 945:

New-NetFirewallRule `
    -DisplayName $ruleName `
    -Direction Inbound `
    -Protocol TCP `
    -LocalPort $port `
    -Action Allow | Out-Null

No -Profile argument, so the rule defaults to Any — Domain + Private + Public. The user agrees to "expose on the LAN" assuming "LAN" means their home network, but the   rule also fires when they join a cafe / hotel / airport network where Windows tags the connection as Public. Combined with auth-off-by-default, that's the same exposure the round-3 prompt is meant to warn about, made more reachable.

Users on Public networks can still reach Bambuddy from localhost (uvicorn's bind doesn't care about firewall profile), and travellers don't have an unauthenticated UI   advertised on whichever shared SSID they joined.

Minor

1. Legacy icons/ directory not migrated. backend/app/api/routes/external_links.py:26 resolves user-uploaded external-link icons to base_dir / "icons". After DATA_DIR moves base_dir to $installDir\data, any custom icons previously uploaded via the External Links admin feature get orphaned at <repo>\icons. Admin-only edge case so not a blocker, but a one-line addition to the $legacyMappings array on line 409 covers it:

@{ Source = (Join-Path $BambuddyDir "icons"); Destination = (Join-Path $DataDir "icons") }
2. Elevated success window closes immediately when no follow-up runs. If the user picks Yes to "Register service" and No to "Start service now", the script returns
  from Register-BambuddyService, prints the summary at lines 1022-1043, then exits — the elevated PowerShell window closes and the summary is gone. Mirror the
  catch-block pattern at line 1079-1082 right before the final closing brace of the try block:

  if ((Test-IsAdmin) -and (-not $script:Yes) -and (-not $script:Silent)) {
      Read-Host "Press Enter to close"
  }
  3. No Windows-side update.sh equivalent. install.sh has a sibling update.sh that handles service stop/start, backup-before-update, and code rollback on failure. A user
  who later does git pull themselves in C:\Bambuddy\bambuddy won't get any of that. Out of scope for this PR — just noting it as the next thing in the same area. Worth
  a wiki paragraph at minimum.

maziggy (approved):

maziggy:
Thanks!

PR #1541, opened 2026-05-27
fix(pwa): reliable Android Chrome install prompt + CSP hardening

Soopahfly:
## Summary

- **Captures `beforeinstallprompt` before React mounts** (`pwa.ts` imported as the first side-effect in `main.tsx`) so the event is never dropped during the auth-loading window — previously the component-level `useEffect` listener missed the event if it fired before the auth API call resolved
- **Always shows the Install button on Android Chrome** — native prompt flow when the event fires, manual toast fallback ("Tap ⋮ → Install app") otherwise, so the button is useful even when Chrome is in a cooldown period
- **Suppresses the mini-infobar on desktop** (≥ 1144 px) via `e.preventDefault()` so the sidebar Download button is the sole install trigger; mobile lets Chrome's native UI appear naturally
- **Adds explicit `worker-src 'self'`** to the CSP in `security_headers_middleware` — without this Chrome 128+ may treat the fallback chain as ambiguous and refuse to count the service worker toward PWA installability
- **No-cache headers** on `/manifest.json`, `/sw.js`, and `/sw-register.js` so stale manifests never block an install eligibility re-check; `/manifest.json` and `/sw-register.js` also gain HEAD method support so health scanners don't return 405
- **Adds nonce to gcode-viewer `script-src`** to match the policy applied to the main SPA

## Test plan

- [ ] Desktop (≥ 1144 px): sidebar Download button appears and triggers Chrome's native install dialog; browser mini-infobar does **not** appear
- [ ] Android Chrome (fresh profile): `beforeinstallprompt` fires, Install button triggers native dialog
- [ ] Android Chrome (prompt unavailable / cooldown): Install button still visible in hamburger menu; tap shows toast with manual instructions
- [ ] Already installed (standalone mode): Install button hidden entirely
- [ ] `/manifest.json`, `/sw.js`, `/sw-register.js` return `Cache-Control: no-cache, no-store, must-revalidate`
- [ ] CSP response header includes `worker-src 'self'`
- [ ] `curl -I /manifest.json` returns 200 (not 405)

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -5653,7 +5653,11 @@ async def serve_manifest():
     """Serve PWA manifest."""
     manifest_file = app_settings.static_dir / "manifest.json"
     if manifest_file.exists():
-        return FileResponse(manifest_file, media_type="application/manifest+json")
+        return FileResponse(
+            manifest_file,
+            media_type="application/manifest+json",
+            headers={"Cache-Control": "no-cache, no-store, must-revalidate"},
+        )
     return {"error": "Manifest not found"}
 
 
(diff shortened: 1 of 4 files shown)


Soopahfly:
Opened in wrong repo — submitting to my fork only for now.

PR #1625, opened 2026-06-03
feat: Unify print dispatch through the scheduler

### LINKED ISSUE #1354: [Feature]: Unify Queue UI and behavior
### Problem or Use Case

The queue function has three issues: terminology inconsistency, modal title confusion, and a hidden behavioral fork depending on origin.

### 1. Terminology chaos across entry points

The same underlying action is labelled differently depending on where you are:

| Location | Button/Menu label | Locale key | What actually happens |
|---|---|---|---|
| Archives context menu | **"Print"** | `archives.menu.print` | Opens PrintModal `mode='reprint'` → immediate dispatch |
| Archives context menu | **"Schedule"** | `archives.menu.schedule` | Opens PrintModal `mode='add-to-queue'` |
| Archive detail toolbar | **"Reprint"** | `archives.detail.reprint` | Same as above — `mode='reprint'` |
| Archive detail toolbar | **"Schedule Print"** | `archives.detail.schedulePrint` | `mode='add-to-queue'` |
| File manager toolbar | **"Print Now"** | `fileManager.print` | `mode='reprint'` |
| File manager toolbar | **"Add to Queue"** | `fileManager.addToQueue` | `mode='add-to-queue'` |
| Queue page | **"Re-print"** | `queue.reprint` | `mode='reprint'` |
| Queue page | **"Add to Queue"** | `queue.addToQueue` | `mode='add-to-queue'` |

The same "send to printer immediately" action is called: **Print**, **Reprint**, **Re-print**, and **Print Now** — four different labels for the same thing, across three pages.

The same "queue for later" action is called: **Schedule**, **Schedule Print**, and **Add to Queue**.

---

### 2. Modal title is wrong for the library file path

([PrintModal/index.tsx:867](frontend/src/components/PrintModal/index.tsx#L867)):

```ts
title: isLibraryFile ? t('queue.print') : t('queue.reprint'),
```

- **From archives** → modal title = **"Re-print"** (makes sense, it's a known past print)
- **From file manager** → modal title = **"Print"** (fine, but the modal body still shows archive/reprint UI, and the submit button says `queue.print` = **"Print"**, not "Print Now")

Meanwhile the button that opened it in the file manager said **"Print Now"**, so users see "Print Now" → click → modal says **"Print"**. The word "Now" disappears, which subtly implies it might not be immediate.

---

### 3. "Schedule" vs "Add to Queue" split is invisible to users

The empty-queue state hint ([en.ts:993](frontend/src/i18n/locales/en.ts#L993)) says:

> *"Schedule a print from the Archives page using the 'Schedule' option in the context menu"*

But on the file manager, the equivalent action is called **"Add to Queue"** — not "Schedule". A user reading that hint and going to the file manager will look for a "Schedule" button and won't find one.

---

### 4. The `add-to-queue` modal title is "Schedule Print" regardless of whether the user is scheduling or just queuing ASAP

In [PrintModal/index.tsx:887](frontend/src/components/PrintModal/index.tsx#L887):

```ts
title: t('queue.schedulePrint'),  // always "Schedule Print"
```

But inside that modal, the user can choose **"ASAP"** timing, which is not scheduling, it's just queuing. The title "Schedule Print" implies a date/time is being set, which is only true for one of the three timing options. Users choosing ASAP are confused by the title.

---

### 5. `_run_reprint_archive` does not call `archive_print`

It just uploads the file and starts the print. A new row is never inserted.
- On print start, `on_print_start` in [main.py:1771-1785](backend/app/main.py#L1771) finds the existing archive via `_expected_prints` and **overwrites `status` and `started_at`** in-place. On completion, `update_archive_status` in [archive.py:1163](backend/app/services/archive.py#L1163) overwrites `status`, `completed_at`, and `failure_reason`.
- So reprinting overwrites the last run's outcome (success → failed → cancelled → row shows "cancelled").

---

### 6. Gating gaps

**Plate-clear gate** 

The scheduler reads `require_plate_clear` and calls `printer_manager.is_awaiting_plate_clear()` at [print_scheduler.py:145](backend/app/services/print_scheduler.py#L145) and [print_scheduler.py:1157](backend/app/services/print_scheduler.py#L1157).

The reprint path goes: `POST /archives/{id}/reprint` → [archives.py:3547](backend/app/api/routes/archives.py#L3547) (only checks `is_connected`) → `background_dispatch._dispatch()` → `_run_reprint_archive`. Neither [archives.py:3505–3581](backend/app/api/routes/archives.py#L3505) nor [background_dispatch._dispatch():270–278](backend/app/services/background_dispatch.py#L270) contain any plate-clear check. The flag is never read.

**Drying hold**

The scheduler's drying hold lives entirely in `PrintScheduler._drying_in_progress` (an instance dict) and is checked at [print_scheduler.py:245–258](backend/app/services/print_scheduler.py#L245). `background_dispatch.py` has no reference to `_drying_in_progress`, `queue_drying_block`, or the scheduler instance at all. A reprint command reaches `_run_reprint_archive` with no knowledge that the printer is mid-drying cycle.

**Post-dispatch hold window**

`PrintScheduler._dispatch_holds` and `_printer_in_dispatch_hold()` live at [print_scheduler.py:86](backend/app/services/print_scheduler.py#L86) and [print_scheduler.py:1099](backend/app/services/print_scheduler.py#L1099). `BackgroundDispatchService._dispatch()` at [background_dispatch.py:256](backend/app/services/background_dispatch.py#L256) has its own "already dispatching" guard (lines 271–278) but that only blocks if a background dispatch job is actively queued or running, it does not enforce the 60–180 s post-dispatch hold the scheduler uses for the H2D's firmware digestion window. The two services have no shared state.

**Filament validation / AMS recompute** 

`_get_missing_filament_types`, `_get_missing_force_color_slots`, and `_compute_ams_mapping_for_printer` all live in `print_scheduler.py` ([line 688](backend/app/services/print_scheduler.py#L688), [line 650](backend/app/services/print_scheduler.py#L650), [line 761](backend/app/services/print_scheduler.py#L761)). `_run_reprint_archive` at [background_dispatch.py:547](backend/app/services/background_dispatch.py#L547) reads `job.options.get("ams_mapping")` and passes it to `start_print` verbatim at line 658 without validation or recompute.

**G-code injection** 

Injection runs at [print_scheduler.py:1713–1737](backend/app/services/print_scheduler.py#L1713) inside `_start_print`, gated on `item.gcode_injection`. `_run_reprint_archive` has no equivalent block; it uploads `archive.file_path` as-is with no opportunity for snippet injection even if the user wanted it.

### Proposed Solution

The simplest fix would be to pick two canonical terms app-wide ("Print" for immediate and "Add to Queue" for queued) and use them consistently everywhere, including the modal title (which should become "Add to Queue" rather than "Schedule Print" when the user isn't explicitly scheduling).

Also, route every print — including "Print Now" — through `check_queue` in `print_scheduler.py`, so all five gating checks apply unconditionally. You could add a `priority` column to `PrintQueueItem`, a scheduler wake event, and make `reprint_archive` and `dispatch_print_library_file` create a queue item instead of calling `_dispatch()`. Then `reprint_archive` HTTP endpoint becomes a thin wrapper, `_run_reprint_archive` and `_run_print_library_file` in `background_dispatch.py` can be removed. 

### Alternatives Considered

_No response_

### Feature Category

Print Queue & Scheduling

### Priority

Nice to have

### Mockups or Examples

_No response_

### Contribution

### Checklist

EdwardChamberlain:
## Description

Unifies frontend print initiation so archive, file manager, queue, and printer-card print actions all use the same Print modal and dispatch workflow through the scheduler-backed queue. Legacy direct-print API routes remain available but now emit deprecation logs.

## Related Issue

Fixes #1354

## Documentation

**Pick one**:
- [X] Docs PR(s): 
- maziggy/bambuddy-website#2
- maziggy/bambuddy-wiki#46
- [ ] No docs update required

## Type of Change

- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [x] Test addition or update

## Changes Made

- Consolidated print UI actions to a single `Print` label and shared `PrintModal` flow.
- Routed print initiation through `POST /api/v1/queue/` so ASAP, queued, and scheduled jobs all pass through the scheduler.
- Added scheduler-aware dispatch options for ASAP, Queue, Schedule, and manual-start queueing.
- Added deprecation logging for legacy direct-print API calls.
- Consolidated translation keys and updated locale parity, including newly added upstream locales.
- Added and updated tests for the unified print modal and queue insertion behaviour.

## Screenshots

image

## Testing

- [x] I have tested with my printer model: P1S, H2S

Validated with:

- `npm run check:i18n`
- `npm test -- PrintModal.test.tsx`
- `npx tsc -b`
- `git diff --check`
- Pre-commit hooks during commit amend
- ./test_frontend.sh: Pass
- ./test_backend.sh: Pass

## Checklist

## Additional Notes

ASAP jobs are inserted at the top of the queue rather than bypassing it. Queue jobs are appended to the back unless manual start is required. Scheduled jobs use the scheduler as a “do not start before” time: if the printer is busy at the scheduled time, the job remains queued until it can be dispatched.

Relatively pivotal change to the despatch of jobs so appreciate this may need some discussion / back and forth.


Keybored02:
Are you sure about the no docs? I would at least update the screenshots, to reflect the change

EdwardChamberlain:
@Keybored02 

Yes, probably a good idea. Will look into a docs PR tonight to update the screenshots and probably the general text around despatch routes etc

EdwardChamberlain:
First pass at the docs is now available.

Conflicts have been resolved. This touches a fairly broad set of files, so there’s a reasonable chance it will pick up further conflicts if it remains open for long. An initial review would be useful when possible.

maziggy:

EdwardChamberlain:
Rebased and conflicts resolved.

maziggy (changes requested):
**Blockers:**

1. Legacy backend routes still dispatch directly. POST /archives/{id}/reprint (archives.py:4045) and POST /library/files/{id}/print (library.py:3986) currently only
  emit a logger.warning and still call background_dispatch._dispatch() -> _run_reprint_archive / _run_print_library_file. Those are the exact bypass paths #1354 calls out. Please either return 410 Gone, or have them internally create a queue item and return it, so there's no remaining route that skips the gates. The frontend fix is the easy half; the API surface is where future regressions hide.

2. No backend test for cleanup_library_after_dispatch. This is the riskiest new backend logic since it unlinks files from disk. Please add a backend/tests/unit/test_scheduler_cleanup_library.py covering: (a) library file unlinked + DB row removed when flag is true, (b) is_external=True skips cleanup entirely, (c) archive-creation failure skips cleanup (no unlinks queued), (d) thumbnail path resolution for both absolute and relative library_file.thumbnail_path, (e) the archive's own copy survives the unlink (the safety invariant the whole feature rests on), (f) OSError during unlink does not crash dispatch.

3. Permission scope change needs explicit handling. FileManager "Print" goes from printers:control to queue:create; PrintersPage upload+print to library:upload AND   queue:create; Archives buttons to queue:create AND archives:reprint. Installations that explicitly granted printers:control without queue:create lose those actions, and   queue:create now implicitly enables immediate dispatch via ASAP. Please add a  CHANGELOG entry calling this out, and double-check whether any docs page lists the   print-related permissions so they stay consistent.

4. Insert-position coverage is thin. Please extend test_print_queue_api.py with: (a) ASAP with quantity>1 shifts existing items by quantity, not 1; (b) unassigned   (printer_id=None) scope insertion; (c) insert_position greater than max position appends cleanly (no gaps, no shift); (d) two consecutive ASAP inserts to the same printer stack in the order they were submitted (the asapInsertionCounts map in PrintModal handles this on the client; verify the server path agrees).

5. Dispatch toast semantics shifted silently. The old "Print Now" path showed a WebSocket dispatch-event toast confirming the print actually started; the new ASAP path shows "Print queued" and trusts the next scheduler tick. If the printer is offline, drying, or under a plate-clear hold, the user gets no further feedback. Please either tighten the toast to "Will start when printer is idle" when the target is not currently dispatchable, or wire the dispatch-event toast through for the first dispatch of an ASAP queue item.

6. Cleanup unlink runs after db.commit(). In print_scheduler.py:2244-2253, the commit (which removes the library row) precedes the disk unlink loop. An OSError only logs a warning, leaving an orphan file with no DB row. Not catastrophic but worth a small follow-up: either move the unlink before the commit and rollback on failure, or log the orphan path in a way a periodic sweeper can pick up.

**Minor:**

7. frontend/src/components/PrintModal/types.ts now describes both 'reprint' and 'add-to-queue' as "Legacy alias for creating a print queue item". If both modes are aliases, collapse the union to a single non-edit mode (e.g. 'create' | 'edit-queue-item') and migrate the call sites. Keeping two aliases that mean the same thing reintroduces exactly the kind of duplication this PR set out to remove.

8. The ASAP test asserts insert_position: 1 and manual_start: false — please also assert that scheduled_time is omitted, so a future refactor that accidentally stamps 
  "now" can't pass silently.

Keybored02:
@EdwardChamberlain are you still on this? Do you need a hand?

EdwardChamberlain:
Have got most of them resolved. Just need to tidy up the last points and then rebase. Will hopefully be today or tomorrow but would appreciate some help reviewing when I get to that point.

maziggy:
Just let me know when you're done.

EdwardChamberlain:
Rebase is complete. 

One related thought to the permissions: it may be worth considering a separate permission specifically for ASAP queueing/immediate dispatch. At the moment, queue:create allows both adding jobs to the queue and jumping the queue via ASAP. In some deployments, it might be useful to let a user submit jobs to the queue without also granting them the ability to bypass the scheduling/order priority.

In a similar vein - it might be interesting to invert manual start to "automatic start" and introduce permissions for "automatic start" which would be interesting for similar deployments as it would allow auditing of jobs in the queue. Both ideas obviously out of scope here 

Some other concerns:
- Dispatch still treats start_print() as success before printer state changes. This is out of scope here but probably should be visited as a follow up PR as it's a serious robustness issue. The unified despatch method does make addressing this a lot easier though.

maziggy:
(feat-unifi-print-despatch)$ npm run build

> frontend@0.0.0 build
> tsc -b && vite build

src/components/PrintModal/index.tsx:2:38 - error TS6133: 'Calendar' is declared but its value is never read.

2 import { AlertCircle, AlertTriangle, Calendar, Code, Layers, Loader2, Pencil, Printer, X } from 'lucide-react';
                                       ~~~~~~~~

src/components/PrintModal/index.tsx:2:48 - error TS6133: 'Code' is declared but its value is never read.

2 import { AlertCircle, AlertTriangle, Calendar, Code, Layers, Loader2, Pencil, Printer, X } from 'lucide-react';
                                                 ~~~~

src/components/PrintModal/index.tsx:2:54 - error TS6133: 'Layers' is declared but its value is never read.

2 import { AlertCircle, AlertTriangle, Calendar, Code, Layers, Loader2, Pencil, Printer, X } from 'lucide-react';
                                                       ~~~~~~


Found 3 errors.

maziggy:
Lemme check. Would like to merge.

maziggy (approved):

PR #1656, opened 2026-06-05
feat(gcode-injection): per-VP opt-in auto-print injection toggle (#1516)

### LINKED ISSUE #1516: [Feature]: Enable gcode auto inject on new item sent to queue or virtual printer
### Problem or Use Case

I'm using Bambuddy for A1 printer automation with swapmod/jobox plate changers

I added the needed gcode to Settings > Workflow

Image

I'm then sending print via virtual printers
which appear correctly in the queue

Image

But then, for the plate switch automation via the gcode, i have to go to the print item manually to enable the gcode

Image

### Proposed Solution

The Settings > Workflow and toggle have been discussed here https://github.com/maziggy/bambuddy/issues/422#issuecomment-4067787532

For proper automation, a bulk edit to add gcode would be needed (currently can't enable gcode via bulk edit)

Image

Maybe an even better solution would be to be able to auto add the gcode when prints are sent to a specific virtual printer. (as only the A1 in my inventory are plate swap compatible)

So maybe a gcode workflow specific to a virtual printer eg here:
Image

Another issue mentions capabilities https://github.com/maziggy/bambuddy/issues/1306 which may be helpful to tag printers that have the plate changer hardware.

### Alternatives Considered

_No response_

### Feature Category

Print Queue & Scheduling

### Priority

Would improve my workflow

### Mockups or Examples

_No response_

### Contribution

### Checklist

phieb:
## Description

Adds a per-virtual-printer opt-in toggle for auto-print G-code injection. A Print Queue mode VP can now have every Bambu Studio "Send" / upload automatically injected with the per-model start/end snippets — without ticking each queue item by hand. Off by default; a no-op when no snippets are configured for the target model, so existing setups don't change.

Bundled are the fixes that make injection actually work end-to-end (same area, validated together on a P1S):
- **P1S end-snippet** now lands correctly — inserted before `; EXECUTABLE_BLOCK_END`
  and the `.gcode.md5` sidecar is recomputed, otherwise the P1S rejects the file
  (HMS 0500-4003) or silently ignores the snippet.
- **Reprint + quantity + injection** now queues *all* copies, so each is injected
  by the scheduler. Previously the first copy dispatched immediately without
  injection and stayed stuck on the plate for auto-eject farms.
- Resolved (placeholder-substituted) snippet logged at DEBUG for troubleshooting.

## Related Issue

Fixes #1516

## Documentation

**Companion docs PRs**:
- Wiki: https://github.com/maziggy/bambuddy-wiki/pull/40

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

## Changes Made

- Per-VP `gcode_injection` toggle (DB column default false, VP card toggle in queue
  mode, wired through the manager + i18n in all locales).
- P1S end-snippet anchored before `; EXECUTABLE_BLOCK_END` + `.gcode.md5` recompute
  (flagged `usedforsecurity=False` so it isn't a security hash).
- Reprint dispatch: with injection on and quantity > 1, queue all copies.
- DEBUG logging of the resolved start/end snippet.
- Regression tests for the reprint dispatch behaviour.

## Screenshots

virtual-printer-gcode

## Testing

Live-tested on a P1S: injection + placeholder substitution confirmed
(`{max_layer_z}` → resolved Z), the P1S accepts the md5-recomputed file, and the
end snippet executes. Full suite green (frontend 2047, backend 5591, ruff, bandit).


phieb:
@maziggy can you let the tests run? Locally everything was green.

phieb:
@maziggy again pls

Issue #1291, opened 2026-05-12
[Bug]: When using the Internal Spool Inventory Storage Location is not saved nor is it displayed in the table


needo37:
### Component

Bambuddy

### Bug Description

In the internal filament inventory, the `storage_location` field is not exposed by the API and cannot be set or retrieved. The field exists in the database but is absent from the spool schemas (`SpoolBase`,`SpoolUpdate`, `SpoolResponse`), so writes are silently ignored and reads always omit it. The inventory table also always shows `-` in the Storage Location. This all works properly if using the Spoolman inventory backend.

### Expected Behavior

Users should be able to set a free-text storage location (e.g., "Drybox #1", "Top shelf") on any internal spool via the UI or API (`PATCH /api/v1/inventory/spools/{id}`), and the inventory table should display the set storage_location in the Storage Location column.

One could argue that the Storage Location and Location are redundant. But I am assuming there was a technical reason to have two location fields. So I am honoring the design choice.

### Steps to Reproduce

1. Go to a Filament in the Filament Tab.
2. Click on any Filament.
3. Scroll down to Storage Location
4. Type any thing in Storage Location
5. Hit Save.
6. Storage Location does not show up in the table nor if you go back into the filament.

### Printer Model

Not printer-related

### Bambuddy Version

0.2.4

### Installation Method

Docker

### Operating System

Docker

### Relevant Logs / Support Package

[bambuddy-support-20260512-081851.zip](https://github.com/user-attachments/files/27639270/bambuddy-support-20260512-081851.zip)

### Screenshots

Fresh table

<img width="1264" height="270" alt="Image" src="https://github.com/user-attachments/assets/37c8a9ab-81e4-4e7f-b703-615da78703f3" />

Edit the filament to add 1 to set 1 as the Storage Location

<img width="527" height="1496" alt="Image" src="https://github.com/user-attachments/assets/65302a7b-f909-481d-846c-441aac5d2944" />

Storage Location in the table is not updated.

<img width="1257" height="280" alt="Image" src="https://github.com/user-attachments/assets/e3ea4d7f-cf1d-4058-9471-0b79f363d509" />

Storage Location in the filament is missing.

<img width="531" height="1494" alt="Image" src="https://github.com/user-attachments/assets/8713dd9c-ef19-4bc1-954a-9544a164f831" />

### Additional Context

I am willing to fix and submit a PR for this bug.

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
[...]

Issue #41, opened 2025-12-31
[Bug]: Slice / Open in Slicer button - bambustudioopen:// unknown handler Win11


2jgit:
### Bug Description

After uploading a 3MF and attempting to launch via Slice or Open in Slicer, app is calling bambustudioopen:// . Adjusting the call to bambustudio:// launches correctly with Win11

Same result with Firefox/Chrome/Edge

I believe bambustudioopen:// is Apple specific

### Expected Behavior

platform.system() (or similar) should detect OS and adjust handler accordingly 

### Steps to Reproduce

From a Win11 browser attempt to slice an uploaded 3MF file from general archive view or Open in Slicer from the individual 3D preview. 

### Printer Model

Not printer-related

### Bambuddy Version

0.1.6b3

### Printer Firmware Version

NA

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

### Relevant Logs

```shell
Chrome: 
Failed to launch 'bambustudioopen://https%3A%2F%2Fbambu.techknowcs.com%2Fapi%2Fv1%2Farchives%2F1%2Ffile%2FCase%2520minisforum%2520MS-A2-01%2520Noctua%2520120mm.3mf' because the scheme does not have a registered handler.

Firefox: 
Prevented navigation to “bambustudioopen://https%3A%2F%2Fbambu.techknowcs.com%2Fapi%2Fv1%2Farchives%2F1%2Ffile%2FCase%2520minisforum%2520MS-A2-01%2520Noctua%2520120mm.3mf” due to an unknown protocol.

Edge:
Failed to launch 'bambustudioopen://https%3A%2F%2Fbambu.techknowcs.com%2Fapi%2Fv1%2Farchives%2F1%2Ffile%2FCase%2520minisforum%2520MS-A2-01%2520Noctua%2520120mm.3mf' because the scheme does not have a registered handler.
```

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I am using the latest version of Bambuddy
- [x] My printer is set to LAN Only mode

maziggy:
Thanks for pointing this out! Fixed in branch 0.1.6b4.

PR #1659, opened 2026-06-05
feat(inventory): CSV import/export for the inventory page (#1576)

### LINKED ISSUE #1576: [Feature]: Native CSV import/export for the Inventory page
### Problem or Use Case

Adding spools in bulk today means clicking through the spool form one entry at a
time. When you receive a batch order (e.g. 20 Polymaker spools), this is slow and
error-prone. There's also no way to bring an existing inventory (a spreadsheet or
another tool) into Bambuddy in one step, nor to get the current inventory back out
for backup or migrating between Bambuddy instances.

A community script already exists for this over the REST API
(github.com/bsaunder/bambuddy_spoolimport), which shows there's demand — but it
lives outside the app, needs Python + a config file, and isn't accessible to most
users.

### Proposed Solution

Add CSV import and export to the Inventory page.

Import — upload a CSV and get a preview before anything is written:
- Backend parses + validates + resolves colors and returns a per-row preview
  (valid / error-with-reason / skipped). Nothing is written yet.
- Color resolution: if `rgba` is present in the CSV it wins; otherwise
  `brand` + `color_name` are matched against the existing Color Catalog and the
  hex / extra_colors / effect_type are filled in automatically; otherwise blank.
- User reviews the preview table, then confirms — only valid rows are created in
  a single atomic transaction. Invalid rows are skipped; the user fixes the CSV
  and re-uploads.

Export — one click downloads the current inventory as CSV using the same column
schema, so import/export round-trips cleanly (also handy for backups and
Bambuddy → Bambuddy migration).

CSV schema (fixed header, case/space tolerant): material (required), brand,
subtype, color_name, rgba, extra_colors, effect_type, label_weight, cost_per_kg,
nozzle_temp_min, nozzle_temp_max, note.

Proposed API:
- POST /api/v1/inventory/spools/import?dry_run=true → preview JSON (no write)
- POST /api/v1/inventory/spools/import → create valid rows, atomic, returns summary
- GET  /api/v1/inventory/spools/export → text/csv of current inventory

New strings would be added to all 9 locales (i18n parity), with a companion
bambuddy-wiki PR documenting the feature.

### Alternatives Considered

A future enhancement could add a column-mapping UI: upload any CSV / spreadsheet /
Spoolman export and map each source column to a Bambuddy field via dropdowns,
instead of requiring the fixed header. I kept it out of this proposal to stay
focused on a small first version, but happy to follow up with it if you'd prefer
that direction.

### Feature Category

Spool Inventory

### Priority

Would improve my workflow

### Mockups or Examples

[mockup-import.html](https://github.com/user-attachments/files/28428314/mockup-import.html)
Image

### Contribution

### Checklist

samedyuksel:
## Description

Adds native CSV import/export to the Inventory page so spools can be added in
bulk and the inventory can be backed up / migrated. Local inventory only —
the buttons are hidden in Spoolman mode (Spoolman has its own data store).

## Related Issue

Fixes #1576

## Documentation

**Companion docs PRs**:
https://github.com/maziggy/bambuddy-wiki/pull/41

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

## Changes Made

- Backend: new `spool_csv` service (parse/validate/serialize) + two routes
  (`GET /inventory/spools/export`, `POST /inventory/spools/import?dry_run=`).
  Import preview classifies each row valid/error/skipped and writes nothing;
  confirm persists only valid rows in one transaction.
- Colour resolution: explicit `rgba` wins, else `brand` + `color_name` resolve
  against the Color Catalog (case-insensitive, in-memory — no N+1), else blank.
- Fixed, case/space-tolerant 15-column schema incl. `weight_used`, derived
  export-only `remaining`, and `last_used`. Validation reuses `SpoolCreate`.
- Frontend: `SpoolCsvImportModal` (file pick → preview table → import valid
  rows), Import/Export buttons on the Inventory header, API client methods.
- i18n: `inventory.csv` namespace added to all 11 locales (parity passes).
- Tests: backend integration + frontend modal tests.

## Testing

Backend: `pytest backend/tests/integration/test_inventory_csv.py` (13 passed).
Frontend: `tsc --noEmit`, `eslint`, `vitest` (modal 3 passed), `npm run check:i18n`
(5085 keys, all 11 locales in parity). `ruff check` + `ruff format --check` clean.
Also deployed and verified live (export header + columns, import round-trip,
colour resolution, remaining ignored on import).

## Checklist


maziggy (changes requested):
Blockers

1. Upload size cap. await file.read() reads the entire body into memory with no limit. An authenticated user can OOM the container with a misclicked .iso (or worse on   purpose). Bambuddy doesn't have a global HTTP-level cap, so it has to live on the route. Could you check file.size against a 5 MB cap before parsing and return a 413 with an i18n message? 5 MB is roughly 20k rows of headroom.

2. CSV formula injection on export. _cell_value writes string fields verbatim. If a note or color_name starts with =, +, -, @, tab, or carriage return, Excel / LibreOffice / Sheets evaluates it as a formula when the file is opened — CWE-1236. The PR explicitly markets export for backup and migration (= file gets opened on another machine), so the surface is real. Prefix any cell whose first char is in that set with a single quote.

3. weight_used bounds. The schema accepts any float — negative values and values exceeding label_weight go through. The CSV preview is a great place to catch it; ideally fix in the Pydantic schema if it's missing there too. Add a row error or warning if weight_used < 0 or weight_used > label_weight.

4. Color resolver silently falls back to wrong-material. In _resolve_color, when no material-specific catalog entry exists, you return matches[0] which can be ANY   material. The user sees the magic-wand "resolved" hint but not that it was a material-mismatch fallback. Either skip the fallback across materials, or surface it as a separate flag in ImportRowResult so the preview can flag it differently.

5. Spoolman mode parity. Bambuddy has a project rule that inventory features ship for both internal AND Spoolman modes in the same drop, with no spoolmanMode ?   undefined : ... UI gates. Your {!spoolmanMode && (...)} block is exactly the forbidden shape. I see your rationale ("Spoolman has its own data store") and it has merit — Spoolman does ship native CSV. So two options:

5. a. Mirror export_spools_csv / import_spools_csv in backend/app/services/spoolman_inventory.py as proxy calls into Spoolman's own CSV API, and branch the frontend onClick on spoolmanMode rather than gating the buttons. This satisfies the rule strictly.

5. b. Leave the gate but make the buttons render in Spoolman mode too, with text like "Use Spoolman's built-in CSV export" linking to the Spoolman UI.

5. I'd take (b) for this PR if it's quick on your end — easier to land, surfaces the feature for Spoolman users instead of hiding it. Let me know which way you want to go.

One small thing

6. The fixed export filename bambuddy_inventory.csv will overwrite the previous download in most browsers' default folder. Could you date-stamp it:
bambuddy_inventory_<YYYY-MM-DD>.csv?

samedyuksel:
Thanks for the thorough review! I've addressed all six points. Summary of changes:

**1. Upload size cap** — Import now rejects bodies over 5 MB with a `413` before
parsing. Checks `file.size` first, then guards the read length too (for chunked
uploads with no `Content-Length`). Returns a structured `{code: "csv_import_too_large", message}`
so the frontend can show a localized message. Constant: `MAX_CSV_IMPORT_BYTES` in `spool_csv.py`.

**2. CSV formula injection** — Every exported cell whose first character is in
`= + - @ \t \r` is now prefixed with a single quote (`_sanitize_cell` in `spool_csv.py`),
so spreadsheets render it as literal text. Verified: a `note` of `=SUM(A1)` exports as `'=SUM(A1)`.

**3. weight_used bounds** — Rows where `weight_used < 0` or `weight_used > label_weight`
are now flagged as **errors** (skipped on import) with the reason shown in the preview,
e.g. `weight_used (1500.0) exceeds label_weight (1000)`. I went with hard errors rather
than warnings to keep it consistent with the other validation failures.

**4. Color resolver fallback** — Kept the cross-material fallback (so a colour still
resolves when only another material's variant exists), but it no longer does so silently:
`_resolve_color` returns a `cross_material` flag, surfaced as `cross_material_color` on
`ImportRowResult` and rendered as a ⚠️ warning icon in the preview row.

**5. Spoolman-mode parity** — Went with option (b). Removed the `{!spoolmanMode && …}`
gate; the Import/Export buttons now render in both modes. In Spoolman mode they're
disabled with a tooltip pointing users at Spoolman's built-in CSV import/export.

**6. Export filename** — Now date-stamped: `bambuddy_inventory_YYYYMMDD.csv`.

Added tests for each (oversized→413, formula-injection neutralisation, weight_used
bounds, cross-material flagging) — backend suite is 20/20, frontend tsc/eslint/vitest
clean, i18n parity holds across all 11 locales (2 new strings added). Also deployed
and verified each fix live.

Let me know if you'd prefer warnings over hard errors for #3, or any other adjustments.

maziggy (changes requested):
Blocker — color resolver false-positive on generic catalog entries. _resolve_color in backend/app/services/spool_csv.py:142:

exact = next((e for e in matches if material_l and e.material and e.material.lower() == material_l), None)

ColorCatalogEntry.material is nullable (backend/app/models/color_catalog.py:18) and a NULL row is the project convention for "matches any material". The e.material and … clause treats those NULL-material entries as non-matches, falls through to matches[0], and sets cross_material=True — so a perfectly correct generic match shows the user a yellow warning that's factually wrong. Suggested fix:

exact = next(
      (e for e in matches if e.material is None or (material_l and e.material.lower() == material_l)),
      None,
  )

Worth a matching test — a material=None catalog row resolving a PLA CSV row, asserting cross_material_color=False.

Important — preview swatch drops the alpha byte. SpoolCsvImportModal.tsx:174:

style={{ backgroundColor: `#${row.rgba.slice(0, 6)}` }}

Project convention since the cream-rewrite removal is getSwatchStyle(rgba) / spoolColorString(rgba) from frontend/src/utils/colors.ts, which paints the checkerboard   underlay for alpha=00 instead of rendering it as solid black. The rest of the inventory surface uses those helpers — the CSV preview should match.

Important — wiki PR is stale on Spoolman-mode behavior. bambuddy-wiki PR #41 still says "in Spoolman mode the buttons are hidden". The code does visible-but-disabled   with a tooltip now. Worth flipping to: "In Spoolman mode the buttons are disabled with a tooltip pointing at Spoolman's own CSV import/export."

  Smaller follow-ups:

- inventory.py:1015-1019 — the body is fully read before the 5 MB post-check fires on chunked uploads. Comment overpromises a pre-read check that only works when   Content-Length is set. Either swap to a chunked-read early-bail, or honest-up the comment.
- weight_used_baseline is left at 0 on import, so a partially-used spool's "Total Consumed" counter starts at the imported weight_used. Defensible either way — but call it out in the wiki schema row so users importing partial spools aren't surprised.
- No dedup / re-import protection. Double-clicking Import duplicates everything (no unique constraints on Spool). At minimum, soft-warn in the preview when N rows match
  an existing brand+material+color.
- Wiki schema table says "Only material is required" — true, but label_weight silently defaults to 1000 g, which mis-computes "remaining" for 750g spools. One-line note   against the relevant rows would prevent the support ticket.
- CSV intentionally omits storage_location, category, low_stock_threshold_pct. storage_location was the subject of #1291; users who built their library on it lose it on   round-trip. Either include the columns or explicitly note they're not part of the round-trip in the wiki.
  - Edge case: a user note starting literally with '= round-trips lossily (export leaves it, import strips the '). Very low likelihood — accept silently or mention as a known limitation, your call.

samedyuksel:
Thanks for the thorough second pass — all of it is addressed. Code is pushed; wiki PR #41 updated alongside.

Blocker

✅ Generic-material false positive. _resolve_color now treats a NULL-material catalog entry as the "matches any material" convention, so it counts as an exact match instead of falling through to matches[0] with cross_material=True. Added the test you suggested — a material=None row resolving a PLA import asserts cross_material_color=False.
Important

✅ Preview swatch alpha. Swapped the inline #${rgba.slice(0,6)} for getSwatchStyle(rgba) from utils/colors.ts, so transparent fills paint the checkerboard underlay like the rest of the inventory surface instead of solid black.
✅ Stale wiki on Spoolman mode. Flipped to "disabled with a tooltip pointing at Spoolman's own CSV import/export" in wiki PR #41.
Smaller follow-ups

✅ Chunked-upload cap. Replaced the pre-read comment with an actual bounded chunked read — it accumulates in 64 KB chunks and bails with 413 the moment the body crosses the cap, so a chunked upload with no Content-Length can't stream past it into memory.
✅ Omitted columns. Added storage_location, category, and low_stock_threshold_pct to the schema so the round-trip is lossless (rather than just documenting the gap). New round-trip + bounds tests cover them.
✅ Re-import dedup. Added a non-blocking soft-warn: the preview flags any row whose material+brand+color_name matches an existing active spool (one in-memory query, no N+1). Behavior is unchanged — the row is still imported as a new spool — it just catches a double-click / re-upload. Documented in the wiki.
✅ weight_used_baseline / label_weight default. Both called out in the wiki schema notes (baseline stays at 0 so Total Consumed jumps for partial imports; label_weight silently defaults to 1000 g, which skews remaining for 750 g spools).
✅ '= round-trip edge case. Noted as a known limitation in the wiki. The single-quote stripping is the exact inverse of the export-side formula guard, so the only lossy input is a literal ' followed by a formula char — extremely rare, and called out.
One extra correctness fix while in here: the formula-injection guard from round 1 was adding a leading ' on export without import stripping it, so a =-leading note accumulated a quote on every export→import cycle. Import now reverses it symmetrically (_desanitize_cell), with a round-trip test.

PR #1676, opened 2026-06-07
feat(inventory): by-tag spool lookup, readable with Manage-Inventory keys (#1663)

bambuman:
## Description

Adds a by-tag spool lookup endpoint and lets a Manage-Inventory API key read inventory, so NFC inventory integrations can dedupe a scan with a single, narrowly-scoped key instead of pulling the whole inventory under the broad `can_read_status` scope.

## Related Issue

Fixes #1663

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#42

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## Changes Made

- Add `GET /api/v1/inventory/spools/by-tag?tray_uuid=…&tag_uid=…` — looks up a single spool; `tray_uuid` (the value the AMS reports over MQTT) is matched first, `tag_uid` is the fallback. Inputs are normalised, archived spools excluded by default, `400` if neither identifier is given, `404` if nothing matches.
- Extend `_APIKEY_SCOPE_BY_PERMISSION` so a permission can map to **either** a single scope flag **or** a tuple of flags (any one grants it). `INVENTORY_READ` now maps to `("can_read_status", "can_manage_inventory")`.
- This is deliberately additive: rather than re-pointing `INVENTORY_READ` at `can_manage_inventory` (which would break existing integrations reading inventory with a `can_read_status` key), read-status keys keep working unchanged and Manage-Inventory keys additionally gain inventory read.
- `_resolve_apikey_scope` → `_resolve_apikey_scopes` (returns a tuple); the scope check passes if any mapped flag is set.

## Testing

- [x] I have tested this on my local machine

New `test_spool_by_tag_lookup.py` (match / normalise / tray-uuid precedence / 400 / 404 / archived / route-collision, plus the three API-key scope cases). Updated the RBAC scope matrix + introspection tests for tuple-valued mappings. `ruff check`/`format` clean; by-tag, RBAC, route-auth-coverage and endpoint-auth suites pass (69 tests).

## Checklist

- [x] My code follows the project's coding style
- [x] I have commented my code where necessary
- [x] My changes generate no new warnings
- [x] I have tested my changes thoroughly

PR #1700, opened 2026-06-09
feat(inventory): by-tag spool lookup, readable with Manage-Inventory keys (#1663)

### LINKED ISSUE #1663: [Feature]: Find a spool by NFC tag, and let Manage-Inventory keys read inventory
### Problem or Use Case

Hi! 👋 I maintain BambuMan (https://github.com/bambuman/BambuMan.App), a small
open-source app that reads the NFC tag on Bambu Lab spools and adds them to your
filament inventory. It now supports Bambuddy as a backend — scan a spool and it's
created or updated in Bambuddy automatically. All a user needs is an Android phone
with NFC — no extra hardware (no kiosk, scale, or SpoolBuddy).

The tag carries a `tray_uuid` — the same value the AMS reports over MQTT — so a spool
BambuMan writes lines up with Bambuddy's existing AMS matching automatically, with no
extra action when that spool is later loaded into the AMS. That `tray_uuid` is also
the natural key for finding a spool again.

On each scan the app has to check whether that tag already exists, so it can update
the spool instead of creating a duplicate. Today the only way is
`GET /api/v1/inventory/spools` (list everything) and match `tray_uuid` / `tag_uid`
on the client.

Two downsides:
- It pulls the whole inventory on every scan just to find one spool.
- Listing spools needs `INVENTORY_READ` → `can_read_status`, which is broad (printer
  status, camera, stats, websocket…). So an inventory-only integration has to ask for
  far more access than it needs — a `Manage Inventory` key can create/update/delete
  spools but can't read them back.

### Proposed Solution

Two small things would make NFC inventory integrations much cleaner:

1. A by-tag lookup endpoint, e.g.
   `GET /api/v1/inventory/spools/by-tag?tray_uuid=…&tag_uid=…`
   returning the matching spool (or 404) — so we don't list the whole inventory
   just to dedupe one scan.

2. Inventory read access for a Manage-Inventory key — a key that can already
   create/update/delete spools being able to read them too. Scoping the new endpoint
   under `can_manage_inventory` instead of `can_read_status` would cover our case
   nicely, but the exact mapping is your call — happy to go with whatever fits
   Bambuddy's auth model best.

### Alternatives Considered

- Listing `GET /inventory/spools` and caching it client-side (what we do now) — works,
  but needs the broad `can_read_status` scope and re-pulls the inventory.
- The SpoolBuddy `POST /spoolbuddy/nfc/tag-scanned` endpoint is `INVENTORY_UPDATE`
  (reachable with a Manage-Inventory key) and returns `{matched, spool_id}`. But it's
  an action endpoint that broadcasts kiosk events, not a lookup — and to get the actual
  spool we'd still call `GET /spools/{spool_id}`, which is `INVENTORY_READ` →
  `can_read_status` again. So every inventory read path (list and get-by-id) still
  needs the broad scope.

### Feature Category

API

### Priority

Would improve my workflow

### Mockups or Examples

_No response_

### Contribution

### Checklist

bambuman:
> **Supersedes #1676** — retargeted to `dev` per CONTRIBUTING.md (branch was rebased `--onto dev`; #1676 could not be reopened after the force-push).

## Description

Adds a by-tag spool lookup endpoint and lets a Manage-Inventory API key read inventory, so NFC inventory integrations can dedupe a scan with a single, narrowly-scoped key instead of pulling the whole inventory under the broad `can_read_status` scope.

## Related Issue

Fixes #1663

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#42

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

## Changes Made

- Add `GET /api/v1/inventory/spools/by-tag?tray_uuid=…&tag_uid=…` — looks up a single spool; `tray_uuid` (the value the AMS reports over MQTT) is matched first, `tag_uid` is the fallback. Inputs are normalised, archived spools excluded by default, `400` if neither identifier is given, `404` if nothing matches.
- Extend `_APIKEY_SCOPE_BY_PERMISSION` so a permission can map to **either** a single scope flag **or** a tuple of flags (any one grants it). `INVENTORY_READ` now maps to `("can_read_status", "can_manage_inventory")`.
- This is deliberately additive: rather than re-pointing `INVENTORY_READ` at `can_manage_inventory` (which would break existing integrations reading inventory with a `can_read_status` key), read-status keys keep working unchanged and Manage-Inventory keys additionally gain inventory read.
- `_resolve_apikey_scope` → `_resolve_apikey_scopes` (returns a tuple); the scope check passes if any mapped flag is set.

## Testing

New `test_spool_by_tag_lookup.py` (match / normalise / tray-uuid precedence / 400 / 404 / archived / route-collision, plus the three API-key scope cases). Updated the RBAC scope matrix + introspection tests for tuple-valued mappings. `ruff check`/`format` clean; by-tag, RBAC, route-auth-coverage and endpoint-auth suites pass (69 tests).

## Checklist


bambuman:
Sorry for the rule miss here — Claude had added a `CLAUDE.local.md` note that wrongly overrode the base branch to `main`, and I didn't catch it before opening the PR. Now fixed: rebased onto `dev` and retargeted to `dev` per CONTRIBUTING.md. The feature commit sits cleanly on `dev`; ruff clean, by-tag + RBAC + endpoint-auth suites pass.

maziggy (changes requested):
**Important**

1. The tuple-valued mapping cleanly encodes any-of semantics, but the next reader      could easily misinterpret a tuple as and-of when they reach for the same shape for a future permission. One-line comment block addition at     `_APIKEY_SCOPE_BY_PERMISSION` would close that — something like "tuple values are evaluated with OR; permissions requiring multiple flags should add an explicit and-of helper rather than overloading this shape".

2. Error message gets slightly awkward with two flags:

API key does not have 'can_read_status' or 'can_manage_inventory' permission

Reads OK but the outer 'singular' framing rubs against the multi-flag content.
Special-case the rendering:

```python
  if len(scope_attrs) == 1:
      detail = f"API key does not have '{scope_attrs[0]}' permission"
  else:
      names = ", ".join(f"'{s}'" for s in scope_attrs)
      detail = f"API key does not have any of the required scopes ({names})"

Nits (reviewer discretion)

- func.upper(column) == value is defensive — normalize_tray_uuid / normalize_tag_uid already uppercase at the write boundary, so the column invariant holds. Cost is negligible today, but if you ever index these columns for a larger NFC inventory the index won't be used. Either leave with a comment about the legacy-row defensiveness or drop.
- Truncation asymmetry in tag_normalization.py is pre-existing (tag_uid keeps the last 16 chars, tray_uuid keeps the first 32) but the new endpoint surfaces both side by side — a one-line comment explaining why would help the next reader.
- The scope-bit cases are tested cleanly. The path I'd add: one integration test where the API key holds Permission.INVENTORY_READ.value in its permissions JSON field (not via the flag bitmap) and the by-tag call succeeds — closes the loop on the new tuple resolution path through the JSON-permission side too.

maziggy (changes requested):
One thing to revisit before merge: the scope-mapping change is wider than what #1663 needs, and broader than I'd like to ship.

The PR currently re-points Permission.INVENTORY_READ from "can_read_status" to ("can_read_status", "can_manage_inventory") at the _APIKEY_SCOPE_BY_PERMISSION layer. That single edit widens every endpoint gated on INVENTORY_READ, not just the new /spools/by-tag — ~21 endpoints across inventory.py, labels.py, spoolbuddy.py,   spoolman_inventory.py. Some of them (SpoolBuddy diagnostics queue, the full spool list, label PDF render, catalog/colors/locations, forecast pages) aren't part of the   BambuMan dedup flow and don't have to become reachable by Manage-Inventory keys just to solve the by-tag case.

In the issue you actually said:

"Scoping the new endpoint under can_manage_inventory instead of can_read_status would cover our case nicely, but the exact mapping is your call."

That's exactly the shape I'd prefer — scope the new endpoint, not the global mapping. And the codebase already has the idiom for it: RequireAnyPermissionIfAuthEnabled is imported in inventory.py and used on the two forecast endpoints (inventory.py:2026, 2110) to OR INVENTORY_READ with INVENTORY_FORECAST_READ per-route.

What I'd ask for:

1. Drop the tuple-mapping refactor. Leave _APIKEY_SCOPE_BY_PERMISSION as dict[Permission, str], leave _resolve_apikey_scope named and shaped as it was, leave   _check_apikey_permissions untouched. Revert the corresponding flatten-aware changes in test_auth_apikey_rbac.py (the introspection tests and   test_inventory_read_allowed_by_either_inventory_scope).

2. On get_spool_by_tag, swap RequirePermissionIfAuthEnabled(Permission.INVENTORY_READ) for RequireAnyPermissionIfAuthEnabled(Permission.INVENTORY_READ,   Permission.INVENTORY_UPDATE) (INVENTORY_UPDATE is what the can_manage_inventory scope already grants, so a Manage-Inventory key passes the check on this route and only this route).

3. Keep the three TestSpoolByTagApiKeyScope test cases — they're the right shape, they just now live against the per-endpoint RequireAnyPermissionIfAuthEnabled instead   of a mapping change.

bambuman:
Sorry for the delay, I have implemented the requested changes. 

Also updated the wiki PR.

PR #1710, opened 2026-06-11
fix(updater): use app_dir for all git ops when DATA_DIR is separate

francescocozzi:
## Description

On native systemd installs the service file sets DATA_DIR=<install>/data, which means settings.base_dir resolves to the data directory rather than the source tree. All git subprocess calls (remote get-url, fetch, reset --hard) were running with cwd=base_dir and safe.directory=base_dir, so git would fail with:

  fatal: not a git repository (or any parent up to mount point /)
  Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set)

This affects every native systemd install that follows the official guide, because that guide sets DATA_DIR=/srv/bambuddy/data in the service unit.

## Related Issue

Fix: derive app_dir = settings.app_dir at the top of _perform_update and use it for every git command. base_dir is kept for non-git operations (pip install path was already correct). The _origin_points_at_repo helper already received the directory as its base_dir parameter — the caller now passes app_dir as that argument, which is the correct working tree.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Testing

### DIFF
--- a/backend/app/api/routes/updates.py
+++ b/backend/app/api/routes/updates.py
@@ -194,7 +194,7 @@ async def _origin_points_at_repo(git_path: str, git_config: list[str], base_dir,
             "remote",
             "get-url",
             "origin",
-            cwd=str(base_dir),
+            cwd=str(app_dir),
             stdout=asyncio.subprocess.PIPE,
             stderr=asyncio.subprocess.PIPE,
         )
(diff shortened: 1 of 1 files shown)


maziggy:

Issue #1711, opened 2026-06-11
[Bug]: Sidecar-generated .gcode.3mf returns ZIP bytes to 3D Preview


yanglei1980:
### Product
Bambuddy

### Area
Archives / library / files

### Bug Description
A `.gcode.3mf` generated through Bambuddy's server-side slicer is saved with `file_type="gcode"`. The 3D Preview endpoint therefore returns the entire ZIP-shaped 3MF as `text/plain`. The viewer receives bytes beginning with `PK` instead of `Metadata/plate_1.gcode`, so no toolpath is rendered. The generated file itself is valid and prints successfully.

### Expected Behavior
Sidecar-generated `.gcode.3mf` outputs should be classified as `gcode.3mf`, and the preview endpoint should extract `Metadata/plate_*.gcode`. Existing affected rows should remain previewable.

### Steps to Reproduce
1. Run Bambuddy v0.2.4.6 with the matching Bambu Studio sidecar.
2. Upload an STL to File Manager.
3. Use server-side Slice and save the generated `.gcode.3mf`.
4. Click **3D Preview** for the generated file.
5. The viewer opens without a toolpath.
6. The library row has a `.gcode.3mf` filename but `file_type=gcode`.
7. `GET /api/v1/library/files/{id}/gcode` returns a body beginning with `PK`.

### Troubleshooting steps already taken
- Confirmed Bambuddy and sidecars are healthy.
- Confirmed the sliced file uploads to the A1 and prints.
- Confirmed the file contains `Metadata/plate_1.gcode`.
- Confirmed viewer assets exist and `/gcode-viewer/` returns HTTP 200.
- Reviewed #1600/#1543; those handle canonical `gcode.3mf` rows, but `slice_and_persist()` still writes `file_type="gcode"`.
- A local regression test reproduces the ZIP response.

### Printer Model
A1

### Bambuddy Version
v0.2.4.6

### SpoolBuddy Version
Not used

### Printer Firmware Version
Not relevant; the generated file prints successfully.

### Installation Method
Manual (git clone), Windows service layout; slicer sidecars run in Docker Desktop.

### Operating System
Windows 11

### Support Package (.zip) and / or relevant logs
No full support package is attached because this is an isolated response-type defect. Deterministic evidence:

```text
filename=figurine.gcode.3mf
file_type=gcode

3MF entries:
Metadata/plate_1.json
Metadata/plate_1.gcode.md5
Metadata/plate_1.gcode

[...]

Issue #1600, opened 2026-06-02
[Bug Report] When using external folder thumbnails are not displayed for gcode.3mf files


maziggy:
When using external folder thumbnails are not displayed for gcode.3mf files

<details>
<summary>System Information</summary>

```json
{
  "generated_at": "2026-06-02T07:29:19.735048",
  "app": {
    "version": "0.2.5b1",
    "debug_mode": false
  },
  "system": {
    "platform": "Linux",
    "platform_release": "6.12.47+rpt-rpi-v8",
    "platform_version": "#1 SMP PREEMPT Debian 1:6.12.47-1+rpt1 (2025-09-16)",
    "architecture": "aarch64",
    "python_version": "3.13.13"
  },
  "environment": {
    "docker": true,
    "data_dir": "/app/data",
    "log_dir": "/app/logs",
    "timezone": "Europe/Berlin"
  },
  "database": {
    "archives_total": 838,
    "archives_completed": 668,
    "printers_total": 9,
    "filaments_total": 0,
    "projects_total": 1,
    "smart_plugs_total": 9
  },
  "printers": [
    {
      "index": 1,
      "model": "P1S",
      "nozzle_count": 1,
      "is_active": true,
      "mqtt_connected": false,
      "state": "unknown",
      "firmware_version": "01.09.01.00",
      "wifi_signal": -48,
      "reachable": false,
      "ams_unit_count": 1,
      "ams_tray_count": 0,
      "has_vt_tray": true,
      "external_camera_configured": false,
      "plate_detection_enabled": false,
      "obico_enabled": false,
      "hms_error_count": 0,
      "developer_mode": true,
      "nozzle_rack_count": 0
    },
    {
      "index": 2,
      "model": "P1S",
      "nozzle_count": 1,
      "is_active": true,
      "mqtt_connected": false,
      "state": "unknown",
      "firmware_version": "01.09.01.00",
      "wifi_signal": -44,
      "reachable": false,
      "ams_unit_count": 0,
      "ams_tray_count": 0,
      "has_vt_tray": true,
      "external_camera_configured": false,
      "plate_detection_enabled": false,
      "obico_enabled": false,
      "hms_error_count": 0,
      "developer_mode": true,
      "nozzle_rack_count": 0
    },
    {
      "index": 3,
      "model": "P1S",
      "nozzle_count": 1,
      "is_active": true,
      "mqtt_connected": false,
      "state": "unknown",
      "firmware_version": "01.09.01.00",
[...]

Issue #1543, opened 2026-05-27
[Bug Report] 3D Previw works on 3mf or stl model. After slice model 3d preview show emty bed


maziggy:
3D Previw works on 3mf or stl model. After slice model 3d preview show emty bed

### Screenshot
![Bug Report Screenshot](https://raw.githubusercontent.com/maziggy/bambuddy/bug-report-assets/screenshots/1d238498723240629b1631f15b4980ac.png)

<details>
<summary>Reporter Contact</summary>

Email: vladimir@svab.sk

</details>

<details>
<summary>System Information</summary>

```json
{
  "generated_at": "2026-05-27T14:37:43.003377",
  "app": {
    "version": "0.2.4.3",
    "debug_mode": false
  },
  "system": {
    "platform": "Linux",
    "platform_release": "7.0.0-15-generic",
    "platform_version": "#15-Ubuntu SMP PREEMPT_DYNAMIC Wed Apr 22 16:06:43 UTC 2026",
    "architecture": "x86_64",
    "python_version": "3.14.4"
  },
  "environment": {
    "docker": false,
    "data_dir": "/opt/[user]/data",
    "log_dir": "/opt/[user]/logs",
    "timezone": "Europe/Bratislava"
  },
  "database": {
    "archives_total": 23,
    "archives_completed": 15,
    "printers_total": 1,
    "filaments_total": 0,
    "projects_total": 1,
    "smart_plugs_total": 0
  },
  "printers": [
    {
      "index": 1,
      "model": "P1S",
      "nozzle_count": 1,
      "is_active": true,
      "mqtt_connected": true,
      "state": "RUNNING",
      "firmware_version": "01.10.00.00",
      "wifi_signal": -47,
      "reachable": false,
      "ams_unit_count": 1,
      "ams_tray_count": 4,
      "has_vt_tray": true,
      "external_camera_configured": false,
      "plate_detection_enabled": false,
      "obico_enabled": false,
      "hms_error_count": 0,
      "developer_mode": true,
      "nozzle_rack_count": 0
    }
  ],
  "settings": {
    "advanced_auth_enabled": "false",
    "smtp_auth_enabled": "true",
    "auth_enabled": "false",
    "setup_completed": "true",
    "auto_archive": "true",
    "save_thumbnails": "true",
    "capture_finish_photo": "true",
    "default_filament_cost": "25.0",
    "currency": "EUR",
    "energy_cost_per_kwh": "0.15",
[...]

Issue #1522, opened 2026-05-24
[Bug]: 3D Preview Not Loading In File Manager


iitazz:
### Product

Bambuddy

### Area

Archives / library / files

### Bug Description

The 3D Preview option doesnt work on a sliced file in File Manager but once it was sent to print and going to archive section, the 3D preview feature now works correctly.

1. File manager view
<img width="1504" height="834" alt="Image" src="https://github.com/user-attachments/assets/4c269ec2-176f-4634-8f99-16db34db7e3a" />
<img width="194" height="269" alt="Image" src="https://github.com/user-attachments/assets/d72c3882-4f4b-4284-b7ab-e172078563fb" />

2. Archive view
<img width="1510" height="839" alt="Image" src="https://github.com/user-attachments/assets/4b53935f-86ba-4060-bf16-77cbfdef046e" />

<img width="311" height="505" alt="Image" src="https://github.com/user-attachments/assets/ca247580-8612-479c-8e33-d095ff20b922" />

### Expected Behavior

Expecting the preview to load the same globally 

### Steps to Reproduce

1. Slice a file
2. Preview the sliced file > won't work in File manager

### Troubleshooting steps already taken

- Built in diagnostic didn't find anything
- Logs doest seem to track it

### Printer Model

P2S

### Bambuddy Version

v0.2.5b1

### Printer Firmware Version

01.02.00.00

### Installation Method

Manual (git clone)

### Operating System

Linux (Ubuntu/Debian)

### Support Package (.zip) and / or relevant logs

[bambuddy-support-20260524-232432.zip](https://github.com/user-attachments/files/28199849/bambuddy-support-20260524-232432.zip)

### Final checks

- [x] I searched existing (open AND closed) issues and this bug hasn't already been reported or resolved
- [x] I checked the [Troubleshooting wiki](https://wiki.bambuddy.cool/reference/troubleshooting/) and the relevant feature page for my issue
- [x] I am using the latest version of Bambuddy (or the latest daily build)
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled
- [x] For any connection / printing / camera issue, I ran the in-app Connection Diagnostic and included the result above

maziggy:
Can you please upload the file you want to preview?

iitazz:
Here

[...]

PR #1746, opened 2026-06-14
fix(printers): serve raw-gcode .gcode.3mf files in G-code Preview (don't 400 "Invalid 3MF file")

dadequate:
## Problem
`get_printer_file_gcode` (`/api/v1/printers/{id}/files/gcode`) assumes any `.gcode.3mf` is a ZIP container. Some printers (observed on P1S/P1P) store a sliced job on the SD card as **raw G-code under a `.gcode.3mf` filename** — not a real ZIP. `zipfile.ZipFile` then raises `BadZipFile`, the endpoint returns `400 "Invalid 3MF file"`, and the File Manager's **G-code Preview** dead-ends with "Failed to load file" even though the file is valid and prints fine.

## Reproduce
1. On a P1-series printer, a sliced job is saved to SD as e.g. `Plate 0.gcode.3mf`.
2. That file is plain G-code (`; HEADER_BLOCK_START` / `; BambuStudio ...`, no `PK` ZIP signature).
3. `GET /api/v1/printers/{id}/files/gcode?path=/Plate%200.gcode.3mf` -> `400 {"detail":"Invalid 3MF file"}`.

## Fix
On `BadZipFile`, sniff for raw G-code (leading `;`/`G`/`M`, or `HEADER_BLOCK_START` in the first 2 KB) and serve it as `text/plain`; otherwise keep the original error. Minimal, no behavior change for real 3MF zips.

## Scope / note
This restores the **G-code Preview** tab. The **3D Model** tab stays empty for these files by design — a sliced G-code has no mesh (expected, not part of this fix).

## Tested
Verified on a live P1-series printer running bambuddy 0.2.4.6: a raw-gcode `.gcode.3mf` on the SD that previously returned `400 "Invalid 3MF file"` from `GET .../files/gcode` now returns `200 text/plain` (~6.9 MB) with this patch, so the G-code Preview can load it. The handler is byte-identical in 0.2.4.6, 0.2.4.7, and the latest daily beta.

## Related
#1711, #1543, #1600, #1522 (related preview / `.gcode.3mf` reports on other code paths).

### DIFF
--- a/backend/app/api/routes/printers.py
+++ b/backend/app/api/routes/printers.py
@@ -1194,6 +1194,12 @@ async def get_printer_file_gcode(
                 gcode_content = zf.read(gcode_files[0])
                 return Response(content=gcode_content, media_type="text/plain")
         except zipfile.BadZipFile:
+            # Some printers store a sliced job as raw gcode under a ".gcode.3mf"
+            # name (not a real ZIP container). Detect raw gcode and serve it for
+            # the G-code Preview instead of failing the whole preview.
+            head = data[:1024].lstrip()
+            if head[:1] in (b";", b"G", b"M") or b"HEADER_BLOCK_START" in data[:2048]:
+                return Response(content=data, media_type="text/plain")
             raise HTTPException(status_code=400, detail="Invalid 3MF file")
 
     raise HTTPException(status_code=400, detail="Unsupported file type")


maziggy:
Sorry, you don't followed our rules from CONTIBUTING.md.

PR #1751, opened 2026-06-15
Fix K-profiles all resolving to 0.4mm nozzle diameter

### LINKED ISSUE #1748: [Bug]: All K-Profiles are Highflow 0.4mm
### Product

Bambuddy

### Area

Filament / AMS / Spoolman

### Bug Description

Hey there!

It seams with my A1Mini (dont think it matters) all K-Profiles show up as 0.4mm and high flow nozzle.. leading to auto kprofile selection to just fail.

Does not matter if the k-profile is setup in bambulab or directly in bambuddy, as soon as its saved it shows up as 0.4mm and high flow. 

Interestingly enough the filtering on the profile page works, so klicking on 0.8 shows just profiles for 0.8. But when selecting a profile it shows 0.4 and high flow. 

### Expected Behavior

K-Profile should stick with the correct settings

### Steps to Reproduce

1. Go to Profiles `/profiles`
2. klick on K_Profile
3. Klick new profile
4. create standard flow 0.8
5. then klick the new profile in the list
6. -> 0.4 high flow

Image
Screenshot shows details, also the s0.4 in the list, and the 0.8 selection

not applicable afaik.

### Printer Model

A1 Mini

### Bambuddy Version

0.2.4.7

### SpoolBuddy Version

_No response_

### Printer Firmware Version

01.07.02.00

### Installation Method

Docker

### Operating System

Docker

[bambuddy-support-20260615-010510.zip](https://github.com/user-attachments/files/28936244/bambuddy-support-20260615-010510.zip)

### Screenshots

Image

### Additional Context

_No response_

### Final checks

Liquidmasl:
## Description

K-profiles always resolve to `0.4mm` nozzle diameter regardless of the actual nozzle, because the `extrusion_cali_get` parser ignores the diameter the printer sends.

The printer reports `nozzle_diameter` **only on the response envelope**, never on the individual `filaments[]` entries:

```jsonc
"print": {
  "command": "extrusion_cali_get",
  "nozzle_diameter": "0.8",        // only here
  "filaments": [
    { "setting_id": "...", "filament_id": "...", "name": "...",
      "k_value": "...", "cali_idx": 914 }   // no nozzle_diameter
  ]
}
```

`_handle_kprofile_response` reads the envelope into `response_nozzle` (`bambu_mqtt.py:3978`) but then builds each profile with `f.get("nozzle_diameter", "0.4")` (`:4013` and `:4040`). Since the filament entries never carry that field, every profile falls through to the hardcoded `"0.4"` default.

This fix uses the envelope value as the fallback instead. Filtering by nozzle worked already because that path passes the diameter as the query — only the per-profile value was wrong, which is what broke the display and auto-matching.

## Related Issue

Fixes #1748

Full root-cause analysis (with the raw MQTT payload and request logs) is in the issue thread.

## Documentation

**Pick one**:
- [x] No docs update required — reason: bug fix with no new config/flags/ports/UI; restores correct nozzle-diameter parsing.

## Type of Change

## Changes Made

- `_handle_kprofile_response`: fall back to the envelope `response_nozzle` instead of the hardcoded `"0.4"` when a filament entry has no `nozzle_diameter` (both the broadcast and pending-response parse paths).

## Testing

- [ ] I have tested with my printer model: A1 Mini

> Note: this is opened as a **draft**. The fix is derived from the captured `extrusion_cali_get` MQTT payload and the request logs attached to #1748, but I have not yet validated it against hardware. Happy to add a regression test (the envelope-only `nozzle_diameter` case is straightforward to fixture) and confirm on an A1 Mini if the approach looks right.

## Checklist

## Additional Notes

There is a second, independent issue visible in the same logs — a response/request correlation race (`_expected_kprofile_nozzle` is a single shared field matched by nozzle rather than `sequence_id`, so concurrent 0.4/0.8 requests discard each other's valid responses and time out). That is **not** addressed here — it needs a proper seq_id-based correlation refactor and concurrency testing. Details are in #1748.


maziggy (changes requested):
Thanks for the careful root-cause analysis — the captured payload makes the diagnosis unambiguous and the diameter fix is correct. The envelope-level nozzle_diameter is the right fallback for both branches, and the or chain is safe given Bambu nozzle values are never falsy strings.

Two things before this can merge — please bundle them into the same PR rather than separate follow-ups:

1. Add the flow-type half of the original bug. The report covered both wrong diameter AND wrong flow type, and your wire capture shows nozzle_id is also missing from the per-filament entries. The two render paths currently default differently on empty  nozzle_id: KProfilesView.tsx list card calls getFlowTypeLabel("") which returns "S", the modal at :187 does profile?.nozzle_id ? getNozzleTypePrefix(...) : 'HH00' which returns High Flow. That asymmetry is exactly the "list shows s0.4, detail shows high flow" split you described. Please investigate where the flow type can actually be derived (does the printer echo nozzle_id on any other call? cached at save time? from live nozzle state?), or — if no reliable source exists — at minimum align both surfaces on the same default so they stop disagreeing.

2. Regression tests — the diameter fixture you already described (envelope nozzle_diameter=0.8, filament entries without the field, assert parsed nozzle_diameter is   0.8), plus a second case for envelope-also-missing falling through to the literal 0.4, plus equivalent coverage for the flow-type fix once you land it. backend/tests/unit/services/test_bambu_mqtt.py is the right home.

Liquidmasl:
Thanks — pushed both. Quick rundown of what's in now and one open question for you.

**Diameter (the approved fix) + tests.** Added `TestKProfileNozzleDiameterParsing` in `test_bambu_mqtt.py`: envelope fallback (`0.8` → parsed `0.8`), envelope-also-missing → literal `0.4`, and an explicit per-filament diameter overriding the envelope.

**Flow type — I chased your "where can it be derived" question down, and the answer is: it can't, for fetched profiles.** Two separate `extrusion_cali_get` captures from my A1 Mini (0.8 and 0.4 nozzles), and in both the `nozzle_diameter` is on the **envelope only** and the `filaments[]` entries carry **no `nozzle_id` and no material/flow field at all**. Checking the other sources you listed:
- **Other calls / printer status:** status only ever sets `nozzle_type` (material: `hardened_steel`/`stainless_steel`) and `nozzle_diameter` on the live nozzle — it never sets `nozzle_id`. So flow isn't echoed anywhere.
- **Cached at save:** only profiles created *through* bambuddy get an `HH00`/`HS00` (from the user's pick). Printer-origin profiles never had one.
- **Live nozzle state:** material only, a different vocabulary, no mapping to the `HH`/`HS` flow code.

So the flow type is genuinely **unknown** for fetched profiles. The UI was inventing it — and the two surfaces invented *different* values: the list card defaulted empty → Standard (`getFlowTypeLabel` fell through to `'S'`), the detail window defaulted empty → `HH00` High Flow (`:187`). That asymmetry is exactly the "list shows s0.4, detail shows High Flow" split in the report.

Rather than fabricate, I made it represent unknown:
- `getFlowTypeLabel` returns `''` for an unknown `nozzle_id` (no longer defaults to Standard)
- detail window shows blank for an existing profile with no `nozzle_id`
- **new** profiles default to `HS00` (Standard) — to match your backend default (`set_kprofile` uses `"HS00-…"`), and since High Flow is the rarer, printer-specific case. (Was `HH00`.)
- save guards against writing a malformed `"-0.4"` when the type is unknown (keeps the original `nozzle_id`)
- fixed an inverted flow label in the type-change auto-namer (`HS00` was being labelled `HF`)
- copy path defaults to `HS00` instead of `HH00`

Auto-apply is unaffected — `extrusion_cali_sel` matches on `cali_idx`/`filament_id`/`nozzle_diameter`, never `nozzle_id`.

**Open question for you (your call, didn't want to design your filter):** with unknown represented as blank, the flow filter (`:899-900`, `startsWith('HH')`/`'HS'`) shows unknown profiles only under "Any". Do you want an explicit "Unknown" filter option, or is "Any-only" fine? Happy to add whichever you prefer.

Minor: the comment at `schemas/kprofile.py:11` calls `HS00` "hardened steel" — it's actually the flow code (Standard); material is the separate `nozzle_type` field.

maziggy:
See comment in realted issue. Thanks.

PR #1797, opened 2026-06-22
Fix camera port diagnostic for A1/P1 printers

lesbass:
## Summary
- use the model-specific camera port in the connection diagnostic so A1/P1-family printers probe chamber-image port 6000 instead of RTSPS 322
- keep the historical RTSPS 322 fallback when the printer model is unknown or when running the pre-save diagnostic
- pass the camera protocol/port to the UI and render the diagnostic title/warning with those values across locales

## Why
A1/A1 Mini/P1 printers can have the camera stream working on port 6000 while port 322 is closed. The existing connection diagnostic always probed 322, which produced a false warning even though the camera route and live view use the chamber-image protocol successfully.

## Tests
- pytest -q backend/tests/unit/services/test_printer_diagnostic.py
- npm run test -- --run src/__tests__/components/ConnectionDiagnosticModal.test.tsx
- npm run check:i18n
- npm run build


maziggy:

lesbass:
You're right, sorry. I missed the CONTRIBUTING.md workflow requirement that every contribution starts from an assigned issue before opening a PR.

I opened #1798 with the bug report and reproduction details, and I'll close this PR now. If you agree with the scope and assign the issue, I can reopen or submit a new PR targeting the expected branch and using the full PR template.

PR #1814, opened 2026-06-24
[Fix] Forecasting: Group spools by color and rework UI

### LINKED ISSUE #1310: [Feature]: Forecast Spool Name/Colour
### Problem or Use Case

Looking at the forcast in Filament - it shows the type of filament (Matte) and then when you narrow down it shows you the spool number, are we able to add a name or colour in that as well?

Image

### Proposed Solution

Names or colours in the titles

### Alternatives Considered

_No response_

### Feature Category

Spool Inventory

### Priority

Nice to have

### Mockups or Examples

_No response_

### Contribution

### Checklist

Keybored02:
## Description
This PR adds color distinction alongside material grouping, so that spools are not clumped only as "Bambu Matte PLA". A UI rework brings visual uniformity with the inventory page and removes duplicated fields.

## Related Issue

Fixes #1310

## Documentation

**Pick one**:
- [ ] Docs PR(s) linked above
- [X] No docs update required — reason: Behaviour and core functionalities are unchanged

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Day-bucketing fix in `computeHistoryRate` (concurrent multi-spool events summed per day before rate computation)
- Added by material and by brand filtering.
- Removed duplicated lead time field from expanded spool menu
- Uniformed table size, color, text fonts to the main Inventory page
- Made action icons more visible

## Screenshots

image
image

## Testing

- [x] I have tested with my printer model:  H2C

## Checklist


maziggy:
All locales except English are missing.

maziggy (changes requested):
**Blocking**

1. Postgres migration drops the wrong constraint name. backend/app/models/filament_sku_settings.py:16 declares UniqueConstraint(..., name="uq_filament_sku"), and Base.metadata.create_all runs on startup (database.py:220) — so the actual constraint on Postgres is uq_filament_sku, not   filament_sku_settings_material_subtype_brand_key. The DROP CONSTRAINT IF EXISTS silently no-ops, the wider (..., color_name) constraint gets added on top, and the old   narrow (material, subtype, brand) constraint still rejects any second row that only differs by colour — exactly the case this PR enables. Please also drop uq_filament_sku (or rename the new one to uq_filament_sku and drop the legacy by name).

2. PR description doesn't match what the diff does. The description says spoolGroupKey includes extra_colors and effect_type and that FilamentSwatch replaces the plain   colour dot. The actual code only adds color_name to skuKey, and the dot is still a <span className="block w-5 h-5 rounded-full ..." style={colorStyle} /> — no FilamentSwatch import. A gradient PLA Matte and a plain PLA Matte with the same color_name will still collapse. Either implement the gradient/effect grouping + swatch as described, or trim the description to match scope.

3. Existing per-SKU settings get orphaned silently. Pre-existing rows have color_name=NULL, but new groups are keyed on the actual spool colour.   settingsMap.get(skuKey(s.material, s.subtype, s.brand, s.color_name)) won't match a NULL setting against a "Black" group, so every user loses their lead-time and safety-margin overrides on first load after upgrade. Either backfill the NULL row into one row per (material, subtype, brand, color_name) tuple in the spools table, or fall back to the NULL-colour lookup in settingsMap when the colour-specific key misses.

**Important**

4. i18n — only en.ts was touched. The new "Spools" column header reuses t('locations.spools') and the new filters reuse t('inventory.material') / t('inventory.brand') — please verify those keys exist in all 7 non-English locales (de/fr/it/ja/pt-BR/zh-CN/zh-TW). The CSV header literal 'Color' at ForecastPanel.tsx:1325 needs an i18n key + translations too — check-i18n-parity.mjs will fail CI on the English leak.

5. groupHasReset bail-out is overly aggressive. If any one spool in the group has weight_used_baseline > 0, the entire group's history rate gets thrown out and falls   back to a single delta — even when the other spools in the group have clean histories. Better to drop just the reset spool's records from groupHistory and keep the rest.

**Nits**

- r.created_at.slice(0, 10) + new Date(day) anchors the day buckets to UTC. Fine for forecasting but worth a one-line comment so the next reader doesn't wonder.
- idx[3] != "u" in the SQLite path could use a one-liner comment that origin='u' means auto-UNIQUE.

Keybored02:
> i18n — only en.ts was touched.

It was a naming fix exclusive to the en locale.

All 7 locales have those keys. Replaced the three hardcoded string literals with t('inventory.brand'), t('inventory.subtype'), and t('inventory.color')  keys that already exist in all locales.

maziggy:
Ahhh...got it.

maziggy (changes requested):
The PR text says spoolGroupKey includes extra_colors and that FilamentSwatch replaces the plain colour dot, but the actual code only adds color_name to the key, and the colour cell is still <span style={colorStyle}> reading f.group.spools[0]?.rgba. So a Bambu PLA Matte Gradient "Black" and a plain PLA Matte "Black" will still collapse into one group, and the colour cell won't show the gradient.

Either way is fine — implement extra_colors/effect_type in the key + swap to FilamentSwatch, or trim those two bullets from the description. The color_name work as   built is already a real improvement.

maziggy:
Thx.

Issue #1830, opened 2026-06-26
[Bug]: HMS Actions don't take effect on H2C — action POST returns 200 but printer doesn't act


Boa-Thomas:
### Component

Bambuddy (HMS Actions — PR #1743, ref #1419)

### Bug Description

On an **H2C**, the HMS-error modal shows the action buttons correctly, but clicking an action does **nothing on the printer** — the fault stays active and still has to be cleared at the printer's own screen.

The command channel to this H2C is fine — **chamber-fan speed control from Bambuddy is honored by the printer**, so MQTT publishes to `device/<serial>/request` do reach and act on this machine. The problem is specific to the HMS action commands.

`execute_hms_action` returns success on *publish*, not on printer *ack*, so the API replies `200 OK` and the UI looks like it worked — masking the failure.

### Steps to Reproduce

1. H2C raises HMS error **`0C00_0300_0002_000C`**.
2. Open the error in Bambuddy → action **"Ignore and resume"** is offered → click it.
3. Observe: `POST /api/v1/printers/<id>/hms/execute-action` → **200 OK**, but the H2C does not act; the error persists and must be resolved at the printer.

### Backend analysis (two candidate root causes)

**(1) Lossy `err` derivation for `hms[]`-array errors.**
In `bambu_mqtt.py`, the `hms[]` branch builds the short code as:
```python
short_code = f"{(attr >> 16) & 0xFFFF:04X}_{code & 0xFFFF:04X}"
```
For this fault — `attr = 0x0C000300`, `code = 0x0002000C` (displayed `0C00_0300_0002_000C`) — that yields **`0C00_000C`**, discarding `attr & 0xFFFF = 0x0300` and `code >> 16 = 0x0002`. The frontend strips the underscore and sends `err = "0C00000C"`, so `execute_hms_action` publishes:
```json
{"print": {"command": "idle_ignore", "err": "0C00000C", "type": 0, "sequence_id": "0"}}
```
If the firmware matches `err` against the active fault's full code, `0C00000C` won't match `0C00_0300_0002_000C` → the command is silently ignored. (This would affect any printer for `hms[]`-sourced faults; H2C is where it surfaced.)

**(2) Action → command mapping.**
[...]

PR #1841, opened 2026-06-27
fix(hms): surface uncataloged HMS errors that carry actions (#1840)

### LINKED ISSUE #1840: [Bug]: HMS errors missing from the frontend catalog are silently filtered out of the UI — actionable H2C faults (e.g. 0500_809C) never render
### Component

Bambuddy (frontend)

### Bug Description

HMS errors that are **not present in the frontend's hardcoded description catalog are silently filtered out of the UI** — including errors that have valid backend actions. The error never renders: no banner, no problem pip, no count, and (critically) no action buttons. The backend has the error and the actions; the user simply can't see or act on it.

This is **distinct from #1830** (which fixed the backend *dispatch* of HMS actions). #1830 makes the action *reach the printer* — but for any HMS code missing from the frontend catalog, the action button never renders in the first place, so the #1830 fix is unreachable. It's also the "HMS errors don't appear at all" half of the original report.

### Environment

- Bambuddy `v0.2.5b2-daily.20260627` (the post-#1830 daily, bundle `index-CWnKKhV6.js`), Docker (`network_mode: host`)
- Printer: **Bambu H2C**, firmware `01.02.00.00`, dual-nozzle
- Auth disabled

### Reproduction

1. Run a print on the H2C that triggers an HMS fault that pauses the print.
2. Backend correctly captures it — `GET /api/v1/printers/1/status` returns:

```
state: PAUSE   connected: True
  code=0x809c   attr=0x0500809C  full_code=0500809C            sev=3  actions=['IGNORE_RESUME','PROBLEM_SOLVED_RESUME']
  code=0x20005  attr=0x05000600  full_code=0500060000020005    sev=6  actions=[]
  code=0x20006  attr=0x05000600  full_code=0500060000020006    sev=6  actions=[]
```

3. The websocket pushes the same `hms_errors` (identical shape to the REST `PrinterStatus`, verified field-by-field via `printer_state_to_dict`).
4. **UI shows nothing** — no HMS indicator anywhere, in any browser (not a stale-socket / cache issue).

### Root Cause

In the bundle, every HMS surface (problem pip, per-card count, the errors panel, the `DW()` status→`error` mapper) gates on `Gp(hms_errors)`:

```js
function Wp(e,t){let n=e>>16&65535||(e>>8&255)<<8|e&255,r=t&65535;
  return `${n.toString(16).padStart(4,"0").toUpperCase()}_${r.toString(16).padStart(4,"0").toUpperCase()}`}

function Gp(e){return e.filter(e=>{let t=parseInt(e.code.replace("0x",""),16)||0;
  return Up[Wp(e.attr,t)]!==void 0})}      // <-- drops anything not in the Up catalog

// e.g.: x = h.some(({status:e}) => !e?.connected || !e.hms_errors ? false : Gp(e.hms_errors).length>0)
```

`Up` is a hardcoded 853-entry map of `MMMM_EEEE` → description (e.g. `"0300_4005":"The hotend cooling fan speed is abnormal."`). `Gp` keeps only errors whose `Wp(attr,code)` key exists in `Up`.

The keys computed for the three H2C faults:

| backend `attr` / `code`            | `Wp()` key  | in `Up`? | has backend actions |
|------------------------------------|-------------|----------|---------------------|
| `0x0500809C` / `0x809c`            | `0500_809C` | **no**   | **yes** (IGNORE_RESUME, PROBLEM_SOLVED_RESUME) |
| `0x05000600` / `0x20005`           | `0500_0005` | **no**   | no |
| `0x05000600` / `0x20006`           | `0500_0006` | **no**   | no |

None of `0500_809C`, `0500_0005`, `0500_0006` exist anywhere in the bundle (catalog `0500_*` entries only cover the `0500_40xx` range). So all three are filtered out and the UI renders nothing — even though `0500_809C` carries actions the backend (and #1830) can dispatch.

### Impact

- HMS faults whose code isn't in the bundled catalog are invisible in the UI — the user sees a silently paused printer with no explanation and no way to act.
- For codes that *do* have actions, this defeats #1830 entirely: the action button is never rendered, so the now-correct dispatch path is unreachable.

### Suggested Fix

Don't hard-filter on catalog membership. Render uncataloged HMS too:

- Keep any error where `actions.length > 0` regardless of catalog membership.
- For uncataloged errors, fall back to a generic label (the raw `MMMM_EEEE` / `full_code`) + the "View on Bambu Lab Wiki" link, instead of dropping them.
- i.e. use `Up[...]` for the *description* but not as a *visibility gate*.

### Secondary observation (low confidence, possibly unrelated)

Backend sends `severity` as an int (`HMSErrorResponse.severity`: 1–4, but these H2C faults report `6`), while the error-detail component compares `e.severity === "error"` (a string) for its icon/color — so severity styling appears to never match the numeric value. Flagging in case it's the same path; the filter above is the actual cause of the no-render.

### Checklist

- [x] I have searched existing issues
- [x] I am using the latest version of Bambuddy (daily)

Boa-Thomas:
Fixes #1840.

### Problem

HMS errors whose `MMMM_EEEE` code isn't in the local `ERROR_DESCRIPTIONS` catalog are filtered out in **two** places in `HMSErrorModal.tsx`:

- `filterKnownHMSErrors()` — drives the badge count, status classification and the "view HMS" affordance across `PrintersPage`/`BulkPrinterToolbar`;
- the modal's internal `knownErrors` filter — drives what the modal lists.

Both keep only `ERROR_DESCRIPTIONS[shortCode] !== undefined`. So an HMS error that the backend has **actions** for, but whose code we don't describe locally, renders **nothing**: no badge, no count, no modal entry, and therefore none of the new #1830 action buttons. The user just sees a silently paused printer with no way to respond from the UI.

### Fix

Surface an error when it is **described OR carries actions** (`isDisplayableHMSError`), used by both filter sites, and fall back to a generic description for uncataloged codes:

- actionable faults always render (badge + modal + Ignore/Resume buttons);
- uncataloged **action-less** codes stay hidden, so no new noise is introduced;
- uncataloged actionable codes get a fallback description + the existing "View on Bambu Lab Wiki" link instead of an empty line.

### Testing

- `npm run build` (`tsc -b` typecheck + `vite build`) — clean
- `eslint` on the changed files — clean
- `vitest src/__tests__/components/HMSErrorModal.test.tsx` — 14 passed, incl. two new cases (an uncataloged error with actions now renders its `[code]`, both action buttons, and the fallback description)
- **Validated live on an H2C (fw 01.02.00.00):** `0500_809C` *"Build plate not properly positioned"* — a false positive from a markerless third-party build plate — previously rendered nothing in the UI and now shows correctly with its `IGNORE_RESUME` / `PROBLEM_SOLVED_RESUME` buttons.

### Notes

Complements #1830 (which fixed the action **dispatch**): without this, the action buttons that #1830 wired up are never reachable for any code missing from the frontend catalog — including safety-relevant warnings like build-plate collision.

### DIFF
--- a/frontend/src/__tests__/components/HMSErrorModal.test.tsx
+++ b/frontend/src/__tests__/components/HMSErrorModal.test.tsx
@@ -25,6 +25,16 @@ const unknownError: HMSError = {
   severity: 1,
 };
 
+// Unknown code (not in the database) that nonetheless carries user actions —
+// e.g. H2C 0500_809C "Build plate not properly positioned". Must NOT be hidden. (#1840)
+const actionableUnknownError: HMSError = {
+  attr: 0x0500809C,
+  code: '0x809C',
+  severity: 3,
+  actions: ['IGNORE_RESUME', 'PROBLEM_SOLVED_RESUME'],
+  full_code: '0500809C',
+};
+
 describe('HMSErrorModal', () => {
   const defaultProps = {
     printerName: 'Test Printer',
@@ -58,6 +68,21 @@ describe('HMSErrorModal', () => {
     it('shows no errors message when errors array is empty', () => {
       render(<HMSErrorModal {...defaultProps} errors={[]} />);
       expect(screen.getByText('No errors')).toBeInTheDocument();
+    });
+
+    it('shows unknown errors that carry actions (never hides an actionable fault)', () => {
+      render(<HMSErrorModal {...defaultProps} errors={[actionableUnknownError]} />);
+      expect(screen.queryByText('No errors')).not.toBeInTheDocument();
+      expect(screen.getByText('[0500-809C]')).toBeInTheDocument();
+      expect(screen.getByText('Ignore this and Resume')).toBeInTheDocument();
+      expect(screen.getByText('Problem Solved and Resume')).toBeInTheDocument();
+    });
+
+    it('shows a fallback description for actionable codes missing from the catalog', () => {
+      render(<HMSErrorModal {...defaultProps} errors={[actionableUnknownError]} />);
+      expect(
+        screen.getByText('Unrecognized HMS code — open the Bambu Lab Wiki for details.'),
+      ).toBeInTheDocument();
     });
   });
 
--- a/frontend/src/components/HMSErrorModal.tsx
+++ b/frontend/src/components/HMSErrorModal.tsx
@@ -896,13 +896,20 @@ function getShortCode(attr: number, code: number): string {
   return `${module.toString(16).padStart(4, '0').toUpperCase()}_${codeNum.toString(16).padStart(4, '0').toUpperCase()}`;
 }
 
-// Helper to filter only known HMS errors (exported for use in badge counts)
+// An HMS error is worth surfacing if we have a local description for it OR it
+// carries user actions (e.g. Ignore/Resume). Hiding an actionable fault leaves
+// the user looking at a silently paused printer with no way to respond to the
+// warning from the UI — the firmware still expects an answer. See #1840.
+function isDisplayableHMSError(error: HMSError): boolean {
+  const codeNum = parseInt(error.code.replace('0x', ''), 16) || 0;
+  const shortCode = getShortCode(error.attr, codeNum);
+  return ERROR_DESCRIPTIONS[shortCode] !== undefined || (error.actions?.length ?? 0) > 0;
+}
+
+// Filter HMS errors down to the ones we should surface (described or actionable).
+// Exported for use in badge counts / status classification.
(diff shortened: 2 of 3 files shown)


maziggy:
i18n locales are missing.

Boa-Thomas:
Good catch — fixed. Added `hmsErrors.unknownCode` to all 10 other locales (de, es, fr, it, ja, ko, pt-BR, tr, zh-CN, zh-TW) with proper translations, not en fallbacks. `tsc -b` + `vite build` clean and the HMSErrorModal tests still pass (14/14).

PR #1844, opened 2026-06-28
fix(hms): dismiss paused print_error dialogs via the `ignore` verb (#1830)

Boa-Thomas:
Addresses #1830 (the part the merged dev fix doesn't cover yet).

### Problem

For a `print_error`-dialog fault, the current `hms_ignore` PAUSE branch dispatches a plain `resume`. On the **H2C** that doesn't resolve the dialog: the firmware re-evaluates the fault on resume and **immediately re-raises it**, so the print stays `PAUSE` and the action reports a 502 (after the §3 ack check). Verified live:

```
BASELINE            gcode_state=PAUSE  print_error=0x0500809C
clean_print_error   gcode_state=PAUSE  print_error=0x0        ← dialog cleared
resume (plain)      gcode_state=PAUSE  print_error=0x0500809C ← re-raised, still PAUSE
```

### What the firmware actually accepts

Sniffed `device/<sn>/request` (read-only) while clicking **Ignore this and Resume** in Bambu Handy on the live H2C:

```json
{"print": {"command": "ignore", "err": "83919004", "job_id": "0", "param": "reserve", "sequence_id": "..."}}
```

i.e. a dedicated **`ignore`** verb (not `resume`/`idle_ignore`), with `err` = the `print_error` as a **decimal** string (`0x0500809C` → `"83919004"`) and `param: "reserve"`. After this exact command the dialog cleared and the print continued. (This is also why the §2 note that "`err`/`param:"reserve"` were silently rejected" held — right fields, wrong verb.)

### Fix

`hms_ignore`'s PAUSE branch now emits the `ignore` command for **8-char (32-bit `print_error`)** faults, converting the hex `print_error` to the decimal `err`. **16-char `hms[]`-array faults keep the plain-resume path** — I haven't captured the `err` encoding for those against a live printer, so this stays conservative rather than guessing.

### Testing

- Dispatch validated against the real `BambuMQTTClient` (loaded in a Bambuddy container so the runtime deps are real): IGNORE_RESUME + PAUSE + 8-char → the `ignore` shape above; default `job_id` → `"0"`; 16-char paused → plain `resume`; RUNNING → `idle_ignore` (unchanged); DONT_REMIND + paused → `ignore`. 9/9 pass.
- Updated `test_hms_actions.py` to pin the new shapes (the two old paused-resume assertions now assert `ignore`, plus a new 16-char-fallback case).
- End-to-end: the `ignore` command itself is the one captured live and confirmed to clear the dialog + continue the print on the H2C.

### Note

Complements the merged #1840 frontend fix (which makes these buttons render in the first place) — together they give working "Ignore and Resume" from the Bambuddy UI on H2C print_error dialogs. I only have an H2C to test on; the routing for other models / the 16-char path would benefit from your eyes.


maziggy (changes requested):
Patch looks good — the 8-char-only branch with 16-char fallback is the right level of caution, and the live capture is convincing.

Two things to fold into this PR before merge:

1. While the sniffer is still attached, please also capture what Handy's "Problem Solved and Resume" button sends on the same plate-detection dialog. Probably the same `ignore` verb with a different `param` value (guessing `"resolved"`), but we want it on record. Document the captured value in a comment next to the `ignore` payload even if no second Bambuddy button is wired right now — future readers shouldn't have to re-derive it.

2. The comment block in `hms_resume` (lines 5469-5472) still claims the err-bearing variant is "silently rejected on print_error- and hms[]-sourced faults alike". That was accurate for `idle_ignore` per #1830 §2, but is no longer correct now that the dedicated `ignore` verb works. Please tighten that comment so a future reader doesn't conclude all err-bearing shapes are dead.

The `persistent` ("don't remind") flag being dropped on the PAUSE path is fine — a print_error dialog is one-shot, so there's nothing recurring to suppress.

Thanks for the live capture and the conservative scoping.

Boa-Thomas:
Both folded in. Thanks for the review.

**1. Captured "Problem Solved and Resume."** Sniffed it off the live H2C on the `0500_8062` plate dialog:

```json
{"print": {"command": "resume", "err": "83918946", "job_id": "0", "param": "reserve", "sequence_id": "..."}}
```

Slightly different from the guess — it's the **`resume`** verb (not `ignore`) with **`param: "reserve"`** (not `"resolved"`). So both dialog buttons send the *same* err-bearing shape (err = decimal `print_error` + `param:"reserve"` + `job_id`); only the verb differs: `ignore` for "Ignore and Resume", `resume` for "Problem Solved and Resume". Documented next to the `ignore` payload in `hms_ignore` and in the `RESUME_*` case. Per your note, I left the Problem-Solved button unwired — it still takes the plain-resume path — but its shape is now on record.

**2. Tightened the `hms_resume` comment.** It no longer claims err-bearing shapes are dead; it now scopes the §(2) "rejected" result to `idle_ignore` and points to the two captured (accepted) err-bearing verbs. One corollary worth flagging: since the live "Problem Solved" command is an err-bearing **`resume`** that the firmware accepts on the H2C, an err-bearing resume isn't universally rejected — whatever the H2D §(2) injection hit looks verb/encoding-specific. Left that as an observation rather than rewriting your H2D finding.

Persistent-flag-on-PAUSE: agreed, no change.

Boa-Thomas:
Follow-up data point while I had the dialog up: clicking the captured "Problem Solved and Resume" (`resume` + decimal `err` + `param:"reserve"`) in Handy on a **still-markerless** plate **re-raises** the error. So that `resume` re-checks the fault ("I fixed it, re-verify" semantics), whereas `ignore` is the actual bypass — which is why this PR wires the Ignore button.

Net: `PROBLEM_SOLVED_RESUME` staying on a re-checking resume is correct by design — wiring the err-bearing resume wouldn't get a genuinely-unsolved fault past the check. Tightened the code comment + CHANGELOG to say exactly that.

maziggy (changes requested):
Thanks for the live capture and the decimal-err finding — excellent evidence, the verb fix itself is in good shape.

Per our project rule we don't ship follow-ups: everything needed for merge goes into the same PR. Two items from your own #1830 trace would need to land in #1844 before it can merge:

1. ACK heuristic. You documented that len(hms_errors) deltas give false 200s on H2C because the sev-6 entries flicker on their own. The acked check needs to require a gcode_state transition out of PAUSE, or the targeted print_error/HMS code actually leaving the list — not just any change in count. Small edit in the execute-action route.

2. 16-char hms[]-array path. The PR keeps these on plain resume conservatively. Either capture device/<sn>/request for one of those (any live 16-char fault that exposes action buttons) and pin the correct verb, or leave the resume path with an inline comment stating exactly what a future capture would need to show to flip it. We don't want a half-fix where the print_error branch works but the hms[]-array branch silently doesn't.
If point 1 is out of scope for you, say so and I'll add it on top — but rolling it in is preferred so the PR is the single source of the fix. Point 2 is yours since you have the only working sniffing setup.

Boa-Thomas:
Reconciled this branch with `dev`: now that #1869 has landed it carries no remaining diff — #1869 covers all of it (the `ignore` verb + the decimal-`err` finding, the ack-detection fix, and the 8-/16-char paths), so this PR is fully superseded and fine to close. Thanks for folding the captures in.

One small piece of process feedback, offered with full respect that it's your call as maintainer: this is the second time (after #1840) that a PR I'd put real work into — including live `device/<sn>/request` captures off an actual H2C — was reimplemented as your own commit rather than merged or iterated on together. The findings clearly made it in, which is what matters most, and your implementations are clean. But from a contributor's side, having the PR itself land — even after requested changes, or with a `Co-authored-by:` trailer — is a big part of what makes the effort feel worth repeating.

If you'd be open to merging contributor PRs once they meet your bar (or pinging the author to adjust theirs to your style instead of rewriting from scratch), I'd gladly keep surfacing this kind of on-device evidence for the H2 series. Either way — glad the H2C ignore path is sorted. 🙏

maziggy:
I've midex things up due to gthe duplicate issues. Anyway.....lemme check.

maziggy:
So, what's the related issue now?

Boa-Thomas:
Related issue is #1830 (HMS actions not taking effect on the H2C — the original report). #1844 was the backend `ignore`-verb fix for it.

Heads-up on the mixup: the fix content is already in `dev` — `hms_ignore_command()` publishing `{command:"ignore", err:"<decimal>", param:"reserve"}` is at `bambu_mqtt.py` ~L5541 — so it isn't lost, even though the commit that carried it got rewritten out during the renumbering (`#1869` now points to the IDR-currency commit).

So #1844 is superseded by that dev content and fine to close. Thanks!

Issue #1861, opened 2026-06-30
[Feature]: Change auto drying thresholds


tchavei:
### Problem or Use Case

Currently it appears that the "fair" humidity threshold is used for both initiating and stopping the dryer function when enabled. If one has a "good" threshold of 15% and a "fair" threshold of 25%, the dryer function will enable once the "fair" threshold is passed but will stop after 30 min if the humidity reading falls below the  "fair" mark. This causes that, in practice, filaments stay around the "fair" threshold rather closer or at the "good" threshold.

### Proposed Solution

Ideally, the auto drying function should kick in once the "fair" threshold is passed (just like it works currently) but then it should use the "good" threshold value as target. If the user has, for example, set a 12h automatic drying cycle, this should be performed until a) the "good" threshold is reached or b) until the 12h expire. This would keep the filaments as close as possible to the "good" threshold.

### Feature Category

SpoolBuddy - AMS Management

### Priority

Would improve my workflow

### Contribution

- [x] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Please note our rules from CONTRIBUTING.md. Thanks!

maziggy:
https://github.com/maziggy/bambuddy/issues/1892

Issue #1605, opened 2026-06-02
[Feature]: AMS Humidity Trigger Per Filament


thenewguy:
### Problem or Use Case

It would be handy if the ams humidity & temperature setpoints could be set per filament type. 

### Proposed Solution

Currently the interface allows one group of setpoints. This is a problem when the printer is used to print multiple filaments with different temperature profiles - i.e. one AMS dedicated to PLA and another dedicated to Nylon. In this situation I would like to set one humidity range and temperature settings for Nylon and another for PLA management. I.e. I would like to try to keep the Nylon enclosure under 10% humidity and PLA is fine up around 20%. Another example is an AMS dedicated to ASA - this filament prints well while the AMS is drying. I would like to have the AMS heater running any time it is needed for ASA, but not for PLA.

### Feature Category

Print Queue & Scheduling

### Priority

Would improve my workflow

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Before we commit to building this, I'd like to gauge community interest. If you'd find this feature useful, please give this issue a thumbs up (👍) reaction so we can prioritize accordingly.

thenewguy:
> Before we commit to building this, I'd like to gauge community interest. If you'd find this feature useful, please give this issue a thumbs up (👍) reaction so we can prioritize accordingly.


Understood. I don't have any real way to advertise this so that people see it for likes. But it is the whole idea behind the X2D, H2D and H2C. Bambu sees multi material as important enough to build machines with two or more nozzles so I think that demonstrates the market need is there. I frequently print asa, pla, pvb, hips and nylon from dedicated ams to maintain humidity. Your solution to automate the ams humidity is a step above the other print farm software I've evaluated for Bambu printers but for it to be able to fully manage the farm for our engineering team this issue is necessary. Thanks!

maziggy:
Our monthly Sponsors voting is coming up. I'll add it as an option.

maziggy:
Will add it to the next Supporter voting.

maziggy:
Available/Fixed in branch dev and available with the next release or daily build. Please let me know if it works for you.

https://wiki.bambuddy.cool/features/ams/?h=humidity#per-filament-humidity-threshold

-----
If you find Bambuddy useful, please consider giving it a ⭐ on [GitHub](https://github.com/maziggy/bambuddy) — it helps others discover the project!

PR #1883, opened 2026-07-02
Update auto-drying stop threshold

tchavei:
## Description

Updates auto-drying so the fair humidity threshold still triggers drying, but an active drying cycle continues until the good humidity threshold is reached or the configured drying duration completes.

## Related Issue

https://github.com/maziggy/bambuddy/issues/1861

## Documentation

- [ ] Docs PR(s) linked above
- [X] No docs update required — reason:  README updated; wiki update pending maintainer guidance. I can create a wiki PR if requested.

## Type of Change

- [X] Documentation update (Readme)
- [ ] Code refactoring
- [ ] Performance improvement
- [X] Test addition or update

## Changes Made

- Load `ams_humidity_fair` as the auto-drying start threshold.
- Load `ams_humidity_good` as the humidity-based auto-stop threshold.
- Keep per-filament humidity overrides as trigger thresholds.
- Add regression tests for the fair-to-good drying.
- Update README auto-drying wording.

## Testing

- [X] I have tested with my printer model: H2D

## Checklist

- [X ] My code follows the project's coding style

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ Optional but recommended — drop the [`slicer-api/` Compose stack](slicer-api/R
 - AMS slot configuration (model-filtered presets, K profiles, color picker, pre-population for configured slots)
 - AMS info card (hover for serial number, firmware version) with custom friendly names that persist across printers
 - **AMS remote drying** — Start, monitor, and stop drying sessions for AMS 2 Pro and AMS-HT directly from the Printers page with filament-based temperature/duration presets, optional spool rotation; automatic PSU detection and HMS power error reporting. Rotate-spool toggle is disabled per-AMS when any tray has filament threaded into the feed tube (the AMS mechanism is locked there — rotating would jam the filament)
-- **Queue auto-drying** — Automatically dry filament between scheduled prints when humidity exceeds threshold; configurable presets per filament type, optional blocking mode
+- **Queue auto-drying** — Automatically dry filament between scheduled prints when humidity exceeds the fair threshold, then continue for the configured drying duration; configurable presets per filament type, optional blocking mode
 - **Ambient drying** — Automatically keep filament dry on idle printers based on humidity, regardless of whether prints are queued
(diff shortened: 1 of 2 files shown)


maziggy (changes requested):
Thanks for this — the fair-start / good-stop split is exactly the right model for #1861, and the tests are on point. Verified locally: ruff clean, the auto-drying suite passes, and since _check_auto_drying is the single path for queue / ambient / print-while-drying, the fix applies to all three modes. One thing to address before merge, plus a related edge case.

Should-fix: the stop threshold uses the global good only, but the start threshold honors the per-filament overrides from #1605. When a user sets a per-filament threshold below the global good — common for hygroscopic materials (PA, PVA, Nylon) — drying now stops wetter than the configured target, and can flap on/off every minimum-drying window.

Concrete example (reproduced against this branch): per-type PA = 30, global good = 40, humidity 35%. Before this PR, PA dried down to 30%; with the PR it stops at 35% (because stop = global good = 40), then restarts on the next cycle since 35 is still above the PA trigger of 30. So it under-dries exactly the filaments that most need it, and oscillates.

Fix is one line — clamp the stop so it is never wetter than the trigger:

effective_stop = min(stop_humidity_threshold, start_humidity_threshold)
and compare humidity <= effective_stop in the already-drying branch. In the normal case (good below the trigger) this is identical to what you have; it only changes the per-type-below-good case.

Bonus: that same clamp also covers a config gap. There is currently no validation that good <= fair (not in the Pydantic schema, the settings route, or the UI — the inputs only clamp 0-100). Pre-PR that was harmless for drying because good was UI-coloring only, but now good drives the stop. An inverted config (good=60, fair=40) would otherwise start at >40 and immediately satisfy <= 60, drying only the minimum window and flapping. With the min() clamp, effective_stop becomes min(60, 40) = 40 and it behaves sanely, so no separate validator is required for this PR.

Could you also add a regression test for the per-type-below-good case (start below the global good, assert it keeps drying past good down to the per-type target)?

tchavei:
Thank you for your constructive feedback. Indeed I didn't consider the specific scenario you described. I was more focused on the general behaviour change I had in mind. A typical case of tunnel vision. I appologise. I implemented your suggestions using a min function so the effective stop is the lowest threshold found. I'm currently testing locally.

maziggy:
@ryansouza your comments should go to the related issue and not here in the PR.

maziggy:

tchavei:

commit #1892 changed the drying behaviour to run the drying schedule fully regardless of humidity threshold. Should I close this?

maziggy:
Oh yes, I'm sorry.

PR #1900, opened 2026-07-05
[Feature]: Server-Side Slicing on linux/arm64 systems

### LINKED ISSUE #1821: [Feature]: Server-Side Slicing on linux/arm64 systems
### Problem or Use Case

Currently server-side slicing is said to not be working on linux/arm64 systems by the [wiki](https://wiki.bambuddy.cool/features/slicer-api/#platform-requirements). The reasons mentioned are non-existent AppImages (in the case of Bambu Studio) or problems with the build/launch pipeline and api (for OrcaSlicer) on arm64.

As a temporary fix, until either slicer works properly, I propose to circumvent the issue by emulating a linux/amd64 environment on the linux/arm64 host. This could be achieved by a number of emulation / vm solutions, ranging wildly in complexity.

The sidecars are already packaged in a docker container. This allows for a lightweight emulation environment by leveraging docker cross-architecture emulation.

This would fix the previously closed issue #1328.

### Proposed Solution

I propose to add the `platform: linux/amd64` line to both services in [slicer-api/docker-compose.yml](https://github.com/maziggy/bambuddy/blob/main/slicer-api/docker-compose.yml). This will have no effect on linux/amd64 hosts, where the slicers are already working. Only on linux/arm64 systems this will trigger the emulation.

For the emulation to work, qemu (the actual emulator) and binfmt (kernel feature triggering the emulation for non-native binaries) need to be installed on the host. In my opinion, the need to install these packages is unfortunate but acceptable in the otherwise docker-only bambuddy environment, as it may be the only way to get server-side slicing working as of right now.

For debian based systems the install can be achieved by the following single command:
`apt-get install qemu-user-static binfmt-support`
No additional configuration is necessary. Similar commands should be available for other flavours of Linux.

The emulation will cause a performance penalty, the significance of which will differ from system to system. On my test machine (8 core, Rockchip RK3588 SoC) the slice time even for large complex models was fully acceptable (less than 2 minutes).

I believe this is an optimal solution/workaround for the current situation, as it requires only a minimal amount of effort both on the end-user and developer side.

### Alternatives Considered

Alternative emulation techniques were not tested, but would likely require significantly more setup effort for the end-user.

### Feature Category

Other

### Priority

Would improve my workflow

### Mockups or Examples

_No response_

### Contribution

### Checklist

Felix-Rm:
## Description

This PR adds the `platform: linux/amd64` option to both slicer services. This forces docker to always pull/build/run the amd64 image. On hosts with a different platform this will activate execution via an emulation layer. This has the effect of making server-side slicing available on linux/arm64 systems without the need for working arm64 slicer builds.  

## Related Issue

Fixes #1821 

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#52

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- add `platform: linux/amd64` to `orca-slicer-api` and `bambu-studio-api` services

## Screenshots

(none)

## Testing

I have tested the changes on a linux/amd64 machine to confirm no change in behavior on previously working platforms. Additionally, tests on a linux/arm64 (Debian 12 on Rockchip RK3588 SoC) confirmed the server-side slicing to be working on non amd64 platforms.

- [x] I have tested this on my local machine (linux/amd64)
- [x] I have tested this on a secondary machine (linux/arm64)

## Checklist

## Additional Notes

(none)


maziggy (changes requested):
Thanks Felix — really appreciate you testing this on actual arm64 hardware, not just asserting it should work. The change itself is clean and it's genuinely additive on amd64, so no concerns there.

Before I merge it into the default compose file, I want to get two things right, because that file carries a "works out of the box on QNAP/Synology/SBC" promise and I don't want to quietly break it:

1. Performance numbers. Emulated slicing is the part I can't set expectations for yet. Could you share rough wall-clock slice times on the RK3588 for a typical model — ideally the same model sliced native on an amd64 box vs. emulated on the arm64 box? Even ballpark (e.g. "a 2-plate PLA model: ~20s amd64 vs ~90s emulated") tells me whether this is a "slightly slower, totally usable" feature or a "here be dragons" one, and that changes how we present it.

2. The binfmt prerequisite. platform: linux/amd64 on its own doesn't provide the emulation layer — the host needs QEMU binfmt_misc registered (qemu-user-static / tonistiigi/binfmt), or the container fails with "exec format error." Your RK3588 presumably had it set up. For everyone else pulling the default file, that would be a confusing first-run failure. So this needs to be documented as a required step.

Given both of those, my inclination is to land this as an explicitly experimental arm64 path rather than in the default services — e.g. a docker-compose.arm64.yml override (or an env-gated variant) that arm64 users opt into, plus the wiki instructions with the binfmt command and a clear performance note. That keeps the default file amd64-clean and out-of-box, and still gives arm64 folks exactly the recipe you proved works. If the slice times come back reasonable (say within ~2x native), I'm happy to be more generous about how prominent we make it.

maziggy:
?

Felix-Rm:
Regarding Point 1:

I have benchmarked the slice times on the arm64 host with a very simple and a more complex model. The time reported for the server side slicing is the time shown in the progress popup on the Bambuddy dashboard when the status changed to "All done, Success". Please note that this used the BambuStudio sidecar. As a comparison benchmark, the native arm64 Flatpack build of OrcaSlicer was used, run via the CLI interface.

Both Slicers were instructed to Slice for the BambuLab P1S printer with 0.4mm nozzle diameter and 0.12mm (fine) layer height, using BambuLab PLA Matte as the target filament.

This benchmark is meant as reference / order of magnitude only. System load during the benchmark runs was not controlled for (but should be similar). Also the mismatch in slicers (BambuStudio vs OrcaSlicer), while I understand they share a large part of the source code, may have an impact on slicing times.

The results averaged over 3 runs each are as follows:

Simple Model: [Calibration Cube](https://www.thingiverse.com/thing:3130609):

- Native Flatpack of OrcaSlicer: ~3s
- Server Side Slicing via BambuStudio: ~10s (about 3.3 times native)

Medium Model: [Stanford Bunny](https://www.thingiverse.com/thing:3731)

- Native Flatpack of OrcaSlicer: ~15s
- Server Side Slicing via BambuStudio: ~95s (about 6.3 times native)

I agree that this should not be marketed as a fix for arm64 systems but as a experimental/temporary workaround while the arm64 builds are unavailable or broken. The performance hit is quite large in relative terms and expected to increase for even more complex models. So the usability is strongly dependent on the user / usage scale.

Personally I think the slice times are still reasonable, even if they would go up to like 5 minutes for multi plate / very complex models (again, probably highly dependent on the actual system). The effort in having a second system always running for the amd64 sidecars or manually downloading and slicing the model on a desktop system does outweigh having to wait a few minutes, at least for me.

Regarding Point 2:

The `qemu-user-static` and `binfmt-support` dependencies are not preinstalled on most systems. A small paragraph explaining the issue as well as the necessary steps to install the needed packages was added to the relevant wiki page in the accompanying [Wiki PR](https://github.com/maziggy/bambuddy-wiki/pull/52).

Let me know if this issue needs highlighting elsewere.

maziggy (changes requested):
Thanks Felix — exactly the data I needed, and much appreciated that you ran it on real hardware with a native comparison.

The numbers confirm my gut feeling though: 3.3-6.3x native and scaling worse with complexity puts this firmly in "experimental workaround" territory (which you say yourself), and an RK3588 is about as fast as ARM SBCs get — a typical ARM NAS would fare considerably worse, and many of those can't even install qemu-user-static, so they'd hit a confusing runtime "exec format error" instead of today's clear pull-time failure.

There's also a forward-compatibility trap: once working arm64 sidecar builds exist (the upstream CLI fix is in progress), a hardcoded platform pin in the default file would silently keep every arm64 user on emulation instead of the native image.

So I'd like to land this as the opt-in variant: please move the two lines into a separate slicer-api/docker-compose.arm64.yml override, so arm64 users run with -f docker-compose.yml -f docker-compose.arm64.yml. Keep the wiki PR, ideally with an explicit note that this is experimental, expect roughly 3-6x slower slicing, and it's a stopgap until native arm64 builds ship.

With that restructure I'm happy to merge both. Thanks again for the thorough testing!

maziggy:
?

Felix-Rm:
First of all: Sorry it took me so long to respond. It has been a busy couple of weeks for me.

I have moved the changes into a seperate override compose file which can be applied in the way you suggested. The wiki PR has been adjusted to highlight this as an experimental feature and show the slight difference in commands used to start the services. I also added some information to the top of the new compose file as well as the README.

Let me know if the wiki and README changes are clear or if they need some further modification.

PS: I rebased both this PR and the wiki PR on top of the base branch changes. They hopefully should merge cleanly now.

maziggy (approved):

PR #2541, opened 2026-07-11
refactor(auth): Make Administrators-group membership authoritative

### LINKED ISSUE #2537: [Bug]: Duplication of Admin / Administrator badge on user accounts with admin permissions
### Product

Bambuddy

### Area

Authentication / users / permissions

### Bug Description

Image

Purple Admin badge is duplicated

### Expected Behavior

Single badge to represent the admin state.
"Admin" preferred for brevity

### Steps to Reproduce

Settings -> Authentication

N/A

### Printer Model

None

### Bambuddy Version

2.4.9

### SpoolBuddy Version

_No response_

### Printer Firmware Version

_No response_

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

N/A

### Screenshots

_No response_

### Additional Context

_No response_

### Final checks

EdwardChamberlain:
## Description

This PR fixes #2537 and makes membership in the canonical `Administrators` group the sole source of administrative authority.

Also removes the system added "Admin" pill and keeps the group name "Administrator" group name pill for UX consistency.

## Related Issue

Fixes #2537

## Documentation

**Pick one**:
- [ ] Docs PR(s) linked above
- [ ] No docs update required — reason: No change to users

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- The `role` field is removed from `POST`/`PATCH /api/v1/users` request bodies. Clients must use `group_ids` to assign Administrators membership; legacy `role` payloads return HTTP 422.
- On first startup after upgrade, existing users with `role="admin"` are migrated into the Administrators group and the legacy field is normalized. This migration is recorded with a startup marker and is not repeated.
- The duplicate Admin pill is removed from the Administrators group display.

## Screenshots
image

## Testing

- [ ] I have tested with my printer model: N/A

## Checklist

## Additional Notes
- Static assets not regenerated


maziggy (changes requested):
Thanks for picking this up — the duplicate is real, but the current approach removes the badge rather than deduplicating it, and that loses information we need.

is_admin is true when a user has role == "admin" OR is in the Administrators group — the two are independent (backend/app/models/user.py:87). For an admin who has the role but not the group, the Admin pill is the only place the UI says so, since the users table has no Role column. With this change they render as an ordinary user, and in UsersPage the empty-groups fallback also lost its && !user.is_admin guard, so they now get labelled "No groups" — the opposite of the truth about someone holding every permission.

That state is reachable from the UI: the setup admin is created with role="admin" plus the Administrators group. Remove them from Administrators in the edit dialog (allowed unless they are the last admin, and the UI never writes role) and they keep full admin while the badge disappears. Users created via POST /api/v1/users with role="admin" and no group_ids land in the same state.

Could you keep the pill but suppress it when the Administrators group pill already covers it?

{user.is_admin && !user.groups?.some(g => g.name === 'Administrators') && (
  <span className="...">{t('settings.admin')}</span>
)}

System groups cannot be renamed (groups.py:187), so matching on the name is stable. That gives exactly one pill for the user in your screenshot, while keeping the badge where it is the only signal. The && !user.is_admin guard on the "No groups" text should then stay. Three places: SettingsPage.tsx:5674, SettingsPage.tsx:5738, UsersPage.tsx:342.

Could you also add a test for a user with role: 'admin', is_admin: true, groups: [] — it must still show "Admin" and must not show "No groups"? The current test passes with or without the change, and UsersPage.tsx has no coverage.

EdwardChamberlain:
This feels closely coupled to the other PR, and I think the duplicate badge is exposing a more fundamental issue: there are currently two independent sources of administrative authority.

A user can be an administrator through role == "admin" or through membership of the Administrators group, which means the system has to infer and deduplicate two representations of the same effective state.

I think we should move towards a single canonical authorisation path. Existing users with role == "admin" should be migrated into the group-based permission model, with administrative authority derived from membership of a single protected administrative group identified by a stable internal flag rather than by its display name.

The proposed frontend check is a sensible compatibility fix for the immediate issue, but matching "Administrators" in several UI locations would introduce further reliance on a magic group name. I would prefer to treat that as temporary and track consolidation of the underlying admin model as the proper fix.

EdwardChamberlain:
Follow-up pushed to address the underlying split-admin model rather than add a conditional UI deduplication.

This changes administrative authority for users to a single canonical source: membership of the protected Administrators group and retires the legacy role approach.

- Existing role="admin" users are migrated once at startup into Administrators, then their legacy role is neutralised to user. A migration marker prevents a later restart from re-adding someone who has subsequently been deliberately removed from the group.
- User.is_admin, admin route gates, setup, and last-admin protections now all use Administrators membership.
- User create/update no longer write a meaningful role; role remains only as an inert response compatibility field.
- The frontend now renders the canonical group pill only, so the duplicate cannot occur and a genuine administrator cannot become visually inconsistent.
- API-key /auth/me identities no longer present themselves as group-admin users.

This removes the duplicated representation that caused the badge problem, rather than preserving two independent admin models and hiding one of their UI signals. Coverage now includes legacy-user migration, no re-promotion after deliberate group removal, and both Settings and Users-page canonical badge rendering.

maziggy (changes requested):
Thanks for this, Ed — the underlying refactor is genuinely well done. Collapsing admin authority onto Administrators-group membership is a clean direction, the startup migration is properly guarded (idempotent via the Settings marker, ordering is correct, dialect-aware upsert), and the last-admin guards cover deactivate/delete/group-removal on every path. Nice work.

The thing I want to sort out before merge is scope and disclosure. As it stands this PR is titled and described as the cosmetic "remove duplicate Admin pill" fix (#2537), with the "No docs update required — no change to users" box checked, but it actually rewrites the authorization model, removes a field from the user API, and mutates existing databases on upgrade. That's a much bigger, security-relevant change and it should be merged as what it is. I'm fine keeping it as one PR — just need the following:

1. Silent breaking API change. UserCreate/UserUpdate no longer accept role, and the schema has no extra policy, so Pydantic silently ignores it. An existing script that POSTs {"username","password","role":"admin"} will now quietly create a non-admin user — no error, no group. Please make this an explicit failure (extra="forbid" -> 422) rather than a silent privilege downgrade, so existing integrations get a clear signal instead of a surprise.
2. Accurate title + description. Retitle to reflect the auth-model change, and describe the two user-visible effects: the role field is removed from the user API (admin is now conferred only via Administrators-group membership / group_ids), and existing role=admin users are migrated to the Administrators group on first startup after upgrade.
3. Docs. Uncheck "no docs required" and add a short note covering the API role removal and the upgrade migration behavior (wiki + anywhere the user API is documented).

Minor:
- AuthContext now recomputes isAdmin from groups even though the backend already sends is_admin computed the same way — harmless, but two sources of one truth.
- The legacy if user.role == "admin" migration in database.py is now largely superseded by the new one — worth a cleanup or a comment.

maziggy:
?

maziggy (changes requested):
A few things to resolve before I merge.

extra="forbid" is broader than the problem. You want legacy role payloads rejected; forbidding all extras means any unrecognised key from any client fails, and the caller gets a bare Pydantic extra_forbidden with no hint that group_ids is the replacement. Anyone scripting POST /api/v1/users breaks with an unhelpful message. Either leave the schemas permissive and just drop role (extras were ignored before), or declare role explicitly and raise a 400 that names group_ids.

UserResponse.role now always lies. It is documented as kept for backward compatibility, but after migration every user including administrators has role="user", so it returns "user" for everyone permanently. A legacy consumer reading it to detect admins silently sees zero admins - worse than the field being gone. Either drop it from the response, or keep it honest with role="admin" if user.is_admin else "user" in the three _user_to_response helpers (users.py, auth.py, mfa.py).

The AuthContext change should be reverted. user.groups.some(g => g.name === 'Administrators') duplicates what the backend already computes and ships as is_admin on the same payload. The only case where the old user?.is_admin differed was API-key identities, and you already fixed that at the source by setting is_admin=False there. So it buys nothing and adds a second place to update.

require_admin() is unreachable. It is only called from RequireAdmin(), which has no call sites anywhere in backend/. require_role is dead too. The live gate is require_admin_if_auth_enabled. Worth deleting all three rather than leaving a new unused factory next to two dead ones.

On your own framing from the first comment - a stable internal flag rather than a display name - what shipped keys admin on the literal string "Administrators" in fifteen backend sites and adds a sixteenth in the frontend. Renaming is blocked for system groups so it cannot drift at runtime, but this is the design goal you named, and an is_admin_group flag on Group would let User.is_admin stop string-matching entirely. Happy for that to be a follow-up, but say so in the PR body rather than leaving it implicit.

Smaller items: users.py and groups.py both define _active_admin_user_ids with the same name and different signatures, and the groups.py one re-checks a name it is already inside a branch for. _ensure_administrators_group re-declares a group DEFAULT_GROUPS already defines and has already drifted on the description string. settings.admin and users.admin are now unreferenced but still present in all thirteen locales. And two tests lost their point: test_users_create_holder_cannot_create_admin now gets its 403 from the route-level admin gate before group_ids is ever parsed, so the escalate-via-group-assignment vector it was written to guard is no longer covered.

One thing outside this PR that is worth knowing: _sync_ldap_user replaces the LDAP-managed slice of a user's groups on every login with no last-admin check. Not a regression from your change - LDAP users were always provisioned role="user" and never had the backstop - but with role retired, group membership is now the sole authority everywhere, and that is the one remaining path that can empty the admin set without hitting any of your guards. I will handle that separately.

PR #2542, opened 2026-07-11
Make default operator and viewer groups editable

### LINKED ISSUE #2538: [Bug]: System created permission groups are not user editable
### Product

Bambuddy

### Area

Authentication / users / permissions

### Bug Description

System user groups Operators and Viewers are not user editable. 

### Expected Behavior

User should be able to configure / delete these groups as required. 

Administrator should remain protected. 

### Steps to Reproduce

Settings -> Authentication

Attempted group change

### Printer Model

None

### Bambuddy Version

0.2.5

### SpoolBuddy Version

_No response_

### Printer Firmware Version

_No response_

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

N/A

### Screenshots

_No response_

### Additional Context

_No response_

### Final checks

EdwardChamberlain:
## Description

Remove system group flag from all default auth groups other than Administrator to give more user flexibility.

## Related Issue

Fixes #2538

## Documentation

**Pick one**:
- [ ] Docs PR(s) linked above
- [ ] No docs update required — reason: N/A

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Removed flag for admin groups
- Added database migration for existing installs
- Added tests

## Screenshots
image

## Testing

- [ ] I have tested with my printer model: N/A

## Checklist


maziggy (changes requested):
Thanks for digging into this. The bug is real, but I'd like the fix to go the other way — Operators and Viewers should become editable, while staying undeletable and unrenameable. So please keep is_system set on all three groups.

The reason: is_system is what blocks deletion (groups.py:249) and renaming (groups.py:188), and I want to keep both blocks. The names are load-bearing in about six places on the backend — the startup backfills, the legacy user migration, and OIDC auto-provisioning, which falls back to a group literally named "Viewers" (mfa.py:2009). Deleting or renaming them silently detaches all of that.

The actual bug is narrower than it looks. The UI already allows this: GroupEditPage only disables the name field for system groups (:193), the warning banner even says "System group names cannot be changed", and the permission checkboxes are live. Tick a permission on Operators and hit Save, and it's the API that rejects it — "Cannot modify permissions of system groups" (groups.py:204). That rejection is #2538.

So the change I'd like to see:

1. Revert the is_system flips in permissions.py, the group.py docstring, and the handlers.ts mock. Nothing becomes deletable or renameable.
2. In groups.py:204, reject the permissions edit on group.name == "Administrators" instead of on is_system. Administrators keeps its fixed ALL_PERMISSIONS set — stripping it is a self-inflicted DoS, and the startup sync would re-add it anyway. Operators and Viewers become editable.
3. Make the edit survive a restart. This part is required either way, and it's the one thing the current PR misses. seed_default_groups() re-grants Operators/Viewers permissions by name on every boot: the OWN tier plus orca_cloud:auth (database.py:3698), inventory forecast (:3717), pipelines (:3744). I measured it — set Operators to ["printers:read"], call seed_default_groups() once, and you get back:

['archives:read_own', 'library:read_own', 'orca_cloud:auth', 'pipelines:read',
 'pipelines:run', 'pipelines:write', 'printers:read', 'queue:read_own']

Seven permissions re-granted with no warning. So an admin revoking pipelines:run from Operators would see it come back after the next restart, which is the worst failure mode a permissions feature can have. Those backfills are upgrade repairs and should run once, not on every boot. An is_customized flag on Group (default false, set by the PATCH route when permissions change, skipped by the backfills and by the DEFAULT_GROUPS permission migration) would do it. That needs a startup migration to add the column — please make it dialect-safe and verify on both SQLite and PostgreSQL.

4. Tests: your Viewers test has exactly the right instinct in calling seed_default_groups() after the change — please apply that to the permissions test, since test_default_operator_group_permissions_can_be_updated currently can't see the revert. Plus: Administrators' permissions still rejected, and delete and rename still rejected for all three.

EdwardChamberlain:
Thanks, that clarifies the current dependencies, but I think the reliance on the literal names Operators and Viewers is the fundamental issue we should address.

Authorisation should be based on permissions, not group names. Where the backend needs a specific group for provisioning, such as the OIDC fallback, it should reference a configured group ID or stable internal identifier rather than a display name.

With that model, the default groups can remain initial templates without being permanently forced on every installation. They could be renamed, deleted, or replaced, with deletion only blocked where a group is actively referenced by configuration.

I’m concerned that adding is_customized would preserve the name coupling and introduce another special group state around it. I would prefer to broaden the fix by removing the name-based assumptions, converting the startup backfills into one-time migrations, and enforcing administrative safety through permissions rather than a magic group name.

maziggy (changes requested):
Nice one, Ed — this is clean and correctly scoped. It does exactly what #2538 asks and I couldn't find anything it breaks.

Two things before merge:

1. Docs. The "No docs update required" box is checked, but this is user-visible: Operators and Viewers are now editable and deletable rather than permanent. Please uncheck it and add a short wiki note that only Administrators is a permanent system group and the others are editable defaults.
2. Heads-up on overlap with #2541 — both touch seed_default_groups in the same region, so whichever merges second will need a rebase.

On your design comment: I agree that keying provisioning and backfills on the literal names Operators/Viewers is the deeper issue, and moving to permission-based checks / stable group IDs plus one-time migrations is the right long-term shape. But it's a larger change and this PR doesn't make the coupling any worse, so I'd rather land this as-is and track the name-decoupling as its own issue. Could you open one so we don't lose it?

maziggy:
?

EdwardChamberlain:
Suggest we merge the other PR first and then I can rebase this one.

Issue #1152, opened 2026-04-28
[Feature]: Allow using the name of the file instead of using metadata


smandon:
### Problem or Use Case

When sending a job to the virtual printer, the name used for the new archive is apparently taken from some metadata in the file. I couldn't find a way in bambu studio to edit this name and it is whatever the original file creator set.
It would be great to be able to use the same name as the name of the uploaded file by default, as this name can be edited in bambu studio when sending.

I guess when sending directly to the printer the name is also taken from metadata, maybe the same logic could be applied?

### Proposed Solution

Add an option to use the name of uploaded files to the virtual printer for new archives instead of relying only on metadata.

### Alternatives Considered

Another option would be to at least allow changing the name when reviewing new archived items. Currently we have to review, archive then edit.

### Feature Category

Print Archiving

### Priority

Would improve my workflow

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Available/Fixed in branch dev and available with the next release or daily build. Please let me know if it works for you.

-----
If you find Bambuddy useful, please consider giving it a ⭐ on [GitHub](https://github.com/maziggy/bambuddy) — it helps others discover the project!

smandon:
I checked with the latest daily image.

If the 3mf file has a title set in the `3D/3dmodel.model` metadata file then the name shown in the review card correctly uses the file name but when archived the card still shows the name from metadata.

<img width="1476" height="217" alt="Image" src="https://github.com/user-attachments/assets/8355761d-882e-4043-9c12-ef528ec892c8" />

<img width="351" height="533" alt="Image" src="https://github.com/user-attachments/assets/d510a71f-e703-4ce5-a0c1-d6705c3592e1" />


If the title is empty in the `3D/3dmodel.model` file, then it always uses the file name in both the review and archive cards, but I guess it was already the case before the change. The name still has ".gcode" at the end, only the ".3mf" was removed from the file name.

<img width="351" height="533" alt="Image" src="https://github.com/user-attachments/assets/a9868e09-1a83-4fe6-89a0-66139dbb4771" />

maziggy:
Thanks for the careful retest, @smandon — both observations land. Pushed the fix to `dev`:

[...]

PR #2610, opened 2026-07-19
Expose prefer_filename_for_name on manual archive upload endpoint

### LINKED ISSUE #2609: [Feature]: Expose prefer_filename_for_name on POST /api/v1/archives/upload
### Problem or Use Case

`ArchiveService.archive_print()` already supports a `prefer_filename_for_name` flag (added for #1152) that makes the archive's display name follow the uploaded filename instead of the `print_name` baked into the 3MF's embedded metadata. It's wired up for the FTP pending-uploads review flow and the virtual-printer dispatch flow, but **not** for `POST /api/v1/archives/upload` — the manual/API upload endpoint.

External integrations that upload 3MF files on a user's behalf (e.g. an inventory/print-queue tool that uploads a file named after the tracked job) currently have no way to make the archive's name match the filename they sent. The archive list shows whatever internal project title the 3MF happened to embed, which can be a stale/generic name from whoever originally sliced the file — confusing when the uploader intentionally renamed it before sending.

### Proposed Solution

Add a `prefer_filename_for_name: bool = False` query parameter to `POST /api/v1/archives/upload`, forwarded straight through to `ArchiveService.archive_print(..., prefer_filename_for_name=prefer_filename_for_name)` — the same pattern already used by the other two callers. Default `False` preserves existing behavior for every current caller.

### Alternatives Considered

- Rewriting the 3MF's internal metadata (`Metadata/model_settings.config` / equivalent) client-side before upload — rejected as needlessly heavy; the flag already exists server-side and just isn't exposed on this one route.
- Client-side-only fix (prefer `filename` over `print_name` when *displaying* an archive) — worth doing too, but doesn't fix search/sort/anything else in Bambuddy that reads `print_name` directly from the DB.

### Feature Category

API

### Priority

Would improve my workflow

### Contribution

### Checklist

Person2099:
## Description

`POST /api/v1/archives/upload` now accepts an optional `prefer_filename_for_name` query param and forwards it to `ArchiveService.archive_print`. When `true`, the archive's display name uses the uploaded filename stem instead of the `print_name` embedded in the 3MF's metadata - same behavior the FTP pending-uploads review flow and virtual-printer dispatch already have (#1152), just not exposed on this route. Default is `false`, so no behavior changes for existing callers.

## Related Issue

Fixes #2609

## Documentation

**Pick one**:
- [x] No docs update required - reason: adds an optional, default-off query param to an existing endpoint; no new config key, no default-behavior change, no UI change.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Added `prefer_filename_for_name: bool = False` param to `POST /api/v1/archives/upload` in `backend/app/api/routes/archives.py`, forwarded to `ArchiveService.archive_print`.
- Added integration tests in `backend/tests/integration/test_archives_api.py` covering both `true`/`false` values and the omitted-param default.

## Screenshots

N/A - API-only change, no frontend/UI touched.

## Testing

- [ ] I have tested with my printer model: N/A, this path doesn't touch printer communication, only 3MF upload/archival

Ran `ruff check`, `ruff format --check`, and the new tests directly (`pytest backend/tests/integration/test_archives_api.py -k upload_archive`) - 3 passed. Also ran the full backend suite (`pytest backend/tests/`): 3439 passed, 1 pre-existing failure in `test_plate_thumbnail.py` (FTP/kqueue `OSError: Bad file descriptor`, unrelated to this change - not touching FTP or thumbnails).

## Checklist

## Additional Notes

Marking as Draft per CONTRIBUTING.md's "no assigned issue = no PR" - opening this now so the branch/diff is visible alongside the issue, will mark ready for review once a maintainer assigns #2609.


maziggy (changes requested):
Nice, clean PR — thanks Seb. I confirmed the parity: archive_print already takes prefer_filename_for_name and uses it exactly as you describe (filename stem wins over the embedded print_name), and the FTP review flow and VP dispatch already pass it, so exposing it here is consistent. Default-off means no change for existing callers, the permission gate and path-traversal/validation logic are untouched, and the tests are well-constructed — parametrized true/false plus the omitted default, and returning a real PrintArchive so the response serialization actually runs. Good stuff.

One thing before merge: the upload-bulk endpoint right below (upload_archives_bulk, also calls ArchiveService.archive_print) doesn't expose this flag, so the two upload routes would diverge — a filename-vs-print_name preference that works on single upload but silently can't be set on bulk. Since it's the same one-line forward, could you add the same prefer_filename_for_name query param to upload-bulk in this PR (applied to every file in the batch), plus a test? I'd rather keep the two upload paths consistent than land them split.

Person2099:
All done. Added `prefer_filename_for_name: bool = False` to `upload_archives_bulk`, forwarded to `archive_print` for every file in the batch, plus two tests (parametrised true/false forwarding, and the omitted-default-false case) mirroring the single-upload tests. Both of the routes are consistent now.

Ran `ruff check`/`ruff format --check` and the archive upload tests (`pytest backend/tests/integration/test_archives_api.py -k upload_archive`) - 6 passed.

PR #2623, opened 2026-07-21
Add Intel QSV camera video processing

### LINKED ISSUE #2621: [Feature]: Intel Quick Sync acceleration for built-in RTSP cameras
## Summary

Add optional Intel Quick Sync Video acceleration for built-in printer cameras that use RTSP.

Bambuddy currently decodes the printer H.264 stream and encodes the browser-facing MJPEG stream entirely in software. On low-power systems this creates a noticeable continuous CPU load while the camera is open.

I implemented and tested a working prototype that adds a video-processing selector to the Camera settings:

- Software processing
- Intel Quick Sync

The selected mode is stored as a normal global Bambuddy setting and is applied when the shared RTSP camera stream is created.

## Current software pipeline

The current pipeline is:

H.264 software decode -> CPU frames -> MJPEG software encode

On an Intel N100 host, with an H2D camera at 1680x1080 and a 15 FPS Bambuddy stream, FFmpeg used an average of 47.90% CPU.

The measurement was taken with:

pidstat -p <ffmpeg-pid> 1 30

## Intel Quick Sync pipeline

The prototype uses FFmpeg QSV for both decoding and encoding:

h264_qsv -> QSV video surface -> mjpeg_qsv

The relevant FFmpeg arguments are:

-init_hw_device vaapi=va:/dev/dri/renderD128
-init_hw_device qsv=qs@va
-filter_hw_device qs
-hwaccel qsv
-hwaccel_device qs
-hwaccel_output_format qsv
-c:v mjpeg_qsv
-global_quality 80

Decoded frames remain in GPU video memory and are passed directly to the MJPEG QSV encoder without hwdownload, software scaling, or a later hwupload.

With the same printer, resolution and 15 FPS output, average FFmpeg CPU usage dropped to 5.66%.

This is approximately:

- 8.5x lower CPU usage
- 88% reduction in CPU load

## Tested environment

- Printer: Bambu Lab H2D
- Camera input: H.264 High, 1680x1080
- Host CPU: Intel N100
- Container: Debian-based LXC
- FFmpeg: Debian Trixie package
- VAAPI driver: Intel iHD
- QSV runtime: oneVPL / libmfx-gen
- GPU device: /dev/dri/renderD128

The resulting MJPEG stream works with Bambuddy existing stream parser, fan-out broadcaster and browser viewer.

## Working prototype

A working prototype has been created in a branch based on the current dev branch.

It currently includes:

- a new global camera_video_processing setting
- allowed values: software and intel_qsv
- software as the default
- validation in AppSettingsUpdate
- frontend API typing
- a selector in Settings -> General -> Camera
- persistence through the existing /settings API
- loading the setting when the built-in camera stream is created
- QSV-specific FFmpeg command construction
- the existing software pipeline remaining unchanged
- preservation of the RTSP TLS proxy, timeout handling, reconnect logic and MJPEG fan-out
- no database migration, because application settings already use key/value storage

The UI currently exposes:

Video Processing
- Software processing
- Intel Quick Sync

## Tests performed

The existing tests remain green:

- 45 camera profile and FFmpeg lifecycle tests passed
- 128 camera and FFmpeg-related unit tests passed
- 68 settings-related unit tests passed
- TypeScript type checking passed
- Python compilation checks passed

The prototype was also tested with the real Bambuddy service and a live H2D camera stream.

## Proposed production behaviour

Before submitting the implementation as a PR, I propose adding:

1. Intel QSV capability detection:
   - /dev/dri/renderD128 exists
   - the Bambuddy service user has access
   - FFmpeg exposes h264_qsv and mjpeg_qsv
   - a real QSV session can be initialized

2. Automatic software fallback:
   - if QSV fails during startup, log the failure
   - restart the stream with software processing
   - avoid leaving the viewer with a blank stream

3. A status message in the UI:
   - Intel Quick Sync available
   - GPU device unavailable
   - insufficient device permissions
   - Intel media runtime unavailable
   - required FFmpeg QSV codec unavailable

4. Unit tests for both generated FFmpeg command variants.

## Compatibility

The feature is optional.

Existing installations continue using software processing by default. Systems without Intel graphics or without the required runtime are not affected.

Intel libraries do not need to become mandatory Bambuddy dependencies unless the user explicitly enables Intel Quick Sync.

The same command-building abstraction could later be extended to generic VAAPI or NVIDIA NVENC.

I can prepare a PR with the current prototype and complete capability detection, software fallback and tests based on maintainer feedback.

pterodaktil02:
## Summary

Adds optional Intel Quick Sync acceleration for built-in RTSP camera processing.

Software processing remains the default. When Intel QSV is selected, the camera pipeline uses hardware H.264 decoding and MJPEG encoding.

## Implementation

- Added global camera video processing setting:
  - Software
  - Intel Quick Sync
- Added QSV FFmpeg pipeline:
  - `h264_qsv`
  - QSV video surfaces
  - `mjpeg_qsv`
- Added settings UI and locale strings
- Added Intel Quick Sync to the technical-name allowlist
- Kept existing software processing behavior unchanged by default

## Performance

Tested on Intel N100:

- Software FFmpeg CPU usage: about 47.9%
- Intel QSV FFmpeg CPU usage: about 5.7%
- Reduction: roughly 8.5x

## Translations

The English and Russian strings were written manually.

Translations for all other supported languages were generated automatically and should be reviewed and validated by native speakers before they are considered final.

## Testing

- Frontend: 2386 tests passed
- Backend: 7073 tests passed
- One unrelated library reslicing integration test failed during highly parallel execution
- The failing test was then run separately six times and passed all six runs

## Docs
https://github.com/maziggy/bambuddy-wiki/pull/59

Closes #2621


pterodaktil02:
I have added an initial QSV capability diagnostic to the Settings page.

Could you please review the overall approach and share any comments, suggestions, objections, or preferences?

In particular, I would like feedback on:

- the separate global diagnostic endpoint;
- the selected diagnostic stages;
- the inline Settings UI;
- the response format and failure codes.

The UI currently uses English fallback strings. I intentionally left the new locale keys out for now, so the wording and UX can be adjusted without updating every translation first.

maziggy (changes requested):
Thanks for this — the QSV pipeline is a real win (8.5x CPU reduction is significant), and I appreciate the honest disclosure on the machine-translated locales and the diagnostic panel as a pre-flight check. I reviewed the whole thing and it's structurally sound: no command injection (everything goes through create_subprocess_exec arg lists with hardcoded flags), the setting is validated to software/intel_qsv and falls back safely on anything unexpected, and default-software means no change for existing installs.

A few things to address before merge:

1. The diagnostic tests a different QSV init than the actual pipeline. Stage 4 in qsv_diagnostic.py uses -init_hw_device qsv=hw:/dev/dri/renderD128, but the camera pipeline uses -init_hw_device vaapi=va:/dev/dri/renderD128 -init_hw_device qsv=qs@va -filter_hw_device qs. The comment says it initializes "the same QSV device used by the camera pipeline," but it doesn't - so a passing diagnostic doesn't actually guarantee the stream works, and a failing one could be a false negative. Please make the diagnostic use the same VAAPI-derived QSV init the pipeline uses.

2. The qsv-diagnostic endpoint spawns ~4 ffmpeg processes (up to ~40s) but is gated only on SETTINGS_READ, which Viewers hold, with no rate limit - a low-privilege user can trigger repeated process spawns. Please gate it on SETTINGS_UPDATE (same as other admin-style probes) and/or rate-limit it.

3. QsvDiagnosticPanel auto-runs the diagnostic on mount - it receives selected but the useQuery has no enabled: selected, so opening Camera settings spawns the ffmpeg probe even for users on software. Gate the query on enabled: selected, or make it a manual "Run check" button. (Together with #2, this means just viewing settings spawns ffmpeg.)

4. No runtime fallback if QSV fails. If QSV is enabled and later breaks (driver update, device busy, permissions), the stream just dies with no fallback to software. The panel helps before enabling, but not after. Please fall back to software on a QSV pipeline failure, or at least surface a clear error.

5. The render device is hardcoded to /dev/dri/renderD128 in both the pipeline and the diagnostic. Hosts whose render node is renderD129 (second GPU or different enumeration) can't use QSV - auto-detecting the node or making it configurable would be more robust.

6. Docs: the "no docs required" box is checked, but this adds a user-facing setting with real host prerequisites (/dev/dri passthrough into the container, render-group permissions, an ffmpeg build with h264_qsv/mjpeg_qsv). Please add a short wiki setup note covering those.

pterodaktil02:
Added the QSV changes requested during review.

Changes:
- Added automatic discovery of compatible `/dev/dri/renderD*` devices instead of assuming `renderD128`.
- Discovered devices are validated using the same VAAPI -> QSV initialization chain used by the actual camera stream.
- Added a process-local device cache with a forced refresh when the manual diagnostic is run.
- Added a proper runtime fallback to software processing for QSV hardware failures without incorrectly switching on normal network or RTSP disconnects.
- Added unit tests for device discovery, caching, and fallback behavior.

Validation:
- Backend: 4946 tests passed.
- Frontend: 2486 tests passed across 189 test files.
- ESLint passed.
- i18n parity passed for all 12 locales.
- TypeScript/Vite production build passed.
- Live-tested on an Intel N100:
  - `/dev/dri/renderD128` was detected automatically;
  - QSV initialized successfully;
  - the RTSP/MJPEG stream worked without falling back to software processing.

Documentation:
Setup, compatibility checks, diagnostics, measured performance, and Proxmox LXC passthrough are documented in:
https://github.com/maziggy/bambuddy-wiki/pull/59

maziggy (changes requested):
Thanks for the rework - the round-one items are all genuinely addressed: renderD* discovery, the probe using the real VAAPI -> QSV chain, the cache with a forced refresh on the manual diagnostic, and the tests around all of it. ruff is clean, and backend (5032 unit + 2418 integration), frontend (2488), tsc, eslint and the build all pass here. 

Two blockers left.

1. The fallback still fires on healthy streams. is_qsv_failure substring-matches "qsv" / "vaapi" / "mfx" anywhere in the captured stderr, but the camera command passes neither -hide_banner nor -loglevel (the probe command does), and _summarize_ffmpeg_stderr only strips the version/built/configuration/lib* lines. So for a stream that runs cleanly and logs nothing, the surviving "last 10 meaningful lines" are ffmpeg's startup block, which always contains 'Stream #0:0 -> #0:0 (h264 (h264_qsv) -> mjpeg (mjpeg_qsv))`. When the printer drops RTSP - routine, and the docstring right there says it is common on P2S - stream_ended runs, is_qsv_failure returns true even though got_any_frames is true, and the stream reconnects in software with the device    cache cleared. QSV silently stops being used after the first reconnect.

I confirmed it by taking test_network_disconnect_after_frame_keeps_qsv and    changing only the stderr from b"" to real ffmpeg startup output; both assertions fail and the log shows "Intel Quick Sync failed ... falling back to software video processing". The current test passes only because _StreamProcess defaults to stderr=b"", which real ffmpeg never produces.

Simplest fix is to gate the post-frame fallback on `not got_any_frames` - a hardware init failure never yields frames. If you want to keep catching mid-stream GPU faults, require an error token (error / failed / cannot / no device / Function not implemented) on the same line as the qsv marker. Either way, please give that test realistic stderr.

2. Ukrainian is missing all 28 new keys. uk.ts landed on dev on 29 July, after your last push, so npm run check:i18n now fails (en 5709 vs uk 5681). Worth noting the inline English defaults in the t() calls hide this at runtime, so it will not show up by clicking around.

Four smaller things I would like in the same round:

3. The diagnostic resolves ffmpeg with shutil.which, but the camera uses    get_ffmpeg_path(), which also checks /usr/bin, /usr/local/bin, /opt/homebrew/bin, /snap/bin and C:\ffmpeg\bin - added specifically for systemd services with a trimmed PATH. The check-ffmpeg endpoint 15 lines above yours uses the helper. As it stands the diagnostic can report ffmpeg_not_found on a box where streaming works, or probe a different binary than the one that streams.

4. The stage order makes the codec check unreachable in the case it exists for. Device probing runs first and its command ends in -c:v mjpeg_qsv, so an ffmpeg build without the QSV encoders fails every probe and returns qsv_initialization_failed with qsv_codecs marked "skipped". The user is told to check the media driver and oneVPL runtime when the real answer is "this ffmpeg has no mjpeg_qsv" - which is the usual cause on distro packages. It also leaves mjpeg_qsv_missing and qsv_codecs_missing as dead branches in the panel. Running the codec query before device probing fixes both.

5. The user-facing strings still hardcode /dev/dri/renderD128 - the settings description and two panel failure messages, in all 12 locales - even though the code now discovers renderD*. Better to correct it before it needs re-translating.

6. qsv_diagnostic.py has no tests at all: 303 lines, a new authenticated endpoint, eight failure codes. probe_qsv_device is untested too. Happy path plus ffmpeg-missing, no-devices, all-probes-fail and codec-missing would cover it.

Nits: negative probe results are not cached, so a machine with QSV selected but broken re-probes every render node (up to 10s each) on every stream start; the setting is read once when the shared upstream starts, so toggling it does nothing until the last viewer disconnects, which is worth a line of help text; on the Windows build discovery is /dev/dri-only so the option can only ever fall back to software; -global_quality 80 against -q:v 5 is not an equivalence, so a comment on what it was tuned against would help; and
camera.py could use the existing get_setting(db, key) helper instead of a raw select on the Settings table.

One thing on the wiki PR: Intel-Quick-Sync-Camera.md is at the repository root, but the wiki is MkDocs with all content under docs/ - a root-level file is outside the build and the nav. It wants to live under docs/features/ with an entry in mkdocs.yml.

pterodaktil02:
Addressed the review feedback:

- prevented QSV fallback after a stream had already produced frames;
- added regression coverage using realistic FFmpeg QSV startup output;
- switched diagnostics to the configured FFmpeg path and check codec support before device probing;
- added diagnostic and QSV device probe tests;
- cached negative probe results;
- added the missing Ukrainian translations and the shared-stream setting note;
- removed the hardcoded render device from user-facing text;
- used the existing settings helper and documented the QSV quality choice.

Backend tests, i18n parity, frontend lint, and the production build pass.

maziggy:

maziggy (changes requested):
Thanks for this — the shape is right and the groundwork is solid. Software stays the default, the setting is validated server-side, the render device is discovered by globbing /dev/dri rather than accepted from the client, and the new log lines go through _summarize_ffmpeg_stderr, so they inherit credential redaction instead of leaking the access code from the RTSP URL. qsv.py and qsv_diagnostic.py have 15 real behavioural tests between them, and the staged diagnostic (codecs before devices, so a codec-less build is not misreported as a driver failure) is a nice touch.

Verified on 5d7b62669: backend unit 5580 passed, integration 2431 passed, ruff check clean, eslint clean, frontend build clean, i18n parity green across all 13 locales.

One CI gate is red and there is one logic issue I would like fixed before merge.

Blocker: ruff format

Four files fail ruff format --check:

backend/app/schemas/settings.py
backend/app/services/qsv_diagnostic.py
backend/tests/unit/services/test_qsv.py
backend/tests/unit/services/test_qsv_diagnostic.py

ruff format backend/ fixes all four. ruff check itself is clean, it is only the formatter.

The QSV fallback fires on any startup failure, and blames QSV in the log

is_qsv_failure() at camera.py:744 is never called - not in the module, not in the tests. The immediate-failure branch falls back on use_intel_qsv and not qsv_fallback_used without looking at the stderr at all.

I drove the real generator with QSV selected and a plain connection-refused stderr, i.e. the printer is simply switched off and QSV is not involved:

ffmpeg invocations: 2   ([0] qsv=True, [1] qsv=False)
clear_qsv_device_cache() calls: 1
log: Intel Quick Sync failed for 192.0.2.17 (stream_id=...):
     [tcp @ 0x55d3] Connection to tcp://127.0.0.1:48521 failed: Connection refused;
     falling back to software video processing

Three consequences. The log misdirects anyone debugging an offline printer. clear_qsv_device_cache() is global, so one unreachable printer discards a valid probe result for every printer and the next stream re-probes from scratch. And the existing "Camera connection failed. Check printer is on and camera is enabled." message is pushed back a full attempt, because the QSV branch sits ahead of the not got_any_frames and reconnect_count == 0 check.

Nothing breaks and it self-heals per stream, so this is not a correctness blocker - but on a setup where a camera is intermittently reachable, QSV gets skipped and re-probed repeatedly while the log blames the hardware.

The fix is the function you already wrote. I lifted is_qsv_failure out and ran it against summarised stderr:

┌──────────────────────────────────────────────────────────────────────────────┬─────────────────┐
│                                    stderr                                    │ is_qsv_failure  │
├──────────────────────────────────────────────────────────────────────────────┼─────────────────┤
│ camera off, connection refused                                               │ False (correct) │
├──────────────────────────────────────────────────────────────────────────────┼─────────────────┤
│ Error initializing an MFX session (your own fixture)                         │ True (correct)  │
├──────────────────────────────────────────────────────────────────────────────┼─────────────────┤
│ Failed to initialise VAAPI connection                                        │ True (correct)  │
├──────────────────────────────────────────────────────────────────────────────┼─────────────────┤
│ healthy QSV stream that later EOFs, Stream mapping ... h264_qsv -> mjpeg_qsv │ True (wrong)    │
└──────────────────────────────────────────────────────────────────────────────┴─────────────────┘

It discriminates correctly on exactly the input the immediate-failure branch receives - the banner and the configuration: --enable-vaapi line are stripped by _summarize_ffmpeg_stderr before it gets there. The last row is why your comment on the stream_ended branch rejected marker matching, and that reasoning is right for that branch. It just got applied to both, which left the accurate discriminator unused at the one site where it works. Gating the immediate-failure fallback on is_qsv_failure(stderr_text) fixes the misattribution, stops the spurious global cache clear, and restores the first-attempt error message. Keep not got_any_frames in stream_ended.

Inline t() defaults disagree with en.ts, and that hides the requirements

The key exists in en.ts, so the locale value wins and the JSX default never renders:

┌──────────────────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────┬──────────────────────┐
│                   key                    │                                       JSX default (dead)                                        │ what actually ships  │
├──────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────┤
│ cameraVideoProcessingIntelQsvDescription │ "Uses Intel Quick Sync for hardware H.264 decoding and MJPEG encoding. Requires Intel media     │ "Processing on Intel │
│                                          │ runtime and access to a DRM render device."                                                     │  GPU"                │
├──────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────┤
│ cameraVideoProcessingSoftwareDescription │ "Uses the CPU for H.264 decoding and MJPEG encoding."                                           │ "Processing on CPU"  │
├──────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────┤
│ cameraVideoProcessingSoftware            │ "Software processing"                                                                           │ "Software"           │
└──────────────────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────┴──────────────────────┘

The first one matters. That requirements sentence is the most useful thing for someone deciding whether to turn this on, and it appears nowhere in the UI - a user selects Intel Quick Sync, reads "Processing on Intel GPU", and has no indication they need /dev/dri mapped into the container. They find out only after running the diagnostic and getting a failure. Either promote the longer text into en.ts and the other 12 locales, or shorten the JSX defaults to match, so the source stops describing UI that does not exist.

selected is always true, so a third of the panel's styling is unreachable

SettingsPage.tsx:2054 renders <QsvDiagnosticPanel selected /> inside the === 'intel_qsv' branch, and that is the only call site. Every : selected ? ... : ... in the component - the red border, the red text, the XCircle icon - can never reach its final arm. Either drop the prop and the dead arms, or render the panel unconditionally, which I suspect is what it was designed for: showing a red unavailable state while software is still selected.

No frontend tests

QsvDiagnosticPanel.tsx is 311 lines with four render states and a nine-case code-to-message map whose default silently swallows any unrecognised summary_code. There are no tests for it, and none for the new setting in SettingsPage. The backend side is well covered, which makes the gap stand out - the mapping in particular is pure and cheap to test, and today a backend code the frontend does not know about degrades to a generic message with nothing to catch it.

Minor

- probe_qsv_device has no test for its render_device_missing or render_device_permission_denied branches, though both surface as distinct user-facing messages. Those are the two most likely outcomes for a Docker user who has not mapped the device.
- find_qsv_render_device is awaited before ffmpeg spawns, so a hung probe can delay the first viewer's stream by up to 10s per render node with no feedback. Low risk in practice - a missing /dev/dri returns instantly and a broken driver fails fast - but the ceiling is there.

One question

The feature needs an ffmpeg built with QSV/VAAPI and /dev/dri passed into the container. Does the shipped image's ffmpeg carry h264_qsv and mjpeg_qsv? If not, the option is selectable but cannot succeed for most installs. The diagnostic panel makes that diagnosable, which is the right call - I just want to confirm whether the image needs a change alongside this, and that the wiki PR covers the compose devices: mapping.

On the translations: the auto-generated locales read fine to me and parity is green, so no concern there. Thanks again - the N100 numbers are a compelling reason to have this.

pterodaktil02:
**Addressed the review findings:**

Applied Ruff formatting across the touched backend files.
Restricted immediate software fallback to failures that are actually identified as QSV-related. A camera connection failure no longer disables QSV or clears the device cache.
Aligned the video-processing labels and descriptions across all locales and removed conflicting inline translation defaults.
Kept the QSV diagnostic panel visible in both processing modes, so its selected/unselected warning states are now reachable.
Added frontend coverage for idle, available, known failure, unknown failure, and both selected/unselected unavailable states.
Added direct backend coverage for missing and inaccessible DRM render devices.

**Verification:**

Backend targeted tests: 21 passed
QSV diagnostic frontend tests: 6 passed
ESLint passed
All 13 locales are in parity
Production frontend build passed

**I have not changed the Docker image or Compose device mapping. I do not normally use the Docker deployment and I am not sufficiently familiar with this project’s container build to make image-level dependency changes blindly**. The native Linux path and application behaviour have been verified, but the published image’s Intel runtime/QSV codec support and /dev/dri mapping still need verification by someone familiar with the Docker setup.

maziggy (commented):
Blockers

1. ruff format --check backend/ fails on four of this PR's files

Would reformat: backend/app/schemas/settings.py
Would reformat: backend/app/services/qsv_diagnostic.py
Would reformat: backend/tests/unit/services/test_qsv.py
Would reformat: backend/tests/unit/services/test_qsv_diagnostic.py

All four belong to this PR, and it is a CI gate (ci.yml:54), so the build is red. ruff format backend/ clears it.

2. is_qsv_failure is never called

camera.py:744 defines it, and nothing in the tree calls it. The immediate-exit branch falls back unconditionally:

if use_intel_qsv and not qsv_fallback_used:
    fall_back_to_software(stderr_text or "FFmpeg exited during QSV startup")
    continue

With QSV selected, a printer that is simply powered off makes ffmpeg exit immediately, which calls fall_back_to_software(), which calls clear_qsv_device_cache() — wiping the process-global device cache so the next camera stream on any printer re-probes every render node from scratch. That is the behaviour the last update describes as fixed: "A camera connection failure no longer disables QSV or clears the device cache."

Ruff does not catch it because it is a nested function, and no test covers it — all three tests in test_camera_qsv_fallback.py stub clear_qsv_device_cache, but none asserts it is not called on a non-QSV failure.

The second fallback site deliberately uses not got_any_frames instead of marker matching, and the reasoning there is right. So the fix is either to call is_qsv_failure at the immediate-exit site, or to delete it and say in a comment why an unconditional first-attempt fallback is acceptable.

3. The diagnostic panel's unselected state cannot render

SettingsPage.tsx:

{(localSettings.camera_video_processing ?? 'software') === 'intel_qsv' && (
  <QsvDiagnosticPanel selected />
)}

selected is hardcoded true and the panel only mounts once QSV is already chosen, so the three : selected ? ternaries in the component — red border, XCircle, red text — are unreachable in the product. The last update lists "Kept the QSV diagnostic panel visible in both processing modes, so its selected/unselected warning states are now reachable."

It also works against the panel's own copy, "Run a compatibility check before using Intel Quick Sync", since you have to switch to QSV first to reach the button.

4. The frontend tests are not in the branch

The update reports frontend coverage for idle, available, known failure, unknown failure and both selected/unselected unavailable states, with 6 tests passing. There is no file matching qsv anywhere under frontend/src, and the PR file list contains only the three backend test files. Possibly committed elsewhere.

5. Inline defaults still conflict, and the informative text is the dead one

Also listed as removed. Comparing SettingsPage.tsx against en.ts:

┌──────────────────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────┬─────────────────────┐
│                   key                    │                                          inline default                                          │   shipped string    │
├──────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────┤
│ cameraVideoProcessingIntelQsvDescription │ Uses Intel Quick Sync for hardware H.264 decoding and MJPEG encoding. Requires Intel media       │ Processing on Intel │
│                                          │ runtime and access to a DRM render device.                                                       │  GPU                │
├──────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────┤
│ cameraVideoProcessingSoftwareDescription │ Uses the CPU for H.264 decoding and MJPEG encoding.                                              │ Processing on CPU   │
├──────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────┤
│ cameraVideoProcessingSoftware            │ Software processing                                                                              │ Software            │
└──────────────────────────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────┴─────────────────────┘

The locale value wins, so users see "Processing on Intel GPU" and never the sentence naming the actual requirements — which, for a feature that silently falls back to software, is the one line that matters. Better to move that text into the locale strings than to drop it.

On the Docker question

You flagged that you did not want to make image-level changes blind. Here is what I found, so it does not have to be blind:

- The codecs are already present. Debian trixie ships ffmpeg 7.1.5-0+deb13u1 with both h264_qsv (decoder) and mjpeg_qsv (encoder) compiled in — verified directly. The image is python:3.13-slim-trixie plus apt-get install ffmpeg, so no ffmpeg change is needed.
- The VAAPI and oneVPL runtimes are already present too. libavutil59 depends on libva2, libva-drm2, libvpl2 and libdrm2, all pulled in transitively.
- The one missing package is the driver: intel-media-va-driver (iHD). Nothing in the image depends on it, which is why -init_hw_device vaapi=va:/dev/dri/renderD128 cannot succeed today.
- It cannot be added unconditionally. The published image is a multi-arch manifest covering linux/amd64 and linux/arm64 (confirmed with docker manifest inspect), and intel-media-va-driver is amd64/i386-only in Debian, so a plain apt-get install breaks the arm64 build. It needs an arch guard:

RUN if [ "$(dpkg --print-architecture)" = "amd64" ]; then \
      apt-get update && apt-get install -y --no-install-recommends intel-media-va-driver && \
      rm -rf /var/lib/apt/lists/*; \
    fi

- docker-compose.yml maps no /dev/dri and adds no render group. Both are required, and devices: - /dev/dri:/dev/dri cannot be enabled by default — compose refuses to start on any host without an Intel GPU, which is most of the install base. It has to ship commented out, pointing at the wiki.
- The container drops privileges via gosu to PUID/PGID, so the runtime user also needs the host's render GID through group_add. That value is host-specific, so it belongs in documentation rather than hardcoded in the compose file.

Until that lands, the setting is safe but inert under Docker: choosing Intel Quick Sync always falls back to software and the diagnostic always reports failure. Since Docker is the primary install path, it is worth either bundling the Dockerfile and compose change here, or gating the option behind a successful diagnostic so nobody can select a mode that cannot work.

Smaller things

- The diagnostic holds _qsv_device_lock for up to 10s per render node. A camera stream starting during a diagnostic blocks on the same lock, so clicking "Run compatibility check" can stall a stream start by tens of seconds on a multi-GPU host.
- qsv_diagnostic.py:225 treats an empty probe list as render_device_missing. That is correct only because it passes refresh=True — with refresh=False an empty list means "cached device reused", so a working setup would report no render device. Worth making find_qsv_render_device return something unambiguous instead of depending on the caller's flag.
- No test for render_device_permission_denied. test_qsv.py covers the ffmpeg-failure branch but neither the device.exists() nor the os.access() branch. The update lists "missing and inaccessible DRM render devices" as covered; only the missing case is, and only at the diagnostic level.
- stageLabel returns undefined for a stage name outside the union. TypeScript pins the union at compile time, but the value arrives from the API at runtime, so a future backend stage renders a blank row.
- camera.py imports get_setting from routes.settings at module top; the convention elsewhere in this repo is a lazy import across route modules.
- The check-i18n-parity.mjs allowlist addition is fine — "Intel Quick Sync" and "FFmpeg" are exactly what that list exists for, and the surrounding strings are genuinely translated. I spot-checked the German and it reads naturally.

pterodaktil02:
@maziggy, thank you for the detailed review. I went through every point, updated the branch, and verified the current state.

Some of the blockers referred to an earlier revision of the PR and were already resolved in the current branch:

ruff format --check backend/ passes: all 682 files are formatted;
is_qsv_failure() is used in the immediate FFmpeg-exit path;
a normal connection failure caused by an offline or unreachable printer does not switch to software processing and does not clear the process-wide QSV device cache;
there is an explicit regression test asserting that clear_qsv_device_cache() is not called in that case;
the diagnostic panel is rendered in both software and QSV modes, so the compatibility check can be run before enabling QSV;
the frontend tests are present in the branch: 7 panel tests currently pass, covering both selected states, known and unknown failures, and the missing QSV runtime case;
the informative processing-mode descriptions are stored directly in the locale files and do not rely on inline defaults.

I also made the following changes:

MFX session -9 is now classified as qsv_runtime_missing;
users receive a specific suggestion to install libmfx-gen1.2 on current Debian and Ubuntu systems;
an unknown diagnostic stage name now falls back to the original stage.name instead of rendering an empty label;
get_setting was moved to a lazy import to avoid a top-level dependency between route modules;
intel-media-va-driver was added to the amd64 Docker image behind an architecture check, so the arm64 build should remain unaffected;
the wiki now documents oneVPL runtime requirements, render-device permissions, Docker-specific considerations, and the Ubuntu 22.04 / FFmpeg 4.4 workaround.

Docker is not installed on my system, so I could not build the image locally. I checked the Dockerfile and the architecture condition statically; the final multi-arch build will need to be confirmed by CI.

I deliberately did not add an active /dev/dri mapping to the default docker-compose.yml, because Compose would refuse to start on hosts where that device does not exist.

I also did not document group_add as a complete solution. The current entrypoint launches the application through gosu PUID:PGID, so supplementary groups are not preserved. Proper support for a separate RENDER_GID would require an entrypoint change and dedicated testing, not just a Compose edit.

I left two points unchanged for now:

_qsv_device_lock is held during probing. This can delay a concurrent camera start, but the lock currently protects the shared cache and serializes hardware probes. I did not want to change that behaviour without separate testing;
an empty probe list has different meanings when the cache is reused and when a full refresh is performed. The diagnostic always uses refresh=True, so there is no current functional bug there, but the function contract could indeed be made more explicit in a separate change.

Checks after the updates:

backend QSV tests: 24 passed;
frontend QSV diagnostic panel: 7 passed;
ruff format --check backend/: 682 files already formatted;
git diff --check: clean.

Could you please take another look at the current branch?

maziggy (changes requested):
Blockers

1. The official Docker image cannot run QSV — it's missing the runtime, not the driver.

I built the image's package set in a container. The result: ffmpeg does carry h264_qsv and mjpeg_qsv, and intel-media-va-driver pulls in libvpl2 — but libvpl2 is only the oneVPL dispatcher. The GPU implementation that actually creates the MFX session, libmfx-gen1.2, is not installed.

That is precisely the state this PR's own code names:

# oneVPL dispatcher is present, but no Intel GPU implementation can create
# an MFX session ... commonly means that the libmfx-gen1.2 runtime package is missing.
return "qsv_runtime_missing"

So on Gen12+ hardware — including the N100 the PR benchmarks — the amd64 image should land on qsv_runtime_missing, and the panel will tell the user "On Debian or Ubuntu, install libmfx-gen1.2", which is not something a user of a published image can act on. The PR's own wiki page gets this right for native installs: apt install ffmpeg intel-media-va-driver libvpl2 libmfx-gen1.2 vainfo. The Dockerfile just didn't follow.

libmfx-gen1.2 is in trixie main at 25.1.4-1, so this is one more package on the amd64 conditional that's already there.

I verified package presence, not runtime behaviour — there's no Intel GPU on this box, so someone with the hardware should confirm the image works once the package is added.

2. The project's i18n parity gate fails on this branch.

node scripts/check-i18n-parity.mjs → ❌ i18n parity check failed. settings.qsvDiagnosticRuntimeMissing is verbatim English in all 11 non-en/ru locales. 29 of the 30 new keys were translated properly; this one arrived in the later Improve QSV runtime diagnostics commit and missed the pass. Given finding 1, it's the string most likely to actually be shown.

3. The software fallback doesn't survive the stream, so every camera open retries a failing QSV pipeline.

qsv_fallback_used is a local of the generator, and fall_back_to_software() calls clear_qsv_device_cache() — which discards the negative cache along with the positive one. Two consecutive camera opens against a QSV pipeline that fails:

total_spawns=4  qsv_attempts=2  cache_clears=2

And on the qsv module itself, after a clear the next lookup re-probes every render node and hands back the same device it just discarded:

probes_run=2  same_device_returned=True

With the real probe that's a subprocess per render node, up to 10 s each, in the path of opening a camera — permanently, for every viewer.

Findings

4. The diagnostic never exercises h264_qsv, which is what makes 3 loop forever.

build_qsv_probe_command is lavfi colour → hwupload → mjpeg_qsv: device init plus the encoder. For the decoder it only greps ffmpeg -decoders for the string h264_qsv — that proves the build has it, not that it initializes on this GPU against this camera's stream. So the panel can report green while the camera silently falls back, and since the probe keeps passing, the fallback never sticks. Either give the probe a decode leg against a tiny H.264 sample, or word the success state to say what was actually tested.

5. The two fallback paths use inconsistent criteria, and one fails badly.

stream_ended falls back on not got_any_frames with no marker matching — the right call, since ffmpeg's own stream-mapping line contains mjpeg_qsv and markers would false-positive on every normal reconnect. But the immediate-failure path requires is_qsv_failure(stderr_text). When that misses, there is no fallback and no retry:

attempts=1  software_attempted=False
user_sees=b'Error: Camera connection failed. Check printer is on and camera is enabled.'

The camera is dead and the message points at the printer. To be straight about the evidence: I synthesized that stderr, so this demonstrates the code path rather than proving a specific real ffmpeg message reaches it — in practice ffmpeg usually prefixes encoder errors with [mjpeg_qsv @ ...], which does match. The argument is the asymmetry: guessing wrong costs a dead camera and a misleading message, while simply falling back whenever QSV was active and ffmpeg died instantly costs ~0.1 s on an unreachable camera that was going to fail anyway.

Worth separating the two concerns while you're in there — "this stream drops to software" is cheap and safe, "the global device selection is bad" is a process-wide side effect that shouldn't be triggered by one printer's network problem.

Minor: ffmpeg_codec_query_failed is the one emitted code with no dedicated panel string; it falls through to the generic message.

maziggy:
This was the 5th code review and the code is still not good to merge. Bambuddy is maintained by one person, and I don't have the resources to do unlimited reviews for each PR.

PR #2625, opened 2026-07-21
feat(oidc): configure an OIDC provider from environment variables

### LINKED ISSUE #2593: [Feature]: Configure an OIDC provider via environment variables
### Problem or Use Case

OIDC providers can currently only be configured through the admin UI, which writes to the `oidc_providers` table. That works well for a hand-tended instance, but it's awkward for containerized/declarative deployments.

In my case Bambuddy runs on Kubernetes with a GitOps flow (ArgoCD), and the OIDC client credentials for our Keycloak live in a secret store and are injected as environment variables. Today that means:

- Every fresh deploy or volume reset needs a manual pass through the UI before anyone can log in via SSO — click-ops in an otherwise declarative stack.
- The `client_secret` has to be pasted into a form and lives in the database, rather than staying in the secret store it's rotated from.
- The provider config can't be version-controlled or reproduced across staging/prod.

This is the same gap that `HA_URL` / `HA_TOKEN` already close for the Home Assistant integration — that one can be fully driven by env, OIDC can't.

### Proposed Solution

Allow **one** OIDC provider to be defined via environment variables, following the pattern the Home Assistant integration already established.

Sketch:

```
BAMBUDDY_OIDC_NAME=Keycloak
BAMBUDDY_OIDC_ISSUER_URL=https://sso.example.com/realms/main
BAMBUDDY_OIDC_CLIENT_ID=bambuddy
BAMBUDDY_OIDC_CLIENT_SECRET=...
BAMBUDDY_OIDC_SCOPES=openid email profile
BAMBUDDY_OIDC_AUTO_CREATE_USERS=true
# ... plus the remaining OIDCProvider fields as optional vars
```

Behaviour:

- **On startup**, if the required vars are set, the provider is **upserted** into `oidc_providers` and flagged as eal provider for every existing read path, so no auth code changes.
- **Coexistence:** UI-created providers are untouched and stay editable. Only the env-managed one is locked.
- **UI:** the env-managed provider renders read-only with a lock icon and an "environment managed" badge — exactly what `SettingsPage.tsx` already does for `ha_url_from_env` / `ha_token_from_env`.
- **Secret handling:** the value from env is assigned through the existing `client_secret` property, so it takes exs a UI-entered secret (`mfa_encrypt`, Fernet when a key is available) — no new secret handling.
- **Typo guard:** the new vars would be registered in `_INTENTIONAL_UNSETTINGS` (`core/config.py`) so they don't trip the "possible typo" INFO log. Side note: `BAMBUDDY_LOCAL_LOGIN` and `BAMBUDDY_EXTERNAL_ROOTS` appear to be missing from that set today and log spuriously — happy to fix that in passing if you want it.

Two implementation details I'd like your opinion on, because they're the actual design decisions here:

**1. Identifying the env-managed row across restarts.** `user_oidc_links.provider_id` is an FK with `ON DELETE CASCupdated stably — recreating it on each startup would cascade away every user's account link. I'd add an`is_env_managed BOOLEAN` column (a two-line idempotent `_safe_execute` in `run_migrations`, matching the existing style) and use it to locate the row. Matching on `name` or on `issuer_url`+`client_id` instead avoids the schema change, but then any legitimate change to those values in env orphans the old row *with its user links still attached* — which seems like a worse trade for a self-hosted app.

**2. How strict the lock should be.** The HA precedent is UI-only: env wins on read, but a `PUT` to `ha_url` still lands in the DB. For OIDC I'd lean toward the API also rejecting writes to the env-managed provider , since a silently-ignored write to an auth provider is more confusing than one to a URL field — and locking oneself out isn't a concern here, because ) already provides the documented "SSO is broken, let me back in" path. Happy to mirror HA exactly instead if you'drather keep the two consistent.

I'm aware this touches auth and adds a column, so I wanted to agree on the approach before writing code rather than to redesign.

### Alternatives Considered

- **Multiple providers via indexed vars** (`BAMBUDDY_OIDC_1_*`, `BAMBUDDY_OIDC_2_*`) — more flexible, but a much larger surface for matching, collisions with UI-created providers, and validation. Single-provider covers the common self-hosted case (one IdP) at a fraction of the complexity. Could be extended later without breaking the flat names.
- **One JSON blob var** (`BAMBUDDY_OIDC_PROVIDERS=[{...}]`) — compact, but doesn't match how any other config in the project is expressed, and gives poor validation errors.
- **Ignore the DB entirely / virtual provider** — not possible: `user_oidc_links.provider_id` FKs to `oidc_providers.id`, so a provider without a row can't be linked to a user.
- **Seed the DB externally** (SQL/init script) — brittle, dialect-specific (SQLite *and* Postgres), duplicates the encryption logic for `client_secret`, and breaks on schema changes.
- **Status quo / workaround:** configure once via UI and back up the volume. Works, but defeats reproducible deploys and keeps the secret in the DB.

### Feature Category

Other

### Priority

Would improve my workflow

### Mockups or Examples

The UI treatment I'm proposing already exists in the app — the Home Assistant section in Settings when `HA_URL`/`HA_TOKEN` are set (lock icon, "auto-enabled via env" badge, disabled inputs). I'd reuse that exact pattern for the env-managed OIDC provider.

### Contribution

### Checklist

munin92:
## Description

Adds a way to configure **one** OIDC provider entirely from environment variables (`BAMBUDDY_OIDC_*`), for deployments driven by a compose file, Helm chart or GitOps repo where nobody clicks through the settings UI.

The provider is upserted on startup and re-applied on every boot. It coexists with providers created in the UI, which stay fully editable.

## Related Issue

Fixes #2593

## Documentation

- Wiki: maziggy/bambuddy-wiki#55
- Website: **not required.** `features.html` already lists Single Sign-On (OIDC / SSO),
  and this PR adds a configuration path rather than a capability. The one claim that
  touches configuration — "the admin UI can opt-in to auto-provisioning … or
  auto-linking" — stays true; it is simply no longer the only way. Noted explicitly
  rather than deleting the line, so the omission reads as a decision.

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- **`is_env_managed` column** on `oidc_providers`, with a migration that reaches existing installations. The row is never delete-recreated: `user_oidc_links.provider_id` is `ON DELETE CASCADE`, so recreating the provider would silently unlink every bound account.
- **Reader** (`core/oidc_env.py`) — all four required variables or nothing; an empty value counts as unset, so a half-configured provider never reaches the database to fail later at authorize time. Booleans use the same `{true, 1, yes}` rule as `BAMBUDDY_LOCAL_LOGIN` (`api/routes/auth.py`), extended only by a `default` parameter, which that call site does not need and two of these variables do.
- **Upsert on startup**, validated through the same `OIDCProviderCreate` schema the API uses — env config cannot reach a state the UI would refuse, including the auto-link/verified-email rule. Nothing raises: a typo must not stop the app from booting, so a rejected config is logged and skipped, leaving any previous provider untouched.
- **Removing the variables disables**, rather than deletes — for the same cascade reason as above. Re-adding them restores the provider with its links intact. The `is_env_managed` flag is cleared at the same time: a provider the API still refuses to edit or delete, with no configuration left behind it, would be a dead end reachable only through the database.
- **`BAMBUDDY_OIDC_AUTOLOGIN=true` clears `is_autologin` on every other provider** on each boot. Autologin is exclusive by nature — two providers both claiming the redirect is not a state the login page can serve — but this is the one place where env config writes to rows it does not own, so it is called out rather than left to be discovered. A provider set as the autologin target in the UI will not stay set while the environment claims it.
- **API refuses writes** to the env-managed provider with 409 on all four mutating routes, including both icon routes (the icon comes from `BAMBUDDY_OIDC_ICON_URL` and would be restored anyway). Extracted as one helper rather than four copies, so a fifth route cannot be added with the guard silently missing.
- **`is_env_managed` in the provider response**, and the settings UI renders the provider read-only with a lock badge — reusing `settings.environmentManagedLabel`, the string the Home Assistant env-managed fields already use, so no new key to keep in parity across eleven locales.
- **Typo-guard registration** for the twelve variables, so a correct config no longer logs "possible typo" on every boot.
- **`.env.example`** documents the variables in the style of the `BAMBUDDY_LOCAL_LOGIN` block.

## Screenshots

Only the new state is shown: before this PR an environment-managed provider could not exist, and providers created in the UI are untouched by it.

**Settings → Authentication → SSO / OIDC — the environment-managed provider**

shot1-redacted

**Login page — the provider's SSO button**

image

## Testing

Backend **7106 passed**, frontend **2382 passed**, `ruff` and `eslint` clean, `tsc --noEmit` clean, all eleven locales in parity.

Verified end-to-end against a real deployment (Kubernetes, Keycloak, existing database), not only in tests:

- the migration reached a grown database and the pod came up
- `Env-managed OIDC provider 'Keycloak' applied.`
- a second boot **updated the same row** rather than creating a duplicate
- the provider appeared on the login page and an SSO login completed (`Auto-linked existing user ... via email match`)
- the settings UI showed it locked, without edit, delete or enable controls

Every guard is pinned by a test that was watched failing first. The migration carries its own test rather than relying on the model test: a table created from metadata already has the column, so that path never exercises the `ALTER` — only `run_migrations` does, and that is the path an upgrade actually takes.

## Checklist

## Additional Notes

**Scope kept deliberately narrow.** The typo-guard also misses `BAMBUDDY_LOCAL_LOGIN` and `BAMBUDDY_EXTERNAL_ROOTS` today; that is a separate one-line fix and is not bundled here.

**No group mapping.** Bambuddy assigns OIDC users a fixed `default_group_id` and does not read a `groups` claim — unlike LDAP, which maps directory groups to Bambuddy groups. Mapping OIDC claims the same way is a genuine gap, but a separate feature, and not part of this PR.


maziggy (changes requested):
This is a genuinely excellent PR, Marian - the write-up is one of the clearest I've reviewed, and the hard parts are all handled correctly: the write-guard is applied to every mutating route (extracted so a fifth can't miss it), config removal disables-and-releases rather than deleting so the ON DELETE CASCADE never unlinks bound accounts, validation reuses OIDCProviderCreate so env config can't bypass the SEC-1 auto-link rule, and the secret is encrypted at rest via the property setter. Migration follows the existing SQLite/Postgres pattern with its own upgrade-path test. Nicely done.

One thing to fix before merge, then two minor notes:

1. The rejection path can log the client_secret. apply_env_oidc_provider does logger.error("BAMBUDDY_OIDC_* config rejected...: %s", exc). client_secret has max_length=512, so a longer secret raises string_too_long, and Pydantic v2's str(exc) embeds input_value='...' - I confirmed it leaks the value. The trigger is narrow, but it violates our no-secrets-in-logs rule. Please log exc.errors(include_input=False) (that strips the input) or just the failing field locations instead of the raw exception.

2. The upsert matches by name, so if a UI-created provider already has the configured BAMBUDDY_OIDC_NAME, startup overwrites its secret/issuer and flips it to env-managed/read-only. I understand why you chose name-matching (flag-matching hit the unique constraint and broke boot), and the bound links survive - but it means an accidental name clash silently absorbs and repoints an existing UI provider. Please call this out prominently in the wiki (#55) so an operator doesn't lose a UI provider to a name collision.

3. "Never raises" is only half true - the OIDCProviderCreate validation is guarded, but the flush/execute/commit calls aren't, so a transient DB error during apply would propagate out of the lifespan and block boot. Very low risk at single-threaded startup, but either wrap the DB writes too or note that a DB failure there is meant to block boot.

Really solid work.

munin92:
Thanks for the thorough review —
**1. Secret in the rejection log (blocker).** Fixed in `1116b43f`. The `except` is now split: a `ValidationError` logs `exc.errors(include_input=False)`, which strips the submitted values, so the `string_too_long` path can no longer surface `client_secret`; any other exception logs only `type(exc).__name__`, never `str(exc)`. Rejection stays survivable — a bad config is still skipped and the app still boots. Added two regression tests: one asserts an over-long secret is rejected without the value (or any fragment) reaching the logs, the other covers the non-`ValidationError` branch. I confirmed the first test fails on the pre-fix code and passes after.

**2. Name-match upsert can absorb a UI provider.** Documented prominently in the companion wiki PR (maziggy/bambuddy-wiki#55) — a `!!! warning` now spells out that the env provider is matched by `BAMBUDDY_OIDC_NAME` on every boot, so a name clash adopts and repoints an existing UI provider (links survive, but it's taken over), with guidance to pick a non-colliding name unless that repoint is intended.

maziggy (changes requested):
Thanks Marian - both of the items you addressed are properly done.

The secret fix is right. Splitting the catch so a ValidationError logs errors(include_input=False) and anything else logs only the class name is a structural guarantee rather than one contingent on which exceptions the validators happen to raise, which is the stronger version of what I asked for. Both regression tests are real, and patching OIDCProviderCreate at its source module rather than on oidc_env is exactly right given the in-function import. The wiki warning on name-matching in #55 is thorough and covers what an operator needs to know before a clash bites them.

Item 3 is still open, and following it up turned into a blocker.

apply_env_oidc_provider matches by name and sets is_env_managed on the match, but it never clears the flag from the row it managed previously. So renaming BAMBUDDY_OIDC_NAME in a compose file leaves two flagged rows. Running it:

  --- after rename (Keycloak -> Authentik) ---
    id=1 name='Keycloak'  enabled=True env_managed=True
    id=2 name='Authentik' enabled=True env_managed=True
    env-managed row count: 2
    boot 3 (vars removed) RAISED: MultipleResultsFound

Two consequences. The old row stays enabled and env-managed, so it keeps appearing on the login page with its stale issuer and secret, and _refuse_if_env_managed now 409s every attempt to edit, disable or delete it - the operator cannot remove it except through the database. That is precisely the dead end reachable only through the database that your release path was written to avoid. Then, unsetting the variables later hits scalar_one_or_none on two rows, MultipleResultsFound propagates out of the lifespan, and the app does not boot. Both reachable by two ordinary config edits, no DB tampering.

This is item 3 in concrete form, so I would rather see it fixed than noted. The upsert already does the right sweep one block down for autologin, and the same shape works here:

  await db.execute(
      update(OIDCProvider)
      .where(OIDCProvider.id != existing.id, OIDCProvider.is_env_managed.is_(True))
      .values(is_env_managed=False, is_enabled=False)
  )

Disable-and-release rather than delete, for the same cascade reason as everywhere else in this PR. Worth a test that renames across two applies and asserts exactly one flagged row, plus one that removes the config afterwards and asserts boot survives.

Three smaller things:

1. The release path clears is_enabled and is_env_managed but not is_autologin. Nothing breaks today since the public /oidc/providers filters on is_enabled, but the released row is now UI-editable with a latent autologin claim, and merely re-enabling it makes it the autologin target again - the exclusivity sweep in update_oidc_provider only runs on body.is_autologin is True. Clearing it alongside the other two is one word.

2. The icon buttons are still offered on the env-managed provider. refresh-icon-{id} and remove-icon-{id} render outside the !provider.is_env_managed block and both routes now 409, so clicking either shows an error toast - the exact thing the comment directly below them gives as the reason for hiding the rest.

3. There is no way to set default_group_id. No BAMBUDDY_OIDC_DEFAULT_GROUP, and
_APPLIED_FIELDS omits the field, so every user auto-created through the env provider falls back to Viewers (mfa.py:2013-2026) - and the lock means it cannot be corrected in the UI either. For a declarative deployment running AUTO_CREATE_USERS=true that is a real limitation. A variable matching on group name would close it, since IDs are not stable across installs; at minimum the wiki should say so. Your call whether that belongs in this PR or a follow-up - unlike the rest of the list it is a scope decision, not a defect.

munin92:
Thanks for the feedback again — all three are in.

d64f5d96 fixes the rename: the upsert now sweeps the flag off any other row, disable-and-release like the autologin block below it. Releasing also clears is_autologin, your point 1 — otherwise the released row keeps the claim and merely re-enabling it in the UI makes it the autologin target again. The release path releases every flagged row instead of scalar_one_or_none(); the sweep should keep that at one, but the invariant isn't enforced anywhere and I'd rather not have a second way to lose the boot if it ever breaks. Tests cover the rename, rename-then-unset (your MultipleResultsFound), a new name colliding with a UI provider, and account links surviving it. All of them fail on the old code.

ac2f4d05 moves the two icon buttons inside the existing !is_env_managed block.

3c679459 adds BAMBUDDY_OIDC_DEFAULT_GROUP, matched by name since ids aren't stable. An unknown name is refused rather than defaulted — quietly landing people in Viewers is the exact thing the variable is there to prevent, and the API answers 422 for an unknown default_group_id anyway. The catch is that on a first boot nothing gets created at all, so the log says which case it hit, and .env.example and the wiki spell it out. Easy to flip to "apply with no group + error line" if you prefer.

Backend 7427, frontend 2493, ruff/eslint/tsc clean. Wiki PR #55 updated to match.

maziggy (changes requested):
This round reads well. The cascade reasoning is right and you've applied it consistently rather than in the one place it was obvious: disable-never-delete on release, adopt-never-recreate on a name collision, and the release path clearing is_autologin so a row re-enabled later cannot silently reclaim the redirect. Validating through OIDCProviderCreate instead of a parallel path means the auto-link/verified-email rule cannot be sidestepped from the environment, which is the thing I would have pushed back on hardest. errors(include_input=False) plus logging only the exception class on the broad branch is exactly right for a secret-bearing config. The guard is one helper across all four mutating routes rather than four copies. And .env.example explains why each variable behaves the way it does instead of just naming it.

One thing to fix before merge.

apply_env_oidc_provider does not honour its "never raises" contract.

The docstring says it never raises so a typo cannot stop the app booting, and the call site in main.py repeats that. But the try only wraps OIDCProviderCreate(**config). The four db.execute() calls and both db.commit() calls sit outside it. Patching commit to raise:

backend/app/core/oidc_env.py:209: in apply_env_oidc_provider
    await db.commit()
E   RuntimeError: database is locked

That propagates straight out of the lifespan. It is the outcome the module is built to avoid everywhere else - the release path deliberately avoids scalar_one_or_none() because, in your own words, losing the boot is too steep a price for an invariant check - and then the database calls around it are unguarded. On Postgres a transient connection blip at startup now takes the instance down; on SQLite a WAL lock contended by another writer does the same. Wrap the body, or the call site, so the claim is true. Worth a test next to test_a_non_validation_error_is_survivable_and_leaks_nothing, which covers the schema branch but not this one.

Adopting a same-named UI provider is silent.

The behaviour is deliberate and you have tested it both ways, and the boot-safety reasoning for matching on name rather than the flag is sound - I am not asking you to change it. But an existing UI provider is overwritten in place. I ran it:

BEFORE: id=1 issuer=https://ui-configured.example.com/realms/main client_id=ui-client-id env_managed=False
AFTER:  id=1 issuer=https://env-configured.example.com/realms/other client_id=env-client-id
        secret=env-secret-value env_managed=True

The admin's issuer, client id and secret are gone, and the row is now locked behind 409. Unsetting the variables releases it but does not bring the old values back. Two asks, both small: log the adoption distinctly, since right now it emits the same "Env-managed OIDC provider 'X' applied." as a routine re-apply and nothing in the log says a UI-created provider was consumed; and say so in .env.example, because an operator choosing BAMBUDDY_OIDC_NAME=Keycloak has no way to know it will claim the "Keycloak" they already created.

Smaller things:

- .env.example does not state the issuer constraints. _validate_issuer_url applies the public-internet policy: https required, and private, loopback, link-local, numeric-encoded and IPv4-mapped forms all rejected. For the compose and GitOps audience this PR targets, http://keycloak:8080 is the natural first attempt, and it fails to one log line with no SSO button and no other signal. A sentence next to the ISSUER_URL example would save a support round.
- GET /oidc/providers is public and now returns is_env_managed to unauthenticated visitors. That route's response model already over-shares (auto_create_users, auto_link_existing_accounts, email_claim, default_group_id), so this widens something pre-existing rather than introducing it. Noting it rather than asking you to fix it here.
- _env_bool duplicates the truthy set in api/routes/auth.py, and you explain that the default parameter is why. Since yours is the superset, having that call site adopt it would leave one rule instead of two - and would close the typo-guard gap you flagged as out of scope, in the same line.

What I ran: ruff check and format clean across 669 files, tsc -b clean, eslint clean on all three changed frontend files, i18n parity across 12 locales with no new keys (confirmed settings.environmentManagedLabel resolves, so the reuse works as you described), 20 frontend tests passing on the new file, your six backend test files at 69 passed, and the full unit suite at 5542 passed.

On the integration suite I get 2465 passed and 4 failed, and those four are not yours. TestBackupKeyFiles in test_security.py fails with "column oidc_providers.is_env_managed does not exist". create_backup_zip uses the app-level engine bound to the configured DATABASE_URL rather than the test's in-memory SQLite, so on a Postgres environment it reads a database that has not had this branch's migration applied yet. It clears once the app boots on the branch. Your run passed because SQLite builds the schema from metadata, which already includes the column. There is a test-infrastructure issue there worth fixing separately; it is not something this PR introduced.

munin92:
All four done.

Blocker: wrapped the whole body, so every execute/commit is covered now (6dff1e96). The rollback was the same trap one level down — suppressed that too (6eea61dc). Both paths tested, no secret in the log either way.

Adoption logs a distinct warning now (c4b5d42f). .env.example has the issuer policy + the name-claim note (f6e0d767). LOCAL_LOGIN uses `_env_bool` (be194614).

The public `is_env_managed` and the four test-infra failures I'll take separately, not here.

maziggy (changes requested):
Two things left, both in the same corner of the reader.

1. An empty optional variable refuses the whole provider

The module docstring and .env.example both promise that an empty value counts as unset. That is implemented for the four required vars and for DEFAULT_GROUP (oidc_env.py:66), but not for SCOPES, EMAIL_CLAIM or ICON_URL — os.environ.get(key, default) falls back only when the key is absent, so an empty value goes straight through to a validator that rejects it:

reader icon_url = ''
REFUSED -> icon_url: Value error, icon_url must start with https://

Same for BAMBUDDY_OIDC_SCOPES= and BAMBUDDY_OIDC_EMAIL_CLAIM=. The operator gets no provider, no SSO button and one log line, from a variable they thought they were leaving at its default.

What lifts this above theoretical is that .env.example ships the trap:

# BAMBUDDY_OIDC_ICON_URL=
# BAMBUDDY_OIDC_DEFAULT_GROUP=

Uncommenting DEFAULT_GROUP is fine. Uncommenting ICON_URL takes the provider down with it. Three or None / or default normalizations in the reader close it, and test_a_blank_default_group_is_unset already has the shape to copy.

2. .env.example describes the opposite of what _env_bool does

The file says:

# Booleans accept true/1/yes; anything else keeps the default.

_env_bool returns the default only when the variable is absent. A variable set to anything outside {true, 1, yes} is False — which test_anything_else_is_false pins deliberately, so the code is doing what you meant and the documentation is what is wrong.

It matters for the two booleans that default to True:

'on'      -> require_email_verified=False
'enabled' -> require_email_verified=False
'y'       -> require_email_verified=False

BAMBUDDY_OIDC_REQUIRE_EMAIL_VERIFIED=on reads as off. The account-takeover combination stays fail-closed, because the SEC-1 validator refuses auto-link plus email_claim=email plus unverified — so this is not a hole. But with auto-link off and auto-create on, a typo silently drops the email_verified gate for every user the provider creates, and BAMBUDDY_OIDC_ENABLED=on silently disables the provider.

Fixing the sentence is the minimum. Rejecting an unrecognized value outright — log and skip, the way this module treats every other input it does not understand — would be better for the two security-relevant ones.

Minor

- auth.py:38 imports _env_bool, a private name, across a module boundary. Worth promoting to env_bool now that it has two call sites in different packages.
- The autologin asymmetry runs both ways. The PR calls out that env config clears is_autologin on rows it does not own, but create_oidc_provider (mfa.py:1401) and update_oidc_provider (mfa.py:1498) will equally clear it on the env-managed row, and it only returns at the next boot. One sentence in the wiki covers it.
- Since the target is Kubernetes: two replicas booting together can both miss the name lookup and race the insert. The unique constraint on name turns the loser into a caught exception rather than a duplicate, so it is harmless — but it will print BAMBUDDY_OIDC_* could not be applied: IntegrityError on some boots, which reads more alarming than it is. Worth a note so nobody chases it.

On the four test failures

You are right that they are not yours. They are column oidc_providers.is_env_managed does not exist, and the cause is that create_backup_zip uses backend.app.core.database.engine rather than the in-memory SQLite test session — so it dumps whatever DATABASE_URL points at, which on my machine is a Postgres instance that has not been booted with this branch yet. Your migration is present and correct, and test_oidc_env_managed_migration.py exercises the real ALTER path. Nothing to do here.
'on'      -> require_email_verified=False

munin92:
All in.

Blank `SCOPES`/`EMAIL_CLAIM`/`ICON_URL` now count as unset instead of refusing the provider (`c547505c`). Unrecognized booleans are rejected through the same clean skip-path as a bad `DEFAULT_GROUP`, not silently `False`, and `env_bool` is public now (`77c9bdd6`). That would've made the `BAMBUDDY_LOCAL_LOGIN` bypass 500 on `=on`, so that caller stays lenient (`9e783fba`).

Wiki #55 got the autologin-both-ways and the multi-replica `IntegrityError` notes.

PR #2653, opened 2026-07-24
fix(plate-detection): use Pillow+numpy instead of cv2 (segfaults on musl/Alpine)

Vilda007:
## Problem

On **Alpine/musl** add-ons (notably Home Assistant OS, e.g. Raspberry Pi aarch64), the OpenCV wheel has to be installed via a `gcompat` (glibc→musl) shim because PyPI ships no `musllinux` wheel. Under that shim, OpenCV's **JPEG codec** *and* **image-processing ops** segfault (SIGSEGV, exit 139) on images **≥ ~512px**, while working fine on small arrays.

Reproduced deterministically (same shim, same image):

| Operation | ≤256×256 | ≥512×512 |
| --- | --- | --- |
| `cv2.imdecode` / `cv2.imread` | OK | **SEGV** |
| `cv2.imwrite` / `cv2.imencode` | OK | **SEGV** |
| `cv2.cvtColor` | OK | **SEGV** |
| `cv2.GaussianBlur` / `cv2.normalize` / `cv2.absdiff` | OK | **SEGV** |

None of the usual knobs help: `setNumThreads(0)`, `setUseOptimized(False)`, `OMP_NUM_THREADS=1`, `cv2.ocl.setUseOpenCL(False)` — all still segfault at ≥512px.

Because plate detection runs every image op through cv2, the crash kills the **uvicorn worker mid-request**. In the UI this surfaces as:

- **Calibrate Empty Plate** → `failed to fetch`, reference counter stuck at **0/5**
- no `uvicorn.access` log line for the calibrate POST (response never sent)
- no error/exception log — it's a segfault, not a Python exception (so it's invisible to normal logging and looks like "the request never arrived")

The camera capture itself (chamber-image protocol on TLS 6000) works fine — the failure is purely in OpenCV's image ops.

## Solution

Route **all** image I/O and processing in `plate_detection.py` through **Pillow + numpy**, which ship their own libjpeg and use their own / numpy kernels and handle real camera frames (1280×720) fine. cv2 is still imported so `OPENCV_AVAILABLE` (and the 503 route guard / `is_plate_detection_available()`) keep their existing semantics — but it is no longer called for any image operation. Plate detection's real runtime dependency is now Pillow.

### Changes

- New helpers `_pil_decode` / `_pil_read` / `_pil_encode` / `_pil_resize_bgr` (BGR uint8 convention, drop-ins for `cv2.imdecode` / `cv2.imread` / `cv2.imencode` / `cv2.resize`).
- `_preprocess_for_comparison`: `PIL.Image.convert("L")` + `ImageFilter.GaussianBlur(radius=8)` (≈ cv2 `(51,51)` sigma ≈ 8) + numpy MINMAX normalization, instead of `cv2.cvtColor` / `GaussianBlur` / `normalize`.
- `cv2.absdiff` → `np.abs(a.astype(int16) - b.astype(int16)).astype(uint8)`.
- Debug image: `PIL.ImageDraw` (rectangle, blended red overlay via numpy, text via `ImageFont.truetype("DejaVuSans.ttf", …)` with `load_default()` fallback) + `_pil_encode`, instead of `cv2.rectangle` / `cvtColor` / `addWeighted` / `putText` / `imencode`.
- **`calibrate()` writes the original JPEG bytes verbatim** (`reference_path.write_bytes(image_data)`) instead of `cv2.imdecode` then `cv2.imwrite` at quality 95. The captured `image_data` is already a valid JPEG; decoding only to re-encode added a lossy generation-loss step (and the segfault). `_pil_decode` is still called once to validate the input is decodable (rejects corrupt bytes, mirroring the old `if frame is None` guard) — the decoded array isn't used further. On-disk format is unchanged (plain JPEGs + metadata JSON), so existing calibrations and the analyze path keep working.

### What did NOT change

- HTTP routes (`backend/app/api/routes/camera.py`) — untouched.
- Calibration on-disk format — untouched.
- `OPENCV_AVAILABLE` / `is_plate_detection_available()` semantics — untouched.
- Public method signatures and `PlateDetectionResult` — untouched.

## Verification

End-to-end on HAOS (RPi 5, aarch64, musl, the same gcompat shim that segfaulted):

```
calibrate empty1.jpg -> True  (1/5 references)
calibrate empty2.jpg -> True  (2/5 references)

GET /api/v1/printers/1/camera/check-plate?use_external=false&include_debug_image=true
HTTP 200 in 3.18s
{'is_empty': False, 'confidence': 1.0, 'difference_percent': 8.33,
 'message': 'Objects detected on plate (difference: 8.3%, best ref 1/2)'}
```

No segfault, no `failed to fetch`, counter 2/5. The "Objects detected" result is correct — the printer was mid-print, so the plate was not empty; this validated the whole capture → decode → preprocess → compare → debug-image pipeline that previously crashed.

## Notes / discussion

- The cv2 segfault is **platform-specific** (musl + gcompat shim). On standard glibc installs (the manylinux wheel running natively) cv2 works fine, so this change is a no-op in behavior there — it just uses Pillow instead of cv2. Pillow is already a common dependency; if it isn't currently declared as a hard requirement of the backend, it may need to be added/gated. I'd be happy to adjust based on the maintainer's preferred dependency story.
- An alternative the maintainer may prefer is keeping cv2 as the primary path and falling back to Pillow only on musl (platform detection). I went with the simpler single-path approach because it's what's verified working and keeps one code path; happy to refactor to a conditional if that's preferred.
- A natural follow-up (kept out of this PR to stay minimal): make `OPENCV_AVAILABLE` / `is_plate_detection_available()` depend on Pillow rather than cv2, and drop the cv2 import entirely from this module. Plate detection no longer needs cv2 at all once this lands.

Happy to split/adjust commits as needed. Thanks for the add-on!

### DIFF
--- a/backend/app/services/plate_detection.py
+++ b/backend/app/services/plate_detection.py
@@ -24,6 +24,84 @@ except ImportError:
     logger.info("OpenCV not available - plate detection feature disabled")
 
 
+# --- Pillow + numpy image helpers ------------------------------------------
+# OpenCV's JPEG codec AND image-processing ops (imdecode/imread/imencode/
+# imwrite, cvtColor/GaussianBlur/normalize/absdiff) segfault under the
+# musl/gcompat shim used to run the manylinux cv2 wheel on Alpine/musl add-ons,
+# on images >= ~512px. Pillow ships its own libjpeg and uses its own (or numpy)
+# kernels, handling real camera frames (1280x720) fine, so all image I/O and
+# processing in this module is routed through Pillow + numpy instead of cv2.
+# cv2 is still imported above so the OPENCV_AVAILABLE gate (and the 503 route
+# guard / availability automation) keep their existing semantics, but it is no
+# longer called for any image operation.
+def _pil_decode(image_data: bytes):
+    """Decode JPEG/PNG bytes into a BGR uint8 HxWx3 array (cv2 convention).
+
+    Drop-in replacement for ``cv2.imdecode(np.frombuffer(buf, np.uint8), cv2.IMREAD_COLOR)``.
+    Returns None on failure (mirrors cv2's None-on-failure contract).
+    """
+    try:
+        from PIL import Image
+        import io
+
+        img = Image.open(io.BytesIO(image_data)).convert("RGB")
+        arr = np.asarray(img)  # RGB HxWx3 uint8
+        return arr[:, :, ::-1].copy()  # RGB -> BGR for cv2 consumers
+    except Exception as e:
+        logger.warning("Pillow decode failed: %s", e)
+        return None
+
+
+def _pil_read(path):
+    """Read a JPEG/PNG file into a BGR uint8 HxWx3 array.
+
+    Drop-in replacement for ``cv2.imread(str(path), cv2.IMREAD_COLOR)``.
+    """
+    try:
+        from PIL import Image
+
+        img = Image.open(str(path)).convert("RGB")
+        arr = np.asarray(img)
+        return arr[:, :, ::-1].copy()
+    except Exception as e:
+        logger.warning("Pillow read failed for %s: %s", path, e)
+        return None
+
+
+def _pil_encode(frame_bgr, quality: int = 95) -> bytes:
+    """Encode a BGR uint8 HxWx3 array to JPEG bytes.
+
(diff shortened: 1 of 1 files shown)


maziggy:

Issue #2576, opened 2026-07-16
[Bug Report] For the X2D it has an left and right AUX fan, under printers the setting for the


maziggy:
For the X2D it has an left and right AUX fan, under printers the setting for the AUX fan only controls the left hand side

### Screenshot
![Bug Report Screenshot](https://raw.githubusercontent.com/maziggy/bambuddy/bug-report-assets/screenshots/a5680e68c8644e249c63868c4ee3c20c.png)

<details>
<summary>Reporter Contact</summary>

Email: m_altenroxel@protonmail.com

</details>

<details>
<summary>System Information</summary>

```json
{
  "generated_at": "2026-07-16T04:50:10.249521+00:00",
  "app": {
    "version": "0.2.4.9",
    "debug_mode": false
  },
  "system": {
    "platform": "Linux",
    "platform_release": "4.4.302+",
    "platform_version": "#86009 SMP Wed Nov 26 18:19:20 CST 2025",
    "architecture": "x86_64",
    "python_version": "3.13.14"
  },
  "environment": {
    "docker": true,
    "data_dir": "/app/data",
    "log_dir": "/app/logs",
    "timezone": "Australia/Melbourne"
  },
  "database": {
    "archives_total": 16,
    "archives_completed": 13,
    "printers_total": 2,
    "filaments_total": 0,
    "projects_total": 0,
    "smart_plugs_total": 0
  },
  "printers": [
    {
      "index": 1,
      "model": "X2D",
      "nozzle_count": 2,
      "is_active": true,
      "mqtt_connected": true,
      "state": "RUNNING",
      "firmware_version": "01.01.01.00",
      "wifi_signal": -40,
      "reachable": true,
      "ams_unit_count": 2,
      "ams_tray_count": 3,
      "has_vt_tray": true,
      "external_camera_configured": false,
      "plate_detection_enabled": false,
      "obico_enabled": false,
      "hms_error_count": 1,
      "developer_mode": true,
      "nozzle_rack_count": 2
    },
    {
      "index": 2,
      "model": "A1",
      "nozzle_count": 1,
      "is_active": true,
      "mqtt_connected": true,
      "state": "IDLE",
      "firmware_version": "01.08.01.00",
      "wifi_signal": -69,
      "reachable": true,
      "ams_unit_count": 0,
      "ams_tray_count": 0,
      "has_vt_tray": true,
      "external_camera_configured": false,
      "plate_detection_enabled": false,
      "obico_enabled": false,
[...]

Issue #2525, opened 2026-07-08
[Feature]: MQTT State for printer waits for plate clearance confirmation


daschaefer:
### Problem or Use Case

Hi there, 

At the moment, I'm missing that I don’t get the information via MQTT when a printer is waiting for plate clearance confirmation. It would be helpful for me if this information were included.

An notification option for that would also be very helpful.

Thank you!

### Proposed Solution

Either by expanding the MQTT print status, or by adding another field.

### Feature Category

Print Queue & Scheduling

### Priority

Would improve my workflow

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Please explain more detailed. What do you mean with MQTT?

daschaefer:
Since MQTT support in general is already implemented, and works flawlessly

<img width="541" height="606" alt="Image" src="https://github.com/user-attachments/assets/1b32ba0f-1104-46e3-8866-96dc97960242" />

I guess your question is related to something else.

I heavily rely on MQTT in my automation workflow, because not everybody will have access to the Bambuddy UI. Most of the tasks can be done via a small leightweight selfwritten Frontend to control the basic day to day tasks. The problem is, that I can't identify which printer awaits for plate clearance because this information is not part in the MQTT messages. Therefore I can't display a notification on the selfwritten frontend to make aware that a plate clearance procedure is need to be done.
 
All I get is the basic states like RUNNING, PAUSE, FAILED, FINISH and IDLE. I'm missing something like "WAITING FOR CLEARANCE" or something like that which reflects the plate clearance state on the Bambuddy WebUI.

Basically I use all of the MQTT information as "live data" and interact (sendind commands) with Bambuddy via the API. Which totally works fine, I just need the information given that the printer is waiting for plate clearance to proceed with its tasks.

maziggy:
Got it.

maziggy:
Before we commit to building this, I'd like to gauge community interest. If you'd find this feature useful, please give this issue a thumbs up (👍) reaction so we can prioritize accordingly.

maziggy:
MQTT bambuddy/printers/{serial}/status now includes awaiting_plate_clear.

Available/Fixed in branch dev and available with the next release or daily build. Please let me know if it works for you.

-----
[...]

PR #2691, opened 2026-07-27
feat(printers): expose P2S/X2D accessory fans (left aux + exhaust)

### LINKED ISSUE #2660: [Feature]: Support the P2S/X2D left auxiliary and exhaust accessory fans
### Problem or Use Case

The P2S (with kit addon) and X2D have two fans that Bambuddy doesn't currently expose:

**1. Left auxiliary part cooling fan** — Bambuddy shows a single "Auxiliary" fan (`big_fan1_speed`, driven with `M106 P2`), which on these printers is only the *right* aux fan. The left one has no field and no control. This is the gap reported in #2576 ("For the X2D it has an left and right AUX fan, under printers the setting for the AUX fan only controls the left hand side"), which was closed after the original reporter didn't follow up with the capture you asked for.

**2. Chamber exhaust fan** — the `big_fan2` badge shows on every P2S labelled "Chamber Fan". On the P2S the exhaust fan is actually an add-on kit, so the badge is dead UI on a base machine, and Bambu's own firmware/UI calls it "Exhaust" on P2S/X2D rather than "Chamber".

On the P2S both fans are purchasable kits; on the X2D they ship from the factory.

I have a P2S with both kits fitted and I've captured the MQTT for all of this — including the capture requested in #2576.

### Proposed Solution

Detect and expose both fans from `device.airduct.parts`, which only lists the fans that physically exist, so each badge appears only when the hardware is present:

- **Left aux fan** = airduct part id **10** (raw id 160 >> 4 — `FAN_REMOTE_COOLING_1` in Bambu Studio's `DevFan::ParseV3_0`). Crucially it is reported **only** in the airduct and is never mirrored into any flat `big_fanX_speed` field, which is exactly why it's currently invisible. Control it with `M106 P10` — the same command Bambu's official P2S machine profiles use in their start / layer-change gcode.
- **Exhaust fan** = airduct part id **3** (`big_fan2_speed`); present only with the kit, and additionally reported as a `get_version` hardware module (`eef` = "Bambu Lab External Exhaust Fan").

Verified fan → field map from my P2S (fw 01.02.00.00), stable across cooling and heating airduct modes:

| Fan | Field | airduct id | Base P2S |
|---|---|---|---|
| Part cooling | `cooling_fan_speed` | 1 | built-in |
| Aux (right; a flap re-tasks it between part-cooling and chamber-filter by mode) | `big_fan1_speed` | 2 | built-in |
| Exhaust | `big_fan2_speed` | 3 | kit |
| Left aux | *(airduct only — no flat field)* | 10 | kit |

I mapped this empirically by toggling each fan individually on the printer's touchscreen while watching MQTT, in both cooling and heating modes.

### Alternatives Considered

- Reading the flat `big_fanX_speed` fields only — doesn't work, the left aux fan simply isn't there.
- Gating the badges on printer model alone — wrong for the P2S, where both fans are optional kits, so the badges would be dead UI on a base machine. Presence-gating on the airduct parts list handles base / kitted / X2D uniformly.

### Feature Category

UI/UX

### Priority

Would improve my workflow

### Mockups or Examples

_No response_

### Contribution

- [X] I would be willing to help implement this feature

### Checklist

- [X] I have searched existing issues to ensure this feature hasn't already been requested

---

**Note to maintainer:** I already have this implemented and tested against my live P2S — branched from `dev`, with unit/integration tests, i18n for all 12 locales, and `ruff` / `pytest` / `npx tsc` / vitest / i18n-parity all passing, plus companion wiki docs ready. Per CONTRIBUTING.md I'm raising this issue first and waiting to be assigned before opening the PR. Happy to adjust scope or naming (e.g. renaming the existing "Auxiliary" badge to "Right Auxiliary") however you prefer.

gzimbric:
## Description

Adds support for two fans on the P2S/X2D that Bambuddy doesn't currently expose: the **left auxiliary cooling fan** (no field and no control today) and the **chamber exhaust fan** (currently shown on every P2S labelled "Chamber Fan", even on machines without the kit). On the P2S both are add-on kits; on the X2D they ship from the factory. Both are detected from the printer's airduct report, so each badge appears only when the hardware is actually present.

## Related Issue

Fixes #2660

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#57

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition or update

## Changes Made

- Added the **left auxiliary part cooling fan**: parsed from `device.airduct` part id 10, exposed as `left_aux_fan_speed`, and controllable through a new `fan=aux2` option on `POST /printers/{id}/fan-speed` (sends `M106 P10`, the command Bambu's official P2S machine profiles use). Its badge shows only when the printer reports the fan.
- Added `exhaust_fan_present` (airduct part id 3) and used it to presence-gate the `big_fan2` badge on P2S/X2D, so a base P2S without the exhaust kit no longer shows a dead tile. The badge is labelled **Exhaust** on P2S/X2D to match the printer's own screen and Bambu Studio; other enclosed models (X1/P1S/H2*) keep the always-present **Chamber Fan** badge unchanged.
- Made the fan-speed response message model-aware via a new `uses_exhaust_fan_label()` helper in `printer_models`, so the success toast says "Exhaust fan set to N%" on P2S/X2D rather than contradicting the badge the user just clicked.
- Ordered the badges part cooling → left aux → aux → chamber/exhaust so the two auxiliary badges read left-to-right in the same order as the physical hardware.
- Surfaced both new fields through the status endpoint, WebSocket broadcasts, and the MQTT relay.
- Added `leftAuxiliary` and `exhaust` translation keys to all 12 locales.
- Added unit and integration tests for airduct part parsing (id 10 speed, id 3 presence, base-P2S absence, malformed entries, diff-push survival), the `aux2 → M106 P10` mapping, the model-aware response message, and badge visibility/ordering/labelling.

## Screenshots

**Left auxiliary fan — new, controllable via the standard speed popover:**

Set Left Auxiliary Fan Speed popover on a P2S printer card

**Exhaust fan — the `big_fan2` badge, labelled "Exhaust" on P2S/X2D:**

Set Exhaust Speed popover on a P2S printer card

**Full fan row — part cooling, left aux, aux, exhaust in physical left-to-right order:**

P2S printer card fan row showing four fan badges

*(Printer name redacted. On a base P2S without the kits, the left aux and exhaust badges are simply absent — the row looks exactly as it does today.)*

## Testing

- [x] I have tested with my printer model: P2S (fw 01.02.00.00, with both accessory kits fitted)

Tested with:
- `./test_backend.sh` — `ruff check` + `ruff format --check` clean (643 files); pytest on the touched suites (`test_p2s_accessory_fans`, `test_printers_api`, `test_printer_manager_status_broadcast`) — **all passing**
- `npx tsc` — clean, exit 0
- `npm run test:run` — PrintersPage suite **67 passed**; `npm run check:i18n` — all 12 locales in parity
- `eslint` — clean on all touched files
- **Live printer:** toggled each fan individually on the P2S touchscreen in both cooling and heating airduct modes while capturing MQTT, to derive the fan → field map below. Verified control end-to-end: `fan=aux2` set the left aux fan to 50% and back to 0%, confirmed in the printer's own reports.

## Checklist

## Additional Notes

**How detection works.** `device.airduct.parts` lists only the fans that physically exist (decoded id = `raw_id >> 4`, mirroring Bambu Studio's `DevFan::ParseV3_0`), so it doubles as a per-kit presence signal. The exhaust kit additionally appears as a `get_version` hardware module (`eef` = "Bambu Lab External Exhaust Fan"). Verified map, stable across cooling and heating airduct modes:

| Fan | Field | airduct id | Base P2S |
|---|---|---|---|
| Part cooling | `cooling_fan_speed` | 1 | built-in |
| Aux (right; a flap re-tasks it between part-cooling and chamber-filter by mode) | `big_fan1_speed` | 2 | built-in |
| Exhaust | `big_fan2_speed` | 3 | kit |
| Left aux | *(airduct only — no flat field)* | 10 | kit |

The key finding for #2576 / #2660: toggling the left aux fan moves **only** airduct part 10 and no flat `big_fanX_speed` field, which is exactly why it was invisible.

**Naming.** Per Bambu's P2S/N7 fan config, part 2 is "Right(Aux)" and part 10 is "Left(Aux)". This PR leaves the existing "Auxiliary" badge as-is (part 2) and adds the new one as "Left Auxiliary Fan" — happy to rename either (e.g. to "Right Auxiliary") if you'd prefer the pair spelled out.

**X2D.** The X2D ships with both fans from the factory, so both badges should always appear there; the presence-based detection handles that the same way as a fully-kitted P2S. I only have a P2S to verify against, so confirmation from an X2D owner would be welcome.

**Absent-hardware case.** Verified present on my kitted machine and simulated for the absent case (airduct parts `[1,2]` only → both fans correctly report as not installed). If you have a base P2S handy, a quick confirm that it omits parts 3 and 10 would fully close the loop.


maziggy (changes requested):
This is one of the best-prepared PRs this project has received — thank you. I verified the protocol claims against Bambu Studio's DevFan sources and ran the full backend and frontend suites locally: everything checks out, including the part id decode, the AIR_FUN ids, and the part-3-presence gate (which is exactly how Studio detects the chamber fan itself).

Two things before I merge:

1. Please mask the part state to the low 8 bits before clamping, i.e. int(part["state"]) & 0xFF. Studio decodes it with get_flag_bits(state, 0, 8), and since the firmware bit-packs sibling fields (range is end<<16|start), a packed state value would currently clamp to 100 instead of decoding correctly.

2. The Left/Right question on the X2D: the original #2576 reporter described the existing aux control as moving the left fan on their X2D, which conflicts with the P2S mapping (part 2 = Right, part 10 = Left). Since you flagged this yourself, I'll ask the X2D owner from #2576 to test this branch and confirm which physical fan each badge drives. If the X2D turns out mirrored, we'll adjust the labels — no functional change needed.

Cosmetic nit: move the uses_exhaust_fan_label import in printers.py to the top of the file with the other imports.

Thanks again!

gzimbric:
Thanks for the thorough review and for verifying against the Studio sources.

**1. Low-8-bit mask — fixed.** You're right, and it's a real decode bug rather than a cosmetic one. `state` is packed the same way `range` is, so a packed value was clamping to 100 instead of decoding. Now `int(part["state"]) & 0xFF` before the clamp, matching `get_flag_bits(state, 0, 8)`. Two tests cover it: a packed `(60 << 16) | 45` now decodes to `45` (it returned `100` before), and plain 0–100 values round-trip unchanged. My P2S only ever sends unpacked values, so this wouldn't have shown up on my hardware — good catch.

**3. Import — moved** to the top of `printers.py` with the other `backend.app.utils` imports.

**2. X2D Left/Right — agreed, and thank you for lining up a tester.** Worth flagging that the conflict may be in the *reporter's description* rather than the mapping: on my P2S the existing "Auxiliary" badge (`big_fan1` / `M106 P2`) drives the **right** fan, and Bambu's own N7 fan config names part 2 `Right(Aux)` and part 10 `Left(Aux)`. If the X2D shares that config, the badges are already correct and #2576's "only controls the left hand side" was just a naming slip. If the X2D genuinely mirrors it, the fix is label-only as you say. Something that might make the test unambiguous: with only one fan running, `device.airduct.parts` reports a non-zero `state` for exactly one id — so the tester can watch which physical fan spins for id 2 vs id 10 without touching the UI labels.

Also, while rebasing onto the current `dev` I picked up two things worth mentioning:
- Resolved a conflict in `mqtt_relay.py` where `awaiting_plate_clear` (#2525) landed in the same block — kept both, with the fan fields grouped alongside the other fan entries.
- The new Ukrainian locale (#2695) arrived mid-review, so I added `leftAuxiliary` / `exhaust` to `uk.ts`; all 13 locales are back in parity.

Re-verified after the rebase: `ruff check` + `format --check` clean, backend suites **226 passed**, `npx tsc` clean, PrintersPage **67 passed**, i18n parity green.

gzimbric:
### On the X2D Left/Right question — the support bundle on #2576 may already answer it

@kpp39 posted a bundle on #2576 after your review. It doesn't contain an `airduct.parts` dump (Bambuddy only logs `device.keys` today, not the parts array), but the one-shot fan snapshot in it is suggestive:

```
[SERIAL] Fan fields in MQTT data: {'aux_part_fan': False, 'big_fan1_speed': '5',
 'big_fan2_speed': '0', 'cooling_fan_speed': '0', 'fan_gear': 26112, 'heatbreak_fan_speed': '0'}
```

Decoding `fan_gear` (26112 = `0x6600`) the way `DevFan::ParseV1_0` does:

| Field | Raw | % |
|---|---|---|
| `cooling_fan_speed` = `fg & 0xFF` | 0 | 0% |
| **`big_fan1_speed` = `(fg >> 8) & 0xFF`** | **102** | **40.0%** |
| `big_fan2_speed` = `(fg >> 16) & 0xFF` | 0 | 0% |

Your instructions were *"set the left AUX fan to one value (say 40%) and the right AUX fan to a clearly different value (say 90%)"*. `big_fan1` reads **exactly 40%** (102/255 to the digit), and **90% appears in no flat field at all** — consistent with the second fan living only in the airduct, exactly as part 10 does on the P2S.

If the reporter followed the 40/90 instruction, that means:

| | P2S (verified on my hardware) | X2D (inferred here) |
|---|---|---|
| `big_fan1` / part 2 / `M106 P2` | Right | **Left** |
| airduct-only / part 10 / `M106 P10` | Left | **Right** |

i.e. **mirrored**, which also independently matches the original #2576 complaint that Bambuddy's Aux control (`M106 P2`) moves the *left* fan on an X2D.

**Caveats, because this is inference rather than proof:** it's a single snapshot; the 90% fan isn't positively identified, only inferred from its absence; and `big_fan1_speed: '5'` on the coarse 0–15 scale reads as 33%, so I'm treating `fan_gear` as the precise source. I wouldn't relabel on this alone.

**Cheap way to settle it conclusively:** this branch already parses `device.airduct.parts`, so a fresh capture from @kpp39 running it would show each part's `state` directly — with the two fans at clearly different values, whichever id reads 90% is the right-hand fan, no UI labels involved. Happy to add a temporary one-shot debug line that dumps the decoded parts array if that would make the capture easier to read.

As you said, this is label-only either way — the part-id plumbing doesn't change. If it does land mirrored, my instinct is to gate the two aux labels on model rather than rename globally, so the P2S keeps matching Bambu's own N7 fan config (`2 = Right(Aux)`, `10 = Left(Aux)`) while the X2D gets the reversed pair.

maziggy (changes requested):
Round-1 items 1 and 3 look good - the packed-state decode is right, and the pair of tests around it (packed value decodes to 45, plain 0-100 values round-trip) pins both directions. One thing to fix, and then some data that answers three of your open questions.

Blocking: two existing tests fail on this branch.

FAILED test_plate_clear_mqtt_notification.py::TestStatusPayload::test_status_payload_carries_awaiting_plate_clear

FAILED test_plate_clear_mqtt_notification.py::TestStatusPayload::test_status_payload_defaults_to_not_awaiting backend/app/services/mqtt_relay.py:285: AttributeError

mqtt_relay now reads state.left_aux_fan_speed, but the SimpleNamespace fixture at test_plate_clear_mqtt_notification.py:29 enumerates its fields explicitly and does not have it. You updated the equivalent fixture in test_printer_manager_status_broadcast.py and missed this one - which fits "backend suites 226 passed", i.e. only the three touched suites. Adding left_aux_fan_speed=None and exhaust_fan_present=False to _state() fixes it. Full unit run here is 2 failed / 5056 passed; both pass on dev.

Now the good part. I went through all 247 support packages in our archive. 37 contain an airduct block, and every single one carries a parts list:

P2S (base)          2 bundles   parts 1, 2
P2S (exhaust kit)   2 bundles   parts 1, 2, 3
X2D                 2 bundles   parts 1, 2, 3, 10
H2C / H2D / H2S    16 bundles   parts 1, 2, 3, 6

That closes three of your open items:

- "If you have a base P2S handy, a quick confirm that it omits parts 3 and 10"
  - confirmed on real hardware. Base P2S lists only 1 and 2.
- "Confirmation from an X2D owner would be welcome" - confirmed. X2D reports 1, 2, 3 and 10 from the factory, so both badges appear exactly as you predicted.
- I also went looking for the failure mode where a P2S/X2D reports airduct but no parts array, which would have silently dropped the exhaust badge. It does not exist in the archive: parts is present in 37 of 37.

On Left/Right, one dead end worth saving you the trip. The func field looked promising - your P2S capture tags part 2 / part 10 as func 6 / func 5, and the X2D bundles tag them 5 / 6, the exact mirror. It does not hold up:

P2S base           (1,0) (2,6)
P2S + exhaust kit  (1,0) (2,8) (3,2)
P2S + both kits    (1,0) (2,6) (3,2) (10,5)   <- your capture
X2D                (1,0) (2,5) (3,2) (10,6)
H2C/H2D/H2S        (1,0) (2,1) (3,2) (6,4)

Part 2 carries func 6, 8 and 5 depending on model and fitted kit, so it is a configuration-dependent mask, not an identity tag.

Your fan_gear arithmetic does check out, by the way - 0x6600 gives big_fan1 = 102/255 = 40.0% to the digit. But note the archive already has X2D parts arrays, so a capture from this branch is not what is missing. What is missing is one with the two aux fans at clearly different speeds; both X2D bundles have them at rest or equal (80/80). Worth asking @kpp39 for exactly that.

Smaller things:

- exhaust_fan_present also goes true on H2C/H2D/H2S, since they report part 3 too. Harmless because the frontend only consults it for P2S/X2D, but the name promises something model-specific that the field is not. A docstring note would do.
- H2 series reports decoded id 6, which nothing maps. Not needed here, just flagging for whoever extends the parser.
- int(part["id"]) >> 4 is not masked while state now is. If state wanted  get_flag_bits(state, 0, 8), id presumably wants get_flag_bits(id, 4, 8). Every raw id in the archive (16/32/48/160) decodes fine either way, so this is consistency rather than a bug.
- fan=aux2 is accepted for every model, so a POST against an A1 sends M106 P10. The docstring says the firmware silently ignores it otherwise, which was not tested - you have a kitted P2S, not a bare one. Rejecting aux2 when the printer reports left_aux_fan_speed as null would close that off. API surface only; the UI already hides the badge.
- The exhaust-label model list exists twice - EXHAUST_FAN_LABEL_MODELS with N7/N6 on the backend, MODELS_WITH_EXHAUST_LABEL with display names on the   frontend. Correct as written since printer.model is a display name there, but nothing asserts the two agree.

gzimbric:
Fixed, and thank you for the archive dig — that's a much better answer than anything I could get from one printer.

**Blocking: fixed.** You diagnosed it exactly right. I updated the `_state()` fixture in `test_printer_manager_status_broadcast.py` and missed the sibling in `test_plate_clear_mqtt_notification.py`, and running only the touched suites is precisely what hid it. Added `left_aux_fan_speed=None` and `exhaust_fan_present=False`. Full `backend/tests/unit` run here: **5031 passed**, with both plate-clear status-payload tests green. Lesson taken — "the suites I touched" is not the same as "the suites my change reaches", and a field read from a shared state object reaches further than the diff suggests.

**On the archive findings.** Those close out my three open questions properly, and the 37-of-37 result on `parts` being present is more reassuring than the absent-hardware case I could only simulate. I've left the parsing as-is since it was already correct for all five layouts.

**`func` — agreed, dead end.** Thanks for actually chasing it. Part 2 carrying func 6/8/5 depending on model and fitted kit settles it: it's a configuration mask, not an identity tag. Worth noting for posterity that it looked *very* convincing on two data points.

**Left/Right on the X2D.** Understood — the archive already has the parts arrays, so what's missing is a capture with the two aux fans at *clearly different* speeds, not a capture from this branch. Both existing bundles being at rest or 80/80 is the problem. I'll ask @kpp39 on #2576 for exactly that, with explicit "set one to 40% and the other to 90%, then grab a bundle" instructions. Until that lands I've changed no labels.

**The smaller items, all addressed:**

- **`exhaust_fan_present` naming.** Docstring now states plainly that H2C/H2D/H2S also report part 3 so the flag goes True there, that it's harmless because only the P2S/X2D badge consults it, and that it must not be read as "an exhaust kit is fitted" without also checking the model. You're right that the name promises something the field doesn't deliver — I'd rather fix the docstring than rename a field that's already in the status payload, but say the word if you'd prefer the rename.
- **Id masking.** Now `(int(part["id"]) >> 4) & 0xFF`, matching `get_flag_bits(id, 4, 8)`. Comment records that every archive id (16/32/48/160) decodes identically either way, so this is consistency rather than a fix.
- **H2 id 6.** Flagged in a comment next to the id branches, with the full observed id map, for whoever extends the parser.
- **`fan=aux2` on printers without the fan.** Now rejected with a 400 when the printer reports no `left_aux_fan_speed`. You were right that I hadn't tested it — I have a kitted P2S, so "the firmware silently ignores it" was an assumption I had no way to check. Gating on reported presence means I don't need it to be true. Two tests: an A1 with `left_aux_fan_speed=None` gets a 400 and `set_fan_speed` is never called, and a base P2S can still drive part/aux/chamber unaffected. I also made `test_fan_id_mapping` set the attribute explicitly, since it was previously satisfying the new gate by accident via MagicMock's auto-attributes.
- **The duplicated model list.** Added `TestExhaustLabelModelListsAgree`, which parses `MODELS_WITH_EXHAUST_LABEL` out of `PrintersPage.tsx` and asserts it against `EXHAUST_FAN_LABEL_MODELS` in both directions — frontend entries must exist on the backend, and backend display names (excluding the N7/N6 internal codes) must be handled by the frontend. The failure messages name the drift explicitly, e.g. *"labels the fan 'Exhaust' in the UI but the backend would report 'Chamber fan'"*. It skips rather than fails if the frontend sources aren't in the checkout.

Re-verified: `ruff check` + `ruff format --check` clean (650 files), `backend/tests/unit` 5031 passed, accessory-fan and plate-clear suites green.

kpp39:
Yo guys you need a support package from me? 
Btw last time i did that i occasionally set 100% instead of 90%

maziggy (approved):

PR #2693, opened 2026-07-28
feat: show live print progress in the browser tab title

### LINKED ISSUE #1041: [Feature]: Show print percentage status in Chrome/Firefox tab name
### Problem or Use Case

I'm always having to open the bambuddy page in order to view current prints' completion percentage.

### Proposed Solution

Would be very handy to have the current prints percentage completion shown in the chrome (in my case) tab name. 
Could be a selection in the options to enable/disable as well as select which printer it is looking at. Could further allow it to blink/flash if there is an error.

### Alternatives Considered

_No response_

### Feature Category

Print Archiving

### Priority

Nice to have

### Mockups or Examples

_No response_

### Contribution

### Checklist

Chachigo:
## Description

Adds an optional **print progress in the browser tab** feature (off by default), toggled under **Settings → Appearance**. When enabled, the tab title shows the soonest-finishing running print's completion percentage (e.g. `42% · Bambuddy`) and the favicon becomes a small green progress ring, updated live over the existing WebSocket. Defaults are restored when nothing is printing or the toggle is off. The preference is per-browser (localStorage) — consistent with the existing dark/light **mode** toggle, which is also stored client-side.

## Related Issue

Fixes #1041

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#58

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

## Changes Made

- `usePrintProgressTitle` hook + unit tests for the print-selection logic
- `progressInTitle` preference in `ThemeContext` (localStorage, off by default)
- Toggle in the Settings → Appearance card
- i18n strings for all 12 locales
- README feature entry

## Screenshots

- Browser tab: default favicon + title → `42% · Bambuddy` with the green progress-ring favicon
- The new toggle in Settings → Appearance
Capture d&#39;écran_20260728_111045
Capture d&#39;écran_20260728_121910

Capture d&#39;écran_20260728_124342

## Testing

- [x] I have tested with my printer model:  A1 Mini

Frontend checks pass locally: `tsc`, `eslint`, and Vitest (including the i18n locale-parity check).

## Checklist


maziggy:
What happens if i have more than one printer?

Chachigo:
It shows the print finishing soonest (across all printers).

maziggy (changes requested):
Thanks for this - clean implementation, and I like that it is off by default and reuses the existing WebSocket cache rather than adding a new feed. tsc, eslint, the build and the full frontend suite (2493 tests) all pass here. Two things need fixing before merge.

1. Ukrainian is missing the new keys. uk.ts landed on dev after you opened this, so the dev merges brought the file in but not the strings. npm run check:i18n currently fails with settings.progressInTitle and settings.progressInTitleDescription missing from uk (5681 vs 5683 leaves). Please add a real Ukrainian translation rather than the English text.

2. A remaining_time of 0 wins the selection. The backend defaults remaining_time to 0 and serialises it as-is, so "ETA not known yet" reaches the UI as 0, not null. Since pickActivePrint does `s.remaining_time ?? Infinity`, that 0 sorts as the soonest-finishing print and always wins. So a printer that just started takes the tab away from one that is three minutes from done, and the tab shows 0%. The equivalent logic on the Printers page guards this with `remaining_time != null && remaining_time > 0` - please do the same and treat anything <= 0 as unknown.

One more I would like in the same round: the effect writes document.title even when the toggle is off, because pct is null and the deps are [pct]. Nothing else under WebSocketProvider sets the title today so there is no visible regression, but it means an off-by-default feature permanently owns the tab title. An early if (!progressInTitle) return;` in both effects keeps it inert until enabled.

Smaller things: the tests only cover pickActivePrint - the title half of the effect is testable in jsdom, and a "no-op when the pref is off" test would lock the point above down. The test also uses 'PAUSED' as the paused state; the real value is 'PAUSE', so that assertion passes for the wrong reason. And the ring is hardcoded green while the app has six selectable accents and the hook already calls useTheme() - worth wiring now, since all 12 locale strings say "green" and changing it later means re-translating.

Chachigo:
Thanks for the thorough review! All addressed:

1. **Ukrainian** — rebased on latest dev to pick up uk.ts and added a real Ukrainian translation. check:i18n is green (13 locales at parity).
2. **remaining_time <= 0** — added an `eta()` helper that treats anything <= 0 (and null) as unknown, matching the Printers page guard, so a just-started print no longer steals the tab. Added a test for it.
3. **Inert when off** — reworked the effect around an `ownsRef` flag: it never touches the tab while the pref is off, and only restores the title/favicon if it previously took over (on disable, idle, or unmount).
4. **Tests** — added "inert while off" and "shows % when enabled" tests (renderHook + jsdom) on top of the pickActivePrint cases.
5. **'PAUSED'** — fixed to 'PAUSE'.
6. **Ring colour** — now reads the `--accent` CSS variable, so it follows the selected accent (and dark/light). Dropped "green" from all 13 locale strings, the README, and the wiki PR.

tsc, eslint, check:i18n and the full frontend suite pass locally.

maziggy (approved):

Issue #2589, opened 2026-07-17
[Bug]: Unresolved ams_mapping [-1] silently forces P1S queue job to empty external spool


Jostxxl:
### Description

A P1S queue item was dispatched with `use_ams=false` even though:

- a regular AMS was connected,
- two compatible PETG spools were detected,
- the queue item still had `use_ams=true`,
- the operator had **not** enabled strict/forced color matching, and
- no filament was loaded on the external spool holder.

The queue item contained `ams_mapping=[-1]`. Bambuddy treated that unresolved slot as an all-external mapping and silently forced `use_ams=false`. The printer then started against the empty external feed and paused with filament-runout HMS `0300_8004`.

### Live reproduction

Printer:

- Name: `135_P1S_691_Nigel`
- Bambuddy printer ID: 82
- Model: P1S
- Build: `fix/issue-2572` at `bab4c1e1`
- Queue item: 92
- Archive: 285
- File/plate: `Turm 2x_plate_12.gcode.3mf`, plate 12

File requirement:

```json
{
  "slot_id": 1,
  "used_g": 141.86,
  "type": "PETG",
  "color": "#161616"
}
```

AMS state at print start:

- AMS 0 / physical slot 1: PETG, `#BCBCBC`
- AMS 0 / physical slot 2: PETG, `#FFFFFF`
- external tray: empty
- strict “match/consider color” option was not selected

Without strict color matching, either PETG tray should have been a valid type-only match.

The persisted queue row was:

```text
id=92
printer_id=82
use_ams=true
ams_mapping=[-1]
plate_id=12
status=printing
```

### Dispatch evidence

At 2026-07-16 18:48:42 the service logged:

```text
Registered expected print: printer=82, file=Turm_2x_plate_12.3mf,
archive=285, ams_mapping=[-1], plate_id=12

[01P00A452600691] All filament slots use external spool — setting use_ams=False
```

The actual MQTT print command then contained:

```json
{
  "use_ams": false,
  "ams_mapping": [-1],
  "ams_mapping2": [{"ams_id": 255, "slot_id": 255}]
}
```

Nine seconds later, the print-start status showed both compatible PETG trays in the AMS but no loaded filament:

```text
PRINT START printer 82: mapping=None, tray_now=255, last_loaded_tray=-1
AMS 0: T0(type=PETG, color=BCBCBCFF), T1(type=PETG, color=FFFFFFFF)
```

The printer paused shortly afterward and currently reports:

[...]

Issue #1780, opened 2026-06-19
[Bug]: H2C nozzle selection not being preserved.


mkoreen:
### Product

Bambuddy

### Area

Print Queue / scheduling

### Bug Description

When sending prints to an H2C virtual printer via bambu studio and queuing the print, bambuddy does not seem to preserve the selected nozzle data from the print.  After turning on debug mode, printer was on R4, I sent a print telling it to use R2, that first print actually did choose R2, but likely that was just luck, R1/R2 are both high flow nozzles (the rest aren't) since the print was high flow, it had to choose R1 or R2 and it chose R2.  

Since this was unfortunately what I didn't want here to show the bug, I cancelled the print and tried again, this time selecting R1 in bambu studio, sent off the print, and it did not even try to switch nozzles, just used R2 again.  

Having done this a few times now, it appears to just select the last nozzle in the set that matches the correct type (HF, Standard, or Nozzle size).  Earlier I did a standard flow print where I told it to use R3 and it chose R4.

[bambuddy-support-20260619-030722.zip](https://github.com/user-attachments/files/29132748/bambuddy-support-20260619-030722.zip)

<img width="696" height="682" alt="Image" src="https://github.com/user-attachments/assets/38f9f0c1-8413-4c17-8137-74a873e306ae" />

Please excuse my not using Host mode, we've had this discussion already.

### Expected Behavior

I expected it to preserve the nozzle I chose in the slicer, (or alternatively let me actually set the nozzle in the "edit print" either in queue or archives.

### Steps to Reproduce

Have H2C, have Virtual printer H2C, send print with bambu studio and choose which right nozzle.

### Troubleshooting steps already taken

Obviously checked the forums, when I didn't see anything from searching, asked on discord, Martin asked me to create an issue.

### Printer Model

H2C

### Bambuddy Version

0.2.5b1 daily probably from a couple days ago.

### Printer Firmware Version

01.02.00.00

### Installation Method

Docker

### Operating System

Linux (Other)

### Support Package (.zip) and / or relevant logs

[bambuddy-support-20260619-030722.zip](https://github.com/user-attachments/files/29132901/bambuddy-support-20260619-030722.zip)

### Screenshots

<img width="696" height="682" alt="Image" src="https://github.com/user-attachments/assets/a0287531-736c-4563-af99-79e70c5ef0b3" />

### Final checks

- [x] I searched existing (open AND closed) issues and this bug hasn't already been reported or resolved
[...]

Issue #1308, opened 2026-05-12
[Bug Report] The ams assignment during a queue is based only on the arrangement of the reels 


maziggy:
The ams assignment during a queue is based only on the arrangement of the reels at the time of queuing, not at the time of the print launch, which means that if empty reels are changed/replaced via another ams slot, a reel of another color can be taken, recalculating the ams positions at the time of launch will solve this problem.

<details>
<summary>Reporter Contact</summary>

Email: 

</details>

<details>
<summary>System Information</summary>

```json
{
  "generated_at": "2026-05-13T00:21:22.878872",
  "app": {
    "version": "0.2.5b1",
    "debug_mode": false
  },
  "system": {
    "platform": "Linux",
    "platform_release": "6.12.86+deb13-amd64",
    "platform_version": "#1 SMP PREEMPT_DYNAMIC Debian 6.12.86-1 (2026-05-08)",
    "architecture": "x86_64",
    "python_version": "3.13.13"
  },
  "environment": {
    "docker": true,
    "data_dir": "/app/data",
    "log_dir": "/app/logs",
    "timezone": "Europe/Berlin"
  },
  "database": {
    "archives_total": 14,
    "archives_completed": 11,
    "printers_total": 1,
    "filaments_total": 0,
    "projects_total": 0,
    "smart_plugs_total": 1
  },
  "printers": [
    {
      "index": 1,
      "model": "X2D",
      "nozzle_count": 2,
      "is_active": true,
      "mqtt_connected": true,
      "state": "RUNNING",
      "firmware_version": "01.01.00.00",
      "wifi_signal": -35,
      "reachable": true,
      "ams_unit_count": 3,
      "ams_tray_count": 7,
      "has_vt_tray": true,
      "external_camera_configured": false,
      "plate_detection_enabled": false,
      "hms_error_count": 4,
      "developer_mode": true,
      "nozzle_rack_count": 2
    }
  ],
  "settings": {
    "advanced_auth_enabled": "true",
    "smtp_auth_enabled": "true",
    "auth_enabled": "true",
    "setup_completed": "true",
    "auto_archive": "true",
    "save_thumbnails": "true",
    "capture_finish_photo": "true",
    "default_filament_cost": "12.0",
    "currency": "EUR",
    "energy_cost_per_kwh": "0.2",
    "energy_tracking_mode": "total",
    "disable_filament_warnings": "false",
    "prefer_lowest_filament": "true",
    "check_updates": "true",
[...]

Issue #1766, opened 2026-06-17
[Bug Report] "Prefer lowest remaining filament" Doesn't work.

After enabling "Prefer lowest 


maziggy:
"Prefer lowest remaining filament" Doesn't work.

After enabling "Prefer lowest remaining filament", when I start a print it still selects the first spool instead of the second one which is the exact same material, manufacturer and color and has less remaining filament.

### Screenshot
![Bug Report Screenshot](https://raw.githubusercontent.com/maziggy/bambuddy/bug-report-assets/screenshots/c8a4d851571c47ec9aeee69baef06218.png)

<details>
<summary>Reporter Contact</summary>

Email: biduleman@gmail.com

</details>

<details>
<summary>System Information</summary>

```json
{
  "generated_at": "2026-06-17T18:59:12.261903+00:00",
  "app": {
    "version": "0.2.4.7",
    "debug_mode": false
  },
  "system": {
    "platform": "Linux",
    "platform_release": "6.12.24-Unraid",
    "platform_version": "#1 SMP PREEMPT_DYNAMIC Sat May  3 00:12:52 PDT 2025",
    "architecture": "x86_64",
    "python_version": "3.13.14"
  },
  "environment": {
    "docker": true,
    "data_dir": "/app/data",
    "log_dir": "/app/logs",
    "timezone": "America/Toronto"
  },
  "database": {
    "archives_total": 6,
    "archives_completed": 3,
    "printers_total": 1,
    "filaments_total": 0,
    "projects_total": 0,
    "smart_plugs_total": 0
  },
  "printers": [
    {
      "index": 1,
      "model": "P1S",
      "nozzle_count": 1,
      "is_active": true,
      "mqtt_connected": true,
      "state": "RUNNING",
      "firmware_version": "01.10.00.00",
      "wifi_signal": -48,
      "reachable": true,
      "ams_unit_count": 1,
      "ams_tray_count": 4,
      "has_vt_tray": true,
      "external_camera_configured": false,
      "plate_detection_enabled": false,
      "obico_enabled": false,
      "hms_error_count": 0,
      "developer_mode": true,
      "nozzle_rack_count": 0
    }
  ],
  "settings": {
    "advanced_auth_enabled": "false",
    "smtp_auth_enabled": "true",
    "auth_enabled": "true",
    "setup_completed": "true",
    "auto_archive": "true",
    "save_thumbnails": "true",
    "capture_finish_photo": "true",
    "default_filament_cost": "15.0",
[...]

PR #2700, opened 2026-07-28
Feature/save ams mapping toggle

### LINKED ISSUE #2684: [Feature]: per-Virtual-Printer 'Save AMS mapping' toggle + reprint auto-apply from saved mapping
### Problem or Use Case

When a print is queued through a Virtual Printer from a slicer (e.g. OrcaSlicer/Bambu Studio), the slicer's AMS/tray-to-filament mapping is currently not preserved anywhere. When the resulting archive entry is later reprinted, the user has to manually re-map every filament slot again, even though the original slicer mapping is still known at print time and is very likely still correct (same physical spools/AMS layout). This gets repetitive for anyone who reprints the same models regularly.

### Proposed Solution

- Add a new per-Virtual-Printer setting, `save_ams_mapping` (bool, default off), configurable from the Virtual Printer card in the frontend.
- When enabled, persist the slicer's AMS mapping (`slicer_ams_mapping`, array indexed by slicer tool index) into `archive.extra_data` at the time the print is added to the queue / restamped.
- The mapping is written only while the toggle is on, but is *read* unconditionally on reprint — turning the toggle off later does not erase previously saved mappings, it only stops recording new ones.
- On the reprint / queue-edit filament-mapping screen, show a "Mapping" toggle button whenever a saved mapping exists for the archive. Clicking it snaps every filament-slot assignment straight to the saved mapping (matched by the slicer's tool index against the target printer's AMS layout), instead of the existing color/material best-effort auto-match. Clicking again reverts to the auto-match result.
- Show an "AMS mapping saved" badge on archive entries that have a stored mapping, so it's visible before opening the reprint dialog.

Scope:
Backend — `backend/app/models/virtual_printer.py` (new column + migration), `backend/app/api/routes/virtual_printers.py`, `backend/app/services/virtual_printer/manager.py` (`_add_to_print_queue` / `_restamp_recent_queue_item`), `backend/app/services/archive.py`.
Frontend — `frontend/src/components/VirtualPrinterCard.tsx`, `frontend/src/components/PrintModal/FilamentMapping.tsx` (+ `types.ts`, `index.tsx`), `frontend/src/pages/ArchivesPage.tsx`.

### Alternatives Considered

A global "require exact filament match" setting was prototyped first but reverted — it changed scheduler/queue matching semantics broadly instead of solving the actual pain point (repetitive manual remapping on reprint), and was a bigger behavior change than needed.

### Feature Category

Print Queue & Scheduling

### Priority

Would improve my workflow

### Mockups or Examples

No mockups — the affected UI is the existing filament-mapping screen in the reprint/queue-edit dialogs; the change adds a small toggle button next to the existing filament rows.

Image

Image

### Contribution

### Checklist

Striker72rus:
## Description

Lets a reprint reuse the AMS slot the slicer itself picked, instead of re-deriving one from the file's static type/colour.

When a Print Queue VP has the new "Save AMS mapping" toggle on, the slicer's own live-resolved `ams_mapping` (from the `project_file` MQTT command) is persisted onto the archive as `extra_data.slicer_ams_mapping`. A later reprint can then reuse the exact physical spool via a new **Mapping** button in the filament-mapping panel — one click snaps every slot to the saved pick, click again reverts to auto-match. Archive cards and queue rows get an "AMS mapping saved" badge so it's visible beforehand.

The queue item's own `ams_mapping` (used for *that* dispatch) is still captured unconditionally whenever the slicer provides it — that part is a correctness fix, not gated behind the toggle. Only the archive persistence for future reprints is opt-in.

`add_to_queue` also falls back to the saved mapping automatically when the caller sends no explicit `ams_mapping` (e.g. a plain reprint with no per-slot edits) — covered by three new integration tests.

**Split from the original combined PR** (see discussion below): this half is genuinely opt-in and low-risk (#2684). The dispatch-time validation gate that keeps a stored mapping honest (#1308) changes behaviour for every existing user and will land as its own PR.

## Related Issue

Fixes #2684

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#60

- [x] Docs PR(s) linked above

## Type of Change

- [x] Test addition or update

## Changes Made

**Backend**
- `models/virtual_printer.py`, `core/database.py`, `api/routes/virtual_printers.py` — new `save_ams_mapping` column + migration, exposed on the VP schema/route
- `services/virtual_printer/manager.py` — new `_extract_slicer_ams_mapping_json` helper (mirrors the existing `nozzle_mapping` passthrough from #1780), gated archive persistence in both `_add_to_print_queue` and the late-MQTT `_restamp_recent_queue_item` path
- `services/archive.py` — promotes `print_data.ams_mapping` to the stable `archive.extra_data.slicer_ams_mapping` key
- `api/routes/print_queue.py` — `add_to_queue` falls back to `archive.extra_data.slicer_ams_mapping` when the caller sends no explicit mapping; response gains `archive_has_slicer_ams_mapping`
- `schemas/print_queue.py` — `archive_has_slicer_ams_mapping` field

**Frontend**
- `components/VirtualPrinterCard.tsx` — toggle UI (queue-mode VPs only)
- `components/PrintModal/FilamentMapping.tsx` (+ `types.ts`, `index.tsx`) — **Mapping** button, `archiveAmsMapping` prop, apply/revert logic
- `pages/ArchivesPage.tsx` — "AMS mapping saved" badge on archive cards
- `pages/QueuePage.tsx` — same badge on queue rows
- `api/client.ts` — API plumbing for the new fields

**i18n**
- New keys added to **all 13** locales (`en`, `de`, `es`, `fr`, `it`, `ja`, `ko`, `pt-BR`, `ru`, `tr`, `uk`, `zh-CN`, `zh-TW`), verified by `src/__tests__/i18n/locales.test.ts`.

**Tests**
- `tests/unit/services/test_virtual_printer.py` — new `TestExtractSlicerAmsMappingJson` covering the parser (missing field, valid list, stringified JSON, unparseable string, non-list/non-int values, bool rejected, empty list, the #2589 all-`-1` unresolved sentinel), plus `TestVirtualPrinterInstance` coverage: queue-item `ams_mapping` captured regardless of the toggle, archive persistence only when the toggle is on, sentinel dropped on both paths.
- `tests/integration/test_print_queue_api.py` — three new tests: `add_to_queue` falls back to the archive's saved mapping when unset, an explicit mapping always wins over the fallback, unrelated `extra_data` content doesn't false-trigger it.

## Screenshots

626987326-9eb5b657-a515-4ffc-808b-43c87a53745f
626987371-203c1c32-9368-4d08-a28e-7212a5113851

## Testing

- [x] I have tested with my printer model: X2D (with AMS + AMS HT)

Rebuilt from a clean `dev` base as a #2684-only branch and re-verified end to end:

| Check | Result |
|---|---|
| `ruff check backend/` | pass |
| `pytest backend/tests/unit` | **5023 passed**, 16 skipped, 1 failed* |
| `pytest backend/tests/integration` | **2420 passed**, 1 skipped |
| `tsc -b` | pass |
| `eslint .` | pass |
| `vitest run` | **2502 passed** (190 files) |

\* `test_bambu_cloud.py::TestBambuCloudRegion::test_china_region_totp_hits_cn_tfa_endpoint` — pre-existing and unrelated; it reaches out for a CSRF token and fails without outbound network. Verified it fails identically on a clean `dev` checkout.

## Checklist

## Additional Notes

**Split per review discussion.** This PR originally bundled #2684 with #1308 (dispatch-time mapping validation). Per your scope question below, split them: this PR is #2684 only now. #1308 will follow as its own PR, including the `loaded_by_tray_id`-empty fix for the blocking issue you found.


maziggy (changes requested):
Thanks for this - the split between "validate and pause" and "silently reassign" is the right call, and keeping skip_filament_mismatch_check separate from skip_filament_check is exactly the detail that usually gets missed. Both test files are thorough. One blocker, one scope question, and some smaller things.

Blocking: the unconditional slot check reads "no AMS telemetry" as "the slot is empty".

compute_filament_mismatch_for_queue_item guards on `if not status: return []`, but PrinterState.raw_data is a default-empty dict and printer_manager.get_status returns the state object as soon as the client exists. Between connect and the first pushall there is no "ams" key at all, so every required slot resolves to tray=None and check 1 - which has no opt-in - reports it. I ran the real function:

  raw_data empty (pre-pushall)      -> mismatches=[{'slot_id': 1, ..., 'loaded_type': None}]
  ams key present but empty         -> mismatches=[{'slot_id': 1, ..., 'loaded_type': None}]
  ams unit present, tray list empty -> mismatches=[{'slot_id': 1, ..., 'loaded_type': None}]
  AMS loaded correctly (control)    -> mismatches=[]

That item has no force_color_match override anywhere - a plain auto-dispatch queue item belonging to someone who never opted into anything.

The second half is what makes it bite. manual_start is only ever cleared by the start route and the webhook, never by the scheduler:

  tick 1 (no AMS telemetry): blocked=True  filament_mismatch=True  manual_start=True
  tick 2 (AMS telemetry OK): blocked=False filament_mismatch=False manual_start=True

So a transient blip permanently converts auto-dispatch items to manual, and then clears the badge that would have explained why. The user finds a stalled queue with no visible reason. One line fixes the premise: bail out when loaded_by_tray_id is empty. Not being able to see the AMS is not evidence that a slot is empty.

Scope question, and I would like your view before you invest more here. This PR bundles #2684 (opt-in per-VP toggle) with #1308 (dispatch gate). Check 1 fires for any item with a resolved ams_mapping regardless of the toggle or any override - so the #1308 half changes queue behaviour for every existing user, while the #2684 half is genuinely opt-in and low-risk. I am inclined to land them separately. Happy to be argued out of it.

Two more I would want before merge:

- add_to_queue falls back to archive.extra_data.slicer_ams_mapping whenever the   caller sends no explicit mapping, with no Mapping button involved. That does match the issue text ("read unconditionally on reprint"), but it is not in the PR description, which frames the saved mapping as button-driven. It is also the path that interacts most with the new gate: reprint an old archive, it silently inherits a stale physical slot, the spool has since changed, item goes manual. Please call it out explicitly and cover it with a test.
- In toggleArchiveMapping the ON branch sets force_color_match only where archiveAmsMapping[idx] >= 0, but the OFF branch clears it for every required slot. Someone who ticked "require color match" by hand before pressing Mapping loses it when they press it again. Undo only what the ON branch set.

Smaller things:

- _extract_slicer_ams_mapping_json: `all(v is None or v < 0 for v in raw)` - the `v is None` arm is unreachable, since the line above already requires every element to be an int. Also isinstance(v, int) accepts bool, so [True, False] passes as a mapping.
- There is Russian text inside a docstring ("Филамент N -> AMS X слот Y") -   comments in this repo are English.
- save_ams_mapping's explanatory comment sits below the column, so it reads as   belonging to gcode_injection. The convention in that file is a trailing comment on the column line.
- usingArchiveMapping is not reset when the plate or archive changes in the modal, so the button can read ON against rebuilt mappings.
- Two untranslated strings sit in the block you are editing - "Click to change slot assignment" and "Re-read". Pre-existing, but cheap while you are in there.
- Screenshots section is still a TODO placeholder.

Do not chase the four test_security.py::TestBackupKeyFiles failures if you see them - they are environmental. Those tests query virtual_printers against a Postgres instance that has not had run_migrations applied yet, so the new column is simply absent. I validated your Postgres DDL separately and it is fine, and the is_sqlite() branching follows the convention.

Striker72rus:
Thanks for the thorough pass — all of it is fixed or moot now, see below.

**Scope question**: agreed, splitting. This PR is #2684 only now — the entire dispatch-gate half (`compute_filament_mismatch_for_queue_item`, `_block_on_filament_mismatch`, the new `print_queue` columns, the 409 gate) is gone from this diff. That also makes the blocking issue you found moot *here* — but I've noted the `loaded_by_tray_id`-empty fix so it lands with #1308 when that PR goes up; you were right that "no AMS telemetry" must not read as "slot is empty."

**`add_to_queue` fallback**: now called out explicitly in the PR description and in a code comment at the call site, and covered by three new integration tests (fallback fires when nothing explicit was sent, explicit mapping always wins, unrelated `extra_data` doesn't false-trigger it).

**`toggleArchiveMapping` force_color_match asymmetry**: moot here too — that logic belonged to the #1308 half (the "require color match" auto-enable) and left with it. The #2684-only `toggleArchiveMapping` only touches `manualMappings`, nothing else to desync.

Smaller things, all applied:
- `_extract_slicer_ams_mapping_json`: dropped the unreachable `v is None` arm, `bool` is now rejected explicitly.
- Russian docstring text translated to English.
- `save_ams_mapping`'s comment is now a trailing comment on the column line.
- `usingArchiveMapping` resets on plate/archive change now (`useEffect` on `[archiveAmsMapping, plateLabel]`).
- "Click to change slot assignment" and "Re-read" are translated, wired through `t()`.

Full verification re-run on the rebuilt branch: `pytest backend/tests/unit` 5023 passed/16 skipped/1 pre-existing failure, `pytest backend/tests/integration` 2420 passed/1 skipped, `tsc -b` + `eslint` clean, `vitest run` 2502 passed (190 files).

maziggy (changes requested):
Thanks for splitting it - this half is far easier to reason about, and the round-2 work shows. The parser is careful in exactly the ways that matter: bool excluded from the int check, a stringified payload accepted defensively, and the all -1 unresolved sentinel dropped rather than trusted. Per-plate stamping mirrors the existing loop fix, the reload plumbing means a toggle takes effect without a restart, the i18n is real translations across all 13 locales, and the wiki PR is linked. Good shape.

Two things block it.

1. The toggle does not gate the archive write.

In services/archive.py:1266 the promotion fires for any print_data carrying ams_mapping:

if print_data.get("ams_mapping"):
    metadata["slicer_ams_mapping"] = print_data["ams_mapping"]

The VP path is fine - it passes None when the toggle is off. But main.py:3403, the real-printer auto-archive inside on_print_start, passes print_data={**data, "status": "printing"}, and data is the callback payload built at bambu_mqtt.py:4138, which always carries "ams_mapping": self._captured_ams_mapping. That value comes from the request-topic interception at bambu_mqtt.py:1306-1317, which by design captures for every print source - slicer-direct LAN prints and Bambuddy's own dispatches alike. data is never reassigned between the callback entry at main.py:2399 and the archive_print call.

I confirmed it at runtime with a throwaway test calling ArchiveService.archive_print with that exact payload shape:

EXTRA_DATA KEYS: ['_print_data', 'slicer_ams_mapping']
slicer_ams_mapping = [0, 1]

So on an instance with no virtual printer configured at all, every Print History entry grows the key, the AMS mapping saved badge appears on nearly every archive card and queue row, and via item 2 below every reprint gets silently pinned. That contradicts both the model comment ("archives don't grow this field unless the user opts in per virtual printer") and the setting description users will read. Either gate the promotion on a distinct key that only the VP path sets, or write extra_data["slicer_ams_mapping"] from the VP path directly instead of inside the shared archive service.

2. A saved mapping gets reused on a printer it was never resolved against.

The fallback at api/routes/print_queue.py:665-668 fires whenever the caller sent no explicit mapping, with no check that the reprint targets the printer the mapping came from. It applies to model-based items (printer_id=None, target_model set) and to a reprint aimed at a different printer. Downstream, print_scheduler.py:1305 returns early for any stored mapping that isn't all -1, so it is never recomputed for the actual dispatch target:

if item.ams_mapping and not _mapping_is_all_unresolved(stored_mapping):
    return

Tray IDs resolved against printer A then get dispatched on printer B. With item 1 in play that is reachable on a stock install with no VP involved. I know the inline comment defers this to the dispatch-time validation PR, but that PR is not here yet and this fallback ships now. Either scope the fallback to the case where the queue item targets the same printer the archive came from, or drop the automatic fallback and keep only the explicit Mapping button - which is opt-in per reprint and does not have this problem.

The same question applies to the deliberately un-gated queue-item stamping in _add_to_print_queue. For a VP set to "Any <model>", self.target_printer_id is None and there is no MQTT bridge (manager.py:1255), so the slicer should see no AMS and send nothing usable - but nothing enforces that, and anything it does send now overrides the scheduler unconditionally. Either only stamp when self.target_printer_id is not None, or add a comment explaining why the model-based case is safe.

Some more I want in before merge:

Multi-plate archives never get the Mapping button. PrintModal/index.tsx:1417 passes archiveAmsMapping to the single-panel FilamentMapping, but the per-plate loop at :1428 does not. The multi-plate Send All path is a primary producer of exactly these archives, so the feature's main source cannot reach its own UI.

The late-MQTT archive patch is untested. _restamp_recent_queue_item gained a new DB write - selecting PrintArchive rows and mutating extra_data - and nothing in the diff exercises it. The added tests cover _add_to_print_queue and the parser only, though the description claims coverage on both paths.

Smaller things:

- usingArchiveMapping does not reset on printer change. PrintModal/index.tsx:554 clears manualMappings when the printer selection changes, but the effect at FilamentMapping.tsx:46 only watches [archiveAmsMapping, plateLabel], so the button stays green after the mappings it applied are gone and it takes two clicks to re-apply.
- The revert branch at FilamentMapping.tsx:52-60 deletes every req.slot_id from manualMappings, including slots the user picked by hand before pressing the button. The comment directly above says the opposite. Track which slots the button set and delete only those.
- The apply branch skips entries below zero, so a slot the slicer resolved to the external spool silently keeps the auto-match. Probably what you intended - worth a comment either way.
- ArchivesPage.tsx:1116 guards with Array.isArray(...) while PrintModal/index.tsx:331 casts with as number[] | undefined. Same field, two levels of trust.
- The migration branches sqlite/postgres for a default both dialects accept. Harmless, and it matches its neighbours - just noting it is copied ceremony.

maziggy (changes requested):
Verification

┌───────────────────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│           Check           │                                                                Result                                                                 │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ ruff check backend/       │ pass                                                                                                                                  │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ ruff format --check       │ FAIL — 1 file                                                                                                                         │
│ backend/                  │                                                                                                                                       │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ pytest backend/tests/ -n  │ 8139 passed (1st run had 4 test_security.py::TestBackupKeyFiles failures; pass in isolation and on a clean re-run — the usual         │
│ 30                        │ parallel flake, not this PR)                                                                                                          │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ eslint src/               │ pass                                                                                                                                  │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ npm run build             │ pass                                                                                                                                  │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ npx vitest run            │ 2559 passed / 191 files — identical to dev, so this PR adds zero frontend tests                                                       │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ i18n parity               │ pass, 13 locales, real translations                                                                                                   │
└───────────────────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

Blockers

1. ruff format fails — CI will be red. backend/tests/unit/services/test_virtual_printer.py:2332:

select_archives_result.scalars = MagicMock(return_value=MagicMock(all=MagicMock(return_value=[mock_archive_row])))
needs wrapping. python -m ruff format backend/tests/unit/services/test_virtual_printer.py fixes it.

2. The unconditional queue-item ams_mapping silently disables Force color match and prefer-lowest.

manager.py:1075 sets ams_mapping=ams_mapping_json on every queue item whenever the VP has a fixed target_printer_id — deliberately not behind save_ams_mapping ("a correctness fix, not a feature toggle"). The consequence looks unintended:

print_scheduler.py:1416 — _ensure_ams_mapping returns immediately when a resolved mapping is already stored, so _compute_ams_mapping_for_printer never runs. That function is the only place a fixed-printer item's filament_overrides are consumed (print_scheduler.py:1493; lines 698-738 are the model-based eligibility branch, which these items don't take). It is also where prefer_lowest_filament, the #1766 AMS-filament-backup gate, and the inventory-remain overrides live.

So for a queue-mode VP pointed at one printer:

- queue_force_color_match still writes filament_overrides (manager.py:1026-1047) onto the item, and they are now never applied. Both toggles sit on the same card, so a user turns one on and the other stops working.
- prefer_lowest_filament and the AMS-backup gating stop applying to every VP queue print, for all existing users, on upgrade — this is not gated by the new toggle and there is no CHANGELOG or doc note.

It needs an explicit precedence decision. Either gate the queue-item mapping behind save_ams_mapping too, or keep it unconditional but skip it when filament_overrides carries force_color_match — and say so in the docs either way.

Should fix

3. No frontend tests. The vitest count is unchanged from dev. The two pieces of new UI logic are exactly the kind that regresses silently: appliedSlotIdsRef undo bookkeeping in FilamentMapping.tsx:55-80 (ON writes only slots with archiveAmsMapping[idx] >= 0, OFF must delete exactly those and no hand-made picks), and the printer_id === effectivePrinterId gate in PrintModal/index.tsx:340-344.

4. Badge copy overstates what happens. ArchivesPage.tsx:1115 and QueuePage.tsx:734 render the badge whenever slicer_ams_mapping.mapping is an array, ignoring the stored printer_id. queue.slicerAmsMapping.rowTooltip says "A reprint reuses that physical spool instead of re-guessing from type/color" — true only when reprinting on the printer it was resolved against. Reprint on any other printer and the badge is there while nothing is reused. Either scope the badge to the origin printer or soften the copy to "saved for <printer>".

Nits

5. Wiki. bambuddy-wiki/docs/features/virtual-printer.md documents the other three queue-mode toggles at line 80 and again at 1290+; the new one has no entry. An external contributor can't touch that repo — that's yours post-merge, together with the precedence decision from #2.
6. _extract_slicer_ams_mapping_json validates type, emptiness and the all--1 sentinel, but not length against the plate's filament count. A short array passes through and _ensure_ams_mapping leaves it. BambuStudio always sends full length, so this is hardening only.
7. Multi-plate Send All stamps one ams_mapping onto every plate, while extract_filament_requirements(file_path, plate_id) right above it is plate-scoped. That is correct if BS indexes ams_mapping by project-global filament slot (which Bambuddy's own buildAmsMapping does). It follows the nozzle_mapping precedent, but #1780 confirmed that one by wire capture — worth the same one-line confirmation.
8. VirtualPrinterCard's updateMutation.onError doesn't revert localSaveAmsMapping, so a failed save leaves the toggle visually on. Same pre-existing gap as queue_force_color_match and gcode_injection, so not a regression.

PR #2673, opened 2026-07-26
[Feature] Scheduling AMS filament drying after a delay / after a specific timestamp

bhamiltoncx:
## Description

I get my energy from the grid using Time-of-Use billing, which means I try to reduce power consumption during peak hours (5 to 9 pm on weekdays for my region).

It's very easy to schedule a print for a specific future time to avoid using power during this time. This PR extends that to manual drying from the Printer UI.

## Related Issue

Fixes #2638

## Documentation

**Companion docs PRs** (delete lines that don't apply):
- Wiki: maziggy/bambuddy-wiki#56

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Printer drying now includes "Start time" option that allows selecting "Now", "After delay", or "At time".
- New DB table for scheduled drying sessions
- Scheduler updated to check for pending drying sessions
- Drying popover is now correctly anchored to AMS dry icon
- Drying popover now correctly handles resizing browser viewport

## Screenshots

Screenshot 2026-07-25 at 6 27 00 PM
Screenshot 2026-07-25 at 6 26 53 PM
Screenshot 2026-07-25 at 6 26 49 PM
Screenshot 2026-07-25 at 6 26 45 PM
Screenshot 2026-07-25 at 6 26 42 PM

## Testing

Confirmed manually drying for "Now" still worked. Confirmed scheduled manual drying for specific time and after delay worked. Confirmed cancelling scheduled drying worked.

- [x] I have tested this on my local machine
- [x] I have tested with my printer model: P2S + AMS 2 Pro

## Checklist

- [x] My code follows the project's coding style
- [x] I have commented my code where necessary
- [x] My changes generate no new warnings
- [x] I have tested my changes thoroughly

## Additional Notes

The AMS 2 Pro is documented to only support drying temps of 45–65ºC but the API supports the full range of the AMS HT, up to 80ºC. The frontend prevents setting higher temps but the API needs to be updated as well.

Issue #2533, opened 2026-07-10
[Bug]: P1S AMS drying not starting


naldo29:
### Product

Bambuddy

### Area

Filament / AMS / Spoolman

### Bug Description

Start Drying Button doesnt do anything on my p1s with ams 2. It doesnt even pop up as a confrimation that i did something for exmaple when i enable a setting a popup appears of what i did. this does thing

### Expected Behavior

Its supposed to start drying

### Steps to Reproduce

you got to the printers secrtion.. click the flame icon on your ams system, and then click on start drying.

### Troubleshooting steps already taken

tried to check the wiki and tried to check github issues and no one has reported this

### Printer Model

P1S

### Bambuddy Version

0.2.4.9

### Printer Firmware Version

01.10.00.00

### Installation Method

Other

### Operating System

Windows

### Support Package (.zip) and / or relevant logs

[bambuddy-support-20260710-134903.zip](https://github.com/user-attachments/files/29900019/bambuddy-support-20260710-134903.zip)

### Final checks

- [x] I searched existing (open AND closed) issues and this bug hasn't already been reported or resolved
- [x] I checked the [Troubleshooting wiki](https://wiki.bambuddy.cool/reference/troubleshooting/) and the relevant feature page for my issue
- [x] I am using the latest version of Bambuddy (or the latest daily build)
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled
- [x] For any connection / printing / camera issue, I ran the in-app Connection Diagnostic and included the result above

maziggy:
Bambuddy is sending the command correctly and your printer is accepting it. The log shows three attempts while the printer was idle, each answered by the firmware with result: success, and the payload matches Bambu Studio's format field for field. Your AMS 2 Pro then stays at dry_status 0. So the printer takes the command and quietly drops it.

Two things were wrong on our side, and both are now fixed for the next build:

1. The Start Drying button gave no confirmation at all. It does now, and so does Stop.
2. We were reporting success purely on the printer's acknowledgement. Bambuddy now watches the AMS after sending, and if it never actually enters a drying cycle it tells you so instead of leaving the card unchanged.

That still leaves the question of why your printer declines. Two things cause it: the AMS 2 Pro's external power adapter not being connected, and the printer not being idle (a P1S cannot dry during a print). Could you check the power adapter, and then try starting drying from Bambu Studio with the printer idle? If Studio also fails to start it, the printer is refusing and we know where to look.

[...]

Issue #1447, opened 2026-05-19
[Bug]: Cannot control drying on P1S & AMS HT


kleinweby:
### Component

Bambuddy

### Bug Description

I can not control the AMS HT drying with Bambuddy which is connected to the P1S.

Where are two issues:
  - the drying dialog is always below the window, so I can not see the button (I workaround by using the web inspector).
  - when starting or stopping the drying nothing happens.

### Expected Behavior

Start/Stop the drying as commanded.

### Steps to Reproduce

1. Go to printer page
2. Click Flame Symbol on HT-A on my P1S
3. (Fix Popoverplacement via Webinspector)
4. Start drying
5. Dialog closes but nothing happens.

Or for stopping:
1. go to printer page
2. Click on "X" on the remaining drying time
3. nothing happens.

### Printer Model

P1S

### Bambuddy Version

0.2.4.1

### Printer Firmware Version

01.10.00.00

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

### Relevant Logs / Support Package

I tried to start the drying around 30s. Waited until approx. 60s and then started it on the device via its screen just to prove it is working and not an device issue.

[bambuddy-support-20260519-170834.zip](https://github.com/user-attachments/files/28020805/bambuddy-support-20260519-170834.zip)

### Screenshots

<img width="790" height="320" alt="Image" src="https://github.com/user-attachments/assets/ba9bf2f8-689f-444b-ae04-44b116dbf1bb" />

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I am using the latest version of Bambuddy
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled

pryamcem:
Probably a stupid question, but is your AMS connected to an external power supply? The P1S can’t provide enough power for drying with the AMS 2 Pro.

kleinweby:
Yes the external power is connected and starting/stopping drying via the P1S screen works just fine.

maziggy:
>     "network_mode_hint": "bridge"

First of all I would highly recommend to use Docker network mode host - as per the docs.

---

Also you should check your configuration. There are lots of error messages in your logs.

- updates / firmware_check "All connection attempts failed" — outbound internet unreachable (or DNS broken: later in the log "No address associated with hostname"). Bambuddy can't reach update-check / Bambu Lab CDN.
[...]

PR #2703, opened 2026-07-28
Feature: Scheduled drying

### LINKED ISSUE #2638: [Feature]: Allow delaying start of filament drying
### Problem or Use Case

I get my energy from the grid using Time-of-Use billing, which means I try to reduce power consumption during peak hours (5 to 9 pm on weekdays for my region).

It's very easy to schedule a print for a specific future time to avoid using power during this time, but I can't schedule filament drying for a future time.

### Proposed Solution

I'd love it if the filament drying feature could be delayed to start after:

1) An time interval (30 minutes, 2 hours, 1 day, etc.) passes, or
2) The current time reaches a specific time and date (`2026-07-23T12:34:56Z` or similar).

Bonus points if this shows up in the print queue (although the AMS 2 Pro isn't going to be able to print directly after drying when the filaments are retracted).

### Alternatives Considered

_No response_

### Feature Category

Print Queue & Scheduling

### Priority

Nice to have

### Mockups or Examples

_No response_

### Contribution

### Checklist

bhamiltoncx:
This is a re-do of #2673 on the correct parent branch (`dev` instead of `main`). There are no other changes.

## Description

I get my energy from the grid using Time-of-Use billing, which means I try to reduce power consumption during peak hours (5 to 9 pm on weekdays for my region).

It's very easy to schedule a print for a specific future time to avoid using power during this time. This PR extends that to manual drying from the Printer UI.

## Related Issue

Fixes #2638

## Documentation

- Wiki: maziggy/bambuddy-wiki#56

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Printer drying now includes "Start time" option that allows selecting "Now", "After delay", or "At time".
- New DB table for scheduled drying sessions
- Scheduler updated to check for pending drying sessions
- Drying popover is now correctly anchored to AMS dry icon
- Drying popover now correctly handles resizing browser viewport

## Screenshots

Screenshot 2026-07-25 at 6 27 00 PM
Screenshot 2026-07-25 at 6 26 53 PM
Screenshot 2026-07-25 at 6 26 49 PM
Screenshot 2026-07-25 at 6 26 45 PM
Screenshot 2026-07-25 at 6 26 42 PM

## Testing

Confirmed manually drying for "Now" still worked. Confirmed scheduled manual drying for specific time and after delay worked. Confirmed cancelling scheduled drying worked.

- [x] I have tested with my printer model: P2S + AMS 2 Pro

## Checklist

## Additional Notes

The AMS 2 Pro is documented to only support drying temps of 45–65ºC but the API supports the full range of the AMS HT, up to 80ºC. The frontend prevents setting higher temps but the API needs to be updated as well.


maziggy:
I have no problem if AI is used for assistance, but to be honest, I don't want to have my repo full of Claude Code commits. Can you please clean up?

bhamiltoncx:
> I have no problem if AI is used for assistance, but to be honest, I don't want to have my repo full of Claude Code commits. Can you please clean up?

Sounds good, I'll clean it up.

bhamiltoncx:
> I have no problem if AI is used for assistance, but to be honest, I don't want to have my repo full of Claude Code commits. Can you please clean up?

I cleaned up the commits and removed one semi-unrelated fix (debouncing MQTT messages with stale status when cancelling a drying session).

Let me know if you'd also like me to clean up the code, the tests, or the comments.

maziggy (changes requested):
Thanks for the re-do on the right parent and for cleaning up the commit history. The shape of this is good - the state machine is well thought out, the waiting_reason vocabulary is the right idea, and the test coverage is genuinely thorough (26 backend tests covering dispatch, grace window, interruption, completion and the auto-drying interaction, plus the frontend ones). The timezone round-trip is correct end to end, the new table needs no migration, and you remembered the explicit delete on printer removal since SQLite does not cascade. Two things block merge, three more I would like fixed first.

Blocking. The i18n parity check fails - uk is missing all ten new printers.drying.* keys, so CI will reject it. Every other locale has a real translation, so it looks like one file got skipped rather than anything systematic.

Also blocking: the scheduled dispatch path skips every guard the immediate endpoint applies. POST /printers/{id}/drying/start refuses screen-only models, refuses unsupported model/firmware combinations, turns dry_sf_reason into a specific message, and backfills the filament type from the first loaded tray. _check_scheduled_dryings does none of these. Schedule a dry on a P1S and the row dispatches, send_drying_command returns True because the publish succeeded, the row goes to running, the printer ignores it, and after the grace window it silently cancels itself - which is the exact failure mode #2533 exists to prevent. Same for a model or firmware that cannot dry. And dry_sf_reason collapses nine distinct reasons into one opaque ams_blocked token, so a run blocked on "plug in the external PSU" waits forever with nothing to tell the user why, even though the frontend already distinguishes that case. The filament default is a behaviour split too: the immediate path sends the loaded tray type, the scheduled path sends PLA. Could you validate model and firmware at schedule time so it fails fast in the UI, and share the dispatch preflight with start_drying rather than reimplementing it?

Three more. computePopoverPosition is shared with every temperature and fan popover, and the new rule flips above when neither side fits - the old comment cited #1447/#1458/#1669 for deliberately not doing that. The drying popover compensates with placement-aware maxHeight and internal scroll, but IndicatorControlPopover uses only top and left with no height cap, so on a short viewport it can now land above with its submit button off screen. The ScheduledDryingBanner adds a 30 second poll per printer card; I would rather have one fleet-wide query than twenty requests every 30 seconds for a list that is nearly always empty, and the list route already handles an omitted printer_id. And the start-mode state is not reset when the popover opens, so scheduling once with "At time" and then opening the popover on another AMS leaves the old timestamp in place, now in the past, and the POST 400s.

Smaller things: ScheduledDryingResponse skips the UTCDatetime serializer print_queue.py uses, so this route returns timestamps without the Z suffix that every queue route has; completed and cancelled rows are never pruned; and wait_for_off_peak ships as a column with no behaviour behind it.

bhamiltoncx:
Thanks for the thorough review!

> The i18n parity check fails - uk is missing all ten new printers.drying.* keys, so CI will reject it.

Fixed, `uk` is now localized.

> the scheduled dispatch path skips every guard the immediate endpoint applies.

Refactored immediate endpoint checks into `backend.app.services.drying_preflight` module and applied same checks to both scheduled dispatch path and immediate endpoint.

> And dry_sf_reason collapses nine distinct reasons into one opaque ams_blocked token, so a run blocked on "plug in the external PSU" waits forever with nothing to tell the user why, even though the frontend already distinguishes that case.

Restored the individual 9 reasons into their own error codes for the scheduled drying path and separated out user-actionable error codes like "plug in the external PSU".

> The filament default is a behaviour split too: the immediate path sends the loaded tray type, the scheduled path sends PLA.

Updated to send the loaded tray type for scheduled dryings.

> Could you validate model and firmware at schedule time so it fails fast in the UI, and share the dispatch preflight with start_drying rather than reimplementing it?

Done, we now fail fast if the model and/or firmware do not support drying and the preflight is refactored so the logic is shared.

> computePopoverPosition is shared with every temperature and fan popover, and the new rule flips above when neither side fits - the old comment cited https://github.com/maziggy/bambuddy/issues/1447/https://github.com/maziggy/bambuddy/issues/1458/https://github.com/maziggy/bambuddy/issues/1669 for deliberately not doing that.

Reverted the `computePopoverPosition` flip rule to the original conservative version.

> The ScheduledDryingBanner adds a 30 second poll per printer card

Changed to use a single fleet-wide poll for all printers.

> And the start-mode state is not reset when the popover opens, so scheduling once with "At time" and then opening the popover on another AMS leaves the old timestamp in place, now in the past, and the POST 400s.

Extracted `openDryingPopover` so both flame-button handlers reset their start mode in one place.

> ScheduledDryingResponse skips the UTCDatetime serializer print_queue.py uses, so this route returns timestamps without the Z suffix that every queue route has

Fixed, `ScheduledDryingResponse` now uses `UTCDatetime`.

> completed and cancelled rows are never pruned

Completed, cancelled, and failed rows are now pruned after 7 days (`SCHEDULED_DRYING_RETENTION_DAYS`) in `_check_scheduled_dryings()`.

> wait_for_off_peak ships as a column with no behaviour behind it.

Whoops, good catch. Removed this, it was for a subsequent PR.

maziggy:

bhamiltoncx:

Done and re-ran tests.

maziggy (changes requested):
Round two looks good - I checked each item against the code rather than the reply, and all nine hold up. The shared drying_preflight module is the right shape, the offline-printer nuance in the schedule-time check (model only, firmware deferred to dispatch) is a nicer answer than what I asked for, and the popover flip rule is back to the conservative version with tests pinning it. Test coverage is now genuinely thorough - grace window, interruption re-queue, deliberate-stop cancel, the auto-drying stop-all interaction and retention pruning all covered. ruff, eslint, build, the i18n parity suite and the full frontend suite are green here.

Two things I would like fixed before merge.

Failed rows are invisible. list_scheduled_dryings filters to pending and running, so a row that goes failed at dispatch - the firmware turns out too old, which is exactly the case that cannot be caught at schedule time on an offline printer - sets error_message, is never returned to any client, and is pruned after 7 days. That is the silent self-cancel from round one, moved to a new place: the run disappears and only the backend log knows why. The column and the schema field already exist, so this is the list route returning them and the banner rendering one with a dismiss.

The banner escapes the card. It renders between </CardContent> and </Card>, so it is full-bleed with no horizontal padding and its corners collide with the card's rounded bottom edge - visible in your own first screenshot. Moving it inside CardContent fixes it.

Smaller things.

The dispatch query has no order_by, so with two due rows on one printer - a state the tests explicitly cover - which one starts is arbitrary. The list route already orders by start_after; the dispatcher should match.

Code 3, "filament is at the AMS outlet, retract it first", is user-actionable but buckets into the generic ams_blocked. ams_not_found and printer_offline render no text at all, so a row blocked on an unplugged AMS shows a bare "Drying scheduled for ..." indefinitely.

ScheduledDryingCreate.filament has no max_length against a String(50) column: that is a 500 on PostgreSQL and accepted on SQLite. API-only, but it is a dialect split.

_update_running_scheduled_drying re-implements the AMS lookup with an unguarded int(a.get("id", -1)) instead of drying_preflight.find_ams_unit, which exists to be exception-safe. It runs inside check_queue, so a throw there costs the whole pass including print dispatch, on every tick.

The banner hardcodes text-[11px] and w-3 h-3, ignoring the --pc-* card-scale variables the rest of the card uses.

One question, not a blocker: an interrupted run re-queues with no upper bound. start_after is the earliest start and nothing caps the latest, so for the time-of-use billing that motivates the PR, a dry interrupted at 3am can restart at 6pm - the thing you are trying to avoid. Was that what wait_for_off_peak was for? Happy to leave it for the follow-up, but it deserves a sentence in the wiki.

bhamiltoncx:
> Failed rows are invisible

Fixed. `list_scheduled_dryings` now returns `failed` alongside `pending`/`running`, and banner now includes failed scheduled drying info. The API endpoint now supports `DELETE` on a `failed` row, and returns `{"status": "dismissed"}` instead of 404'ing.

> The banner escapes the card.

Moved inside `CardContent` so it's part of the card's padding. New screenshot:

Screenshot 2026-08-14 at 7 14 50 PM

> The dispatch query has no order_by

Thanks, fixed. `_check_scheduled_dryings` now matches the list route.

> Code 3, "filament is at the AMS outlet, retract it first", is user-actionable but buckets into the generic ams_blocked. ams_not_found and printer_offline render no text at all, so a row blocked on an unplugged AMS shows a bare "Drying scheduled for ..." indefinitely.

Added a new `WAITING_REASON_RETRACT`. The frontend now maps all 7 reasons to UI (we now add 9 new localized strings across the 13 locales).

> ScheduledDryingCreate.filament has no max_length against a String(50) column

Now using `Field("", max_length=50)` with length tests.

> _update_running_scheduled_drying re-implements the AMS lookup with an unguarded int(a.get("id", -1)) instead of drying_preflight.find_ams_unit

Fixed to call `drying_preflight.find_ams_unit()`.

> The banner hardcodes text-[11px] and w-3 h-3, ignoring the --pc-* card-scale variables the rest of the card uses.

Fixed to use the card scale variables.

> an interrupted run re-queues with no upper bound. start_after is the earliest start and nothing caps the latest, so for the time-of-use billing that motivates the PR, a dry interrupted at 3am can restart at 6pm - the thing you are trying to avoid. Was that what wait_for_off_peak was for? Happy to leave it for the follow-up, but it deserves a sentence in the wiki.

Yeah, I forgot to clean it up, the `wait_for_off_peak` will come in a separate PR now. This will be addressed in the subsequent PR. 

Updated https://github.com/maziggy/bambuddy-wiki/pull/56/ with a description of the re-queueing behavior and a warning about avoiding peak usage times.

maziggy (approved):

Issue #2707, opened 2026-07-29
[Bug]: External-camera captures that compete with the live view silently lose data (layer-timelapse, finish-photo)


bitbarista:
## Product
Bambuddy

## Area
Camera / live view / timelapse

## Bug Description
**I already have a working, tested fix for this and would like to submit it as a PR once this issue is assigned.**

Several capture paths open a fresh capture against a printer's *external* camera without first checking whether a live viewer is already connected. For a USB (V4L2) camera this is a hard exclusivity conflict — only one process can hold the device open — so the second capture attempt fails outright; the first (the live view) can also get caught up in it depending on which code path is involved. Confirmed on a Logitech C930e on an OrangePi Zero 2W, but the missing check is source-agnostic and would affect any external camera type under load.

This is the same class of problem `is_stream_active()`/`try_get_active_buffered_frame()` were built to solve for the *built-in* camera path (#1348, #1271) — reuse the live viewer's buffered frame instead of opening a second socket. That guard was just never extended to cover these external-camera call sites:

1. `backend/app/services/layer_timelapse.py` — `TimelapseSession.capture_layer()` grabs one frame on every layer change, unconditionally.
2. `backend/app/main.py` — `on_finish_photo_moment()`'s stage-22/last-layer pre-capture, and `_background_finish_photo()`'s fallback-chain capture, both call `external_camera.capture_frame()` directly with no guard.

(Also present but not yet triggered in my setup: `obico_detection.py`'s `_capture_frame()` external-camera branch, and `plate_detection.py`'s `capture_camera_image()` — same missing check, listed here for completeness in case anyone else hits them via Obico or plate-detection instead of timelapse/finish-photo.)

**Confirmed live impact** (debug logs, real prints, live viewer connected the whole time):
- Layer-timelapse: **0 of 87** layer captures succeeded on one print, **0 of 105** on another — every single attempt collided with the live view and failed. The post-print stitch produced either nothing or a badly truncated video.
- Finish photo: on one print, *both* the pre-capture and the fallback capture collided and failed back to back — the finish-photo notification went out with no image attached at all.

[...]

Issue #2705, opened 2026-07-29
[Bug]: Concurrent one-shot camera captures open competing RTSP connections when no viewer is attached


gzimbric:
### Product

Bambuddy

### Area

Camera / live view / timelapse

### Bug Description

Bambu firmware allows exactly one camera connection. `capture_camera_frame_bytes()` has no coalescing, so two backend consumers that both need a frame while no fan-out stream is active each open their own RTSP socket and compete.

The existing guards prevent each consumer from competing with the **fan-out broadcaster** (`is_stream_active()` / `try_get_active_buffered_frame()`, #1271 and #1348), but nothing coordinates the one-shot capturers with **each other**. Three call sites can fire independently:

- `obico_detection.py` — poll loop, every `obico_poll_interval` seconds
- `main.py` `_capture_snapshot_bytes` — `/camera/snapshot`
- `camera_diagnose.py` — the diagnose tool

Observed on a P2S (fw 01.02.00.00) with Obico enabled at a 10s interval:

```
13:08:22,165 INFO  [SNAPSHOT] Capturing fresh frame for printer 1
13:08:22,167 INFO  Capturing camera frame bytes from 10.0.2.43 using RTSP (model: P2S)
13:08:22,374 INFO  Capturing camera frame bytes from 10.0.2.43 using RTSP (model: P2S)
```

Two RTSP connections 207 ms apart. It recurs — 13:07:33 and 13:07:34 show the same pair. Downstream this shows up as the fan-out stream failing and being reaped:

```
13:05:24 WARNING RTSP read timeout for 10.0.2.43 (stream_id=1-fanout)
13:05:23 INFO    Killing stale ffmpeg stream 1-fanout (no frames for 58s)
```

**Why this isn't a duplicate.** Three closed issues sit next to this one and none cover it:

- **#1348** — fixed the **viewer vs Obico** case. Obico taps the fan-out buffer while a viewer is attached, and skips a poll rather than opening a competing socket. That guard works; I verified it still holds.
- **#1271** — the X2D single-connection constraint that motivated that guard.
- **#2580** — fixed recovery **after** a stall.

This is the remaining sibling case: **capturer vs capturer, with no viewer attached at all**. `is_stream_active()` returns False for everyone, so every one-shot consumer correctly concludes it isn't competing with a viewer — and then they collide with each other. It may well be a contributing cause of the #2580 symptom rather than a separate story, since a collision is one way the fan-out stream stops receiving frames in the first place.

[...]

Issue #1177, opened 2026-04-30
[Bug]: External camera images show as black in notifications/timelapse


nkm8:
### Component

Bambuddy

### Bug Description

With an external camera configured - tried both as a direct device connection and an external mjpeg stream hosted in go2rtc - the images all show as black.

### Expected Behavior

Images/timelapse images show actual camera stream contents.

### Steps to Reproduce

1. Configure external camera in go2rtc as mjpeg stream
2. Add in bambuddy
3. Start a print
4. Receive notifications with black image attachments

### Printer Model

A1 Mini

### Bambuddy Version

v0.2.4b2

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

### Relevant Logs / Support Package

[bambuddy-support-20260430-100200.zip](https://github.com/user-attachments/files/27246370/bambuddy-support-20260430-100200.zip)

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I am using the latest version of Bambuddy
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled

nkm8:
Additional context - I don't think it shows in the logs I shared, but I noticed that eventually the notification images started working mid-way through the print. I theorize this happened after I went into the Settings page and clicked the 'Test' button on the camera. From the logs, after the test:
```
10:02:26,738
[backend.app.api.routes.camera][4b9c879e] Using external camera (mjpeg) for printer 1 at 15 fps
10:02:31,248
[backend.app.services.external_camera][4b9c879e] MJPEG stream cancelled
10:02:31,248
[backend.app.services.external_camera][4b9c879e] External MJPEG stream ended, reconnecting (attempt 1/3)...
10:02:31,249
[backend.app.api.routes.camera][4b9c879e] External camera stream ended for printer 1
10:02:31,496
[backend.app.api.routes.camera][bbbfdf6b] Stopped 0 camera stream(s) for printer 1
10:02:31,498
[uvicorn.access][-] 192.168.1.8:50806 - "POST /api/v1/printers/1/camera/stop HTTP/1.1" 200
10:02:31,895
[backend.app.services.external_camera][-] capture_frame called: type=mjpeg, url=http://192.168.1.61:1984/api/stream.mjpeg?src=prin...
10:02:47,400
[backend.app.services.external_camera][-] capture_frame called: type=mjpeg, url=http://192.168.1.61:1984/api/stream.mjpeg?src=prin...
[...]

PR #2721, opened 2026-07-30
fix(camera): share one connection between concurrent one-shot external-camera captures

bitbarista:
Closes the one piece you left open on #2707: single-flight coalescing for `external_camera.capture_frame()`, matching #2705's shape for `capture_camera_frame_bytes()`.

## Why

A V4L2 USB device allows exactly one open handle, same as Bambu's own RTSP limit. `is_stream_active()` only stops a capturer from competing with an attached *viewer* — it does nothing for two one-shot capturers (Obico polling, the in-print frame bank, the finish-photo moment, plate detection, the notification snapshot) colliding with each other when no viewer is attached at all. That's exactly the #2705 gap, just reached through `external_camera.capture_frame()` instead of `capture_camera_frame_bytes()`.

## What

`capture_frame()` is now a coalescing wrapper around the actual dispatch logic (moved to `_capture_frame_uncoalesced`), keyed by `(url, camera_type, snapshot_url)` — `snapshot_url` is part of the key since #1177's override routes to a completely different endpoint, so two printers sharing a `camera_url` but differing only in `snapshot_url` must not share a capture.

Mirrors #2705's properties directly:
- Coalesces, doesn't cache — a call after the previous one finishes always captures fresh.
- Each caller keeps its own timeout (`wait_for(shield(leader), timeout=...)`) rather than inheriting whichever caller happened to open the connection.
- A follower whose leader fails takes its own turn instead of inheriting a failure it never had a chance to avoid — bounded at two rounds.
- Cancellation is disambiguated via `leader.cancelled()`, so a follower's own cancellation still propagates while a cancelled leader is treated as a failed one and the follower gets its own turn.
- `capture_in_flight()` mirrors `camera.py`'s helper, for any caller that needs to know it will join rather than measure its own connection.

## Testing

12 tests in `test_external_camera_capture_coalescing.py`, mirroring `test_camera_capture_coalescing.py`'s structure: the reported-shape collision, five callers sharing one connection, per-camera and per-snapshot-URL isolation, staying coalescing rather than becoming a cache, registry cleanup, a failed capture not poisoning its followers, bounded retry, a follower abandoning its wait without sabotaging the capture, and cancellation from either side.

Full suite passes (7818 tests as of this branch).

CHANGELOG entry added alongside the #2705 entry it follows.


maziggy:
Thanks!

Issue #1353, opened 2026-05-15
[Bug]: Timelapse on external camera not recording


Andlar94:
### Component

Bambuddy

### Bug Description

I have an external camera connected via RTSP stream, and a go2rtc feed from frigate for still frames. I have entered both into the camera section in the settings menu. It is stated that this will enable timelapse videos, and also uses the camera for plate detection. 

Currently on my setup, it does neither. Standard docker install. I've tried different camera streams, just rtsp stream, removing and re-adding the printer, turning camera on and off for printer etc.

It seems to capture images, per the log, but they are nowhere to be found, and I do not see any timelapse stiching happening

### Expected Behavior

Expected behavior: saving timelapse videos, using external camera feed in plate detection

### Steps to Reproduce

1. Start a print with external camera enabled
3. Check to see if timelapse is saved

### Printer Model

A1

### Bambuddy Version

V0.2.4

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

### Relevant Logs / Support Package

[bambuddy-support-20260515-110308.zip](https://github.com/user-attachments/files/27795161/bambuddy-support-20260515-110308.zip)

### Screenshots

<img width="619" height="810" alt="Image" src="https://github.com/user-attachments/assets/02ba48ac-8895-4617-b849-34f556ec9dad" />
External camera:

<img width="1396" height="979" alt="Image" src="https://github.com/user-attachments/assets/9fc1b829-933e-4833-a858-41fd11e85df3" />

A1 camera still used for plate detection too:

<img width="571" height="634" alt="Image" src="https://github.com/user-attachments/assets/a9e2ebe5-7389-46a9-903d-f67c26e085e0" />

directory structure on my NAS, just in case:

<img width="1078" height="103" alt="Image" src="https://github.com/user-attachments/assets/dbfb924b-ec85-4667-86c3-00235cbdcbff" />

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I am using the latest version of Bambuddy
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled

maziggy:
Thanks for the support bundle — it was load-bearing for finding this.

[...]

PR #2722, opened 2026-07-30
fix(camera): sweep orphaned timelapse session directories on startup

### LINKED ISSUE #2709: [Bug]: Layer-timelapse session directories orphaned permanently if the app restarts mid-print
## Product
Bambuddy

## Area
Camera / live view / timelapse

## Bug Description
**I already have a working, tested fix for this and would like to submit it as a PR once this issue is assigned.**

`layer_timelapse.py` tracks active timelapse sessions in an in-memory dict (`_active_sessions`). If the Bambuddy process restarts for any reason while a session is active — crash, container redeploy, OOM, power loss — that session's frame directory (`timelapse_frames/<printer_id>/<session_id>/`) is orphaned: nothing in the new process knows it exists, so `on_print_complete`/`cancel_session`'s cleanup never runs for it. If stitching had already started before the restart, a partially-written `timelapse_<session_id>.mp4` can be left behind too. There's no equivalent of the ffmpeg-process orphan janitor (#2675) for this.

**Confirmed live:** after several routine restarts during testing this week, `timelapse_frames/` had accumulated 38MB of exactly this — 3 orphaned frame directories and 2 corrupt 48-byte `.mp4` files (stitches interrupted mid-write).

**Fix:** `cleanup_orphaned_timelapse_sessions()`, run once at app startup. For each `printer_id` under `timelapse_frames/`, removes any frame directory or stitched-output file that doesn't match that printer's currently active session and is older than a defensive margin (5 minutes, to avoid any race with a session that's genuinely mid-creation). A restart-recovered print never gets a new timelapse session either (`_maybe_start_layer_timelapse` only fires on fresh `PRINT_START` events, per #1353), so nothing this sweep finds can ever legitimately be resumed — safe to always remove once old enough.

**Verified:** deployed with 38MB of real orphaned data present; startup log showed all 5 artifacts removed, `timelapse_frames/` dropped to 8KB.

## Expected Behavior
A restart shouldn't leave permanent, silently-growing storage leaks behind.

## Steps to Reproduce
1. Configure an external camera (layer-timelapse auto-starts on every print with one configured — no toggle to disable it, worth a separate UX conversation, but not this bug).
2. Start a print, and while a timelapse session is active, restart the Bambuddy container/process.
3. Check `timelapse_frames/<printer_id>/` — the frame directory from the interrupted session is still there and will never be cleaned up or stitched.

## Troubleshooting steps already taken
- Found by direct filesystem inspection (`du -sh timelapse_frames`) after noticing unexpectedly high usage; traced the specific leftover files back to session IDs matching timestamps of past container restarts.
- Confirmed the fix live: startup log listed each artifact removed by name, directory size dropped from 38M to 8K immediately after deploy.
- Ran the in-app Connection Diagnostic — all pass, not a connectivity issue.

## Printer Model
A1 (not printer-specific — this is a filesystem/process-lifecycle bug)

## Bambuddy Version
1.2.6b1 (dev)

## Installation Method
Docker

## Operating System
Docker (host: Linux ARM64, OrangePi Zero 2W)

## Support Package (.zip) and / or relevant logs
Not attached — already fixed and deployed; the orphaned data that reproduced this has already been cleaned up by the fix itself. Evidence: `du -sh` before (38M) and after (8K) the deploy, plus the startup log listing each removed artifact.

## Additional Context
Found while investigating overall timelapse storage behavior after fixing #2707 (device-contention) and #2708 (rotation) in the same feature. Separate observation, not part of this bug report: layer-timelapse has no user-facing enable/disable setting at all — it's implicitly on for anyone with `external_camera_enabled` set. Worth a UX discussion but out of scope here.

In-app Connection Diagnostic (all pass):
```json
{"printer_id":1,"ip_address":"192.168.8.202","overall":"ok","checks":[
  {"id":"port_mqtt","status":"pass"},
  {"id":"port_ftps","status":"pass"},
  {"id":"port_rtsps","status":"pass","params":{"port":6000,"protocol":"Chamber Image"}},
  {"id":"network_mode","status":"pass","params":{"mode":"host"}},
  {"id":"subnet","status":"pass"},
  {"id":"external_storage","status":"skip"},
  {"id":"mqtt_auth","status":"pass"},
  {"id":"developer_mode","status":"pass"},
  {"id":"printer_publishing","status":"pass"}
]}
```

## Final checks
- [x] I checked the Troubleshooting wiki and the relevant feature page for my issue

bitbarista:
Fixes #2709, assigned per CONTRIBUTING.md.

## Why

`layer_timelapse.py` tracks active sessions in an in-memory dict (`_active_sessions`). A process restart for any reason — crash, redeploy, power loss — loses that bookkeeping instantly, but the frames already written for the interrupted session (and any partially-stitched output `.mp4`) stay on disk with nothing left that knows they exist. `on_print_complete`'s cleanup never runs for them, since it has no session to find. A restart-recovered print doesn't get a replacement session either (`_maybe_start_layer_timelapse` only fires on a fresh `PRINT_START` event, #1353), so an orphaned directory can never be resumed or claimed by anything.

Confirmed live: after several routine restarts during testing, `timelapse_frames/` had accumulated 38MB — three abandoned frame directories and two 48-byte `.mp4` files from stitches that never finished writing.

## What

`cleanup_orphaned_timelapse_sessions()`, run once at app startup. For each `printer_id` under `timelapse_frames/`, removes any frame directory or stitched-output file that doesn't match that printer's current active session and is older than a defensive margin (5 minutes, so it can't race a session that's genuinely mid-creation at the moment this runs).

Verified against the real 38MB of leftovers on deploy: startup logged each of the five removed artifacts by name, and the directory dropped to 8KB.

## Testing

5 tests: the orphaned-directory-and-stray-output case, sparing a genuinely active session, sparing anything too recent to be sure about, and two defensive cases (no `timelapse_frames/` directory yet, an unrelated non-numeric entry under it).

Full suite passes (7811 tests as of this branch).

CHANGELOG entry included.


maziggy:
Thanks!

PR #2723, opened 2026-07-30
fix(camera): apply camera_rotation to finish photos and layer-timelapse frames

### LINKED ISSUE #2708: [Bug]: camera_rotation setting ignored for finish photos and layer-timelapse videos
## Product
Bambuddy

## Area
Camera / live view / timelapse

## Bug Description
**I already have a working, tested fix for this and would like to submit it as a PR once this issue is assigned.**

`camera_rotation` (Settings → General → Camera) is only ever applied in one place: `_apply_camera_rotation()`, wired solely into `_capture_snapshot_for_notification()` (the print-start notification photo, and the in-print frame bank used by the finish-photo pre-capture as a *fallback source*). Every other path that saves a still image from the camera writes it straight to disk with no rotation:

- The finish-photo pipeline's own primary capture paths — `on_finish_photo_moment()`'s stage-22/last-layer pre-capture, and `_background_finish_photo()`'s external-camera and buffered-frame fallback branches.
- `layer_timelapse.py`'s `TimelapseSession.capture_layer()` — every single frame of every timelapse video.

With a rotation configured (my camera is physically mounted upside-down relative to the frame, `camera_rotation=180`), the live view and the print-start photo come out correctly oriented, but the finish photo and the entire layer-timelapse video come out upside-down.

This was likely present for a long time but rarely noticed, because (see #2707, a separate device-contention bug) these same two capture paths were *also* silently failing most of the time — so there usually wasn't a photo/video to look at in the first place. Fixing that bug made this one visible.

**Fix:** extracted the rotation logic out into a shared `apply_camera_rotation(image_data, rotation, logger)` in `services/camera.py` (takes the rotation value directly rather than a printer object, so both the printer-shaped callers in `main.py` and `layer_timelapse.py`'s plain int field can use it), and applied it at every write point in the finish-photo pipeline and in `capture_layer()` (via `asyncio.to_thread`, since the PIL rotate/re-encode is CPU-bound and shouldn't block the event loop).

**Verified:** finish photos and every layer-timelapse frame confirmed correctly rotated across 2 consecutive full prints after the fix.

## Expected Behavior
`camera_rotation` should apply consistently to every saved image from that camera — live view, notification snapshot, finish photo, and timelapse frames alike.

## Steps to Reproduce
1. Set a non-zero `camera_rotation` for a printer (Settings → General → Camera → rotation).
2. Enable "Capture finish photo" and print with an external camera configured.
3. Compare the finish photo (and/or the attached timelapse video) against the live view — the live view is correctly rotated, the finish photo/timelapse are not.

## Troubleshooting steps already taken
- Confirmed via DEBUG logs: `_capture_snapshot_for_notification`'s own log line (`Applied N° camera rotation: ...`) never appeared for finish-photo or timelapse captures before the fix — those paths simply had no such logging or PIL call at all.
- Confirmed the fix live: after wiring in the shared helper, the same log line now appears for every layer capture and every finish-photo save, with output byte counts matching a genuine rotate+re-encode.
- Ran the in-app Connection Diagnostic — all pass, not a connectivity issue.

## Printer Model
A1

## Bambuddy Version
1.2.6b1 (dev)

## Installation Method
Docker

## Operating System
Docker (host: Linux ARM64, OrangePi Zero 2W)

## Support Package (.zip) and / or relevant logs
Not attached — already fixed and deployed. Evidence: before/after DEBUG logs as described above; visually confirmed the finish photo and timelapse video orientation directly.

## Additional Context
Found immediately after fixing #2707 (a related device-contention bug in the same capture paths), once finish-photo/timelapse actually started reliably producing output to look at. See also #2709 (orphaned timelapse session cleanup), found the same week in the same area.

In-app Connection Diagnostic (all pass):
```json
{"printer_id":1,"ip_address":"192.168.8.202","overall":"ok","checks":[
  {"id":"port_mqtt","status":"pass"},
  {"id":"port_ftps","status":"pass"},
  {"id":"port_rtsps","status":"pass","params":{"port":6000,"protocol":"Chamber Image"}},
  {"id":"network_mode","status":"pass","params":{"mode":"host"}},
  {"id":"subnet","status":"pass"},
  {"id":"external_storage","status":"skip"},
  {"id":"mqtt_auth","status":"pass"},
  {"id":"developer_mode","status":"pass"},
  {"id":"printer_publishing","status":"pass"}
]}
```

## Final checks
- [x] I checked the Troubleshooting wiki and the relevant feature page for my issue

bitbarista:
Fixes #2708, assigned per CONTRIBUTING.md.

## Why

`camera_rotation` (Settings → General → Camera) was only ever applied in one place: `_apply_camera_rotation()`, wired solely into `_capture_snapshot_for_notification()` (the print-start notification photo, and the in-print frame bank). Every other path that saves a still image from the camera wrote it straight to disk unrotated:

- The finish-photo pipeline's own primary capture paths — `on_finish_photo_moment()`'s stage-22/last-layer pre-capture (consumed and written in `_background_finish_photo`'s cached-frame branch), and `_background_finish_photo`'s external-camera and buffered-frame fallback branches.
- `layer_timelapse.py`'s `TimelapseSession.capture_layer()` — every single frame of every timelapse video.

With a rotation configured, the live view and the print-start photo came out correctly oriented; the finish photo and the entire layer-timelapse video did not. Likely present for a long time without being noticed, since until #2707 both paths usually produced nothing at all while a viewer was watching — there was rarely a photo or video to notice was misoriented.

## What

Extracted the rotation logic into a shared `apply_camera_rotation(image_data, rotation, logger)` in `services/camera.py` — takes the rotation value directly rather than a printer object, so both `main.py`'s printer-shaped callers and `layer_timelapse.py`'s plain `int` field can use it. `main.py`'s original `_apply_camera_rotation` becomes a thin compatibility wrapper, so its existing call sites are unchanged.

Applied at every point a frame is about to be written to disk: the three `_background_finish_photo` branches, and `capture_layer()` (via `asyncio.to_thread`, since the PIL rotate/re-encode is CPU-bound and shouldn't block the event loop) — after #2707's `live_frame_for_capture()` resolves the frame, regardless of whether it's fresh or reused from the live view's buffer. `TimelapseSession` gains a `rotation` field, threaded from `printer.camera_rotation` in `_maybe_start_layer_timelapse`.

Built-in camera's final `capture_finish_photo()` fallback (writes its own file, returns only a filename) is left unrotated — out of scope here, would need its own change to `services/camera.py`.

## Testing

Layer-timelapse is covered by tests for both frame sources (fresh capture, reused from the live view's buffer) and for the rotation call being skipped entirely when nothing is configured. The finish-photo branches live inside `_background_finish_photo`, a nested closure with no existing unit-test harness (same gap as `_background_finish_photo` generally) — verified live instead, against real hardware with a configured rotation, confirming the finish photo and every timelapse frame came out correctly oriented across two consecutive prints.

Full suite passes (7809 tests as of this branch).

CHANGELOG entry included.


maziggy:
Using AI for assistance ins totally fine for me, but to be honest, I don't ewant to have all these Claude commits in the repository. For any future PR please omit them. Thanks.

bitbarista:
> Using AI for assistance ins totally fine for me, but to be honest, I don't ewant to have all these Claude commits in the repository. For any future PR please omit them. Thanks.

Thanks for the feedback. I do use AI as an assistant when working on contributions, but I'll make sure to clean up and squash those commits before submitting future PRs. Thanks for the review.

PR #2724, opened 2026-07-30
feat(queue): keep bed warm between prints + smart chamber soak reduction

ticfinack:
## Summary

Two related improvements for printers that require chamber heating (ASA, ABS, PA, PC etc.) when running consecutive prints through the queue.

### Keep bed warm between prints

When a printer finishes a print (FINISH state) and the next queued item requires chamber heating, BamBuddy now maintains the bed at the item's `bed_temperature` until the plate is cleared and the next print dispatches. This prevents the chamber from cooling during the bed-clearing window, avoiding a full re-soak on the next print.

- Controlled by the `queue_keep_bed_warm` setting (default: `false`)
- Only fires when the next queued item's filament maps to a non-zero chamber target via `preheat_filament_targets` — PLA/PETG/TPU are skipped automatically
- Bed temperature is read from the next item's archive metadata, not a hardcoded value
- No-op when the printer is being dispatched this tick (`_preheat_and_soak` takes over)

### Chamber temperature history + smart soak reduction

The scheduler now samples every printer's chamber temperature each tick (every 3–30 s) into a 2-hour rolling per-printer history. Before running the heat-soak wait, `_chamber_soak_remaining()` scans the history backwards to find the last moment the chamber dropped below target. Time elapsed since then is credited against the configured soak duration.

- If the chamber has been continuously above target for the full soak duration, the soak (and the entire preheat stage) is skipped via a fast-path return
- If the chamber dipped below target partway through, only the remaining soak time is waited
- Printers without a chamber sensor (P1S, A1 etc.) are unaffected — they always run the full soak timer
- History is in-memory and resets on restart (conservative: full soak assumed on first print after restart)

### Practical effect

For back-to-back ASA prints with `require_plate_clear` enabled: the bed stays hot during bed clearing, the chamber never drops, and by the time the second print dispatches the soak history shows ≥30 min above target — preheat is skipped entirely and the print starts immediately.

## Test plan

- [ ] Enable `queue_keep_bed_warm` and queue two consecutive ASA prints on an X1C; confirm bed stays at 90°C during plate clear and second print skips preheat
- [ ] Confirm PLA queue items do not trigger keep-warm (chamber target = 0)
- [ ] Confirm P1S/A1 (no chamber sensor) still runs full soak timer
- [ ] Restart container mid-queue; confirm first print after restart runs full soak (conservative default)
- [ ] Confirm `queue_keep_bed_warm = false` (default) leaves existing behaviour unchanged

### DIFF
--- a/backend/app/schemas/settings.py
+++ b/backend/app/schemas/settings.py
@@ -376,6 +376,15 @@ class AppSettings(BaseModel):
         le=1800,
         description="Additional hold time at temperature after the chamber reaches the target (or after max_wait_seconds elapses). 0 = no soak.",
     )
+    queue_keep_bed_warm: bool = Field(
+        default=False,
+        description=(
+            "While a printer is in FINISH state awaiting plate-clear and the next queued item requires "
+            "chamber heating, maintain the bed at that item's bed_temperature so the chamber stays hot "
+            "during the bed-clearing window. Only fires for filaments with a non-zero chamber target "
+            "(ASA, ABS, PA, PC etc.); PLA/PETG prints are skipped automatically."
+        ),
+    )
 
     # User-configurable presets for the printer-card temperature / fan-speed
     # popovers. Each is a JSON array of exactly 3 ints (the "Off" button is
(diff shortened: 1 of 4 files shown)


maziggy:

PR #2727, opened 2026-07-30
feat(queue): keep bed warm + smart chamber soak reduction for consecutive prints requiring chamber heat

### LINKED ISSUE #2726: feat(queue): keep bed warm + smart chamber soak reduction for consecutive prints requiring chamber heat
## Problem

When running consecutive queued prints where filaments have a non-zero chamber target configured in Settings → Preheat & Heat Soak, the chamber cools during the bed-clearing window between prints. The next print then waits the full configured soak time even though the chamber was still hot — resulting in an unnecessary re-soak.

## Proposed Solution

Two complementary features:

**1. Keep bed warm between prints**

While a printer is in FINISH state awaiting plate-clear, maintain the bed at the next queued item's target temperature so the chamber stays hot during the bed-clearing window. Only fires when the next queued item's filaments map to a non-zero chamber target (via the existing per-filament `preheat_filament_targets` map). Controlled by a new `queue_keep_bed_warm` setting (default off). Requires plate-clear confirmation to be enabled, since the keep-warm window only exists when plate-clear confirmation is active.

**2. Smart chamber soak reduction via temperature history**

The scheduler samples each printer's chamber temperature every tick into a 2-hour rolling in-memory history. Before running the heat-soak wait, it checks how long the chamber has continuously been above target and credits that time against the configured soak duration — reducing or skipping the soak entirely if the chamber has been at temperature long enough.

Together these eliminate the re-soak penalty for back-to-back prints where the chamber never actually cooled between jobs.

## Affected Printers

- Printers with a chamber sensor (X1C, P2S, H2C, H2D etc.) — both features apply fully
- Printers without a chamber sensor (P1S, A1, A1 Mini etc.) — keep-warm still applies; soak reduction is skipped since there is no sensor to read

ticfinack:
## Description

Consecutive queued prints in chamber-heated materials (ASA, ABS, PA, PC) each paid a full heat-soak from cold, even when the print that just finished had left the chamber at temperature. Two complementary changes remove that cost.

**Keep bed warm between prints**

While a printer sits in FINISH awaiting plate-clear and the next queued item needs chamber heat, the bed is held hot so the chamber does not cool during the bed-clearing window. The bed is the chamber's heating element here rather than a print surface — nothing is printing, and the next print's own gcode re-targets the bed the moment it starts — so the hold runs at a configurable `queue_keep_warm_bed_temp` (default 90 °C), raised to the item's own bed temperature whenever that is higher.

The chamber requirement is resolved exactly as preheat already does it: the per-item chamber override wins if set, otherwise `_derive_chamber_target()` takes the maximum across the **loaded AMS filaments**. PLA/PETG-only loads derive 0 and are skipped automatically.

Bounds and safety:

- **Gated in the backend** on `queue_keep_bed_warm` AND `require_plate_clear` AND `preheat_enabled`, all re-checked every tick, so a stale UI can never leave the feature running.
- **Bounded** by `queue_keep_warm_max_minutes` (default 120). When it elapses the bed is switched off and the hold latches until the printer is next a candidate, so a plate nobody clears cannot leave the bed hot indefinitely.
- **Released on every other exit** too — item deleted, queue emptied, or any gate toggled off mid-hold.
- **Never fights the user.** The release only sends bed-off when firmware still reports the exact target keep-warm set; a temperature changed by the user or a print is left alone.
- **Idempotent.** No MQTT publish when the firmware already holds the target.

**Smart soak reduction from chamber history**

The scheduler samples each connected printer's chamber temperature every tick into a 2-hour rolling in-memory history. Preheat credits the time the chamber has already spent at temperature against the configured soak, shortening it or skipping it.

Credit starts no earlier than the newest sample, the most recent unbroken run of samples, or the end of the last real dip below target. A dip only counts once it outlasts a grace period, because an enclosed chamber cannot lose and regain several degrees quickly — measured on an X1C, cooling from 55 °C to below 48 °C takes 23–73 minutes (~0.2 °C/min), while the fastest drop ever recorded was 27 °C/min, which that thermal mass cannot physically do. Brief low readings are therefore a door opening or sensor noise rather than lost soak, and a plate swap — exactly when keep-warm runs — produces one. A stale history credits nothing: at that cooling rate the chamber can cross the threshold unobserved, so the full soak runs instead.

Printers without a chamber sensor (P1S, A1, A1 Mini) are unaffected by the soak reduction, since there is nothing to read; keep-warm still applies to them.

## Behaviour changes to preheat itself

Three changes here affect **all** preheat users, not only those who enable the new toggle. Calling them out explicitly for review:

1. **Cancelling or deleting a queued item now stops a preheat already running for it.** Those routes only write `status` to the database, which a dispatch coroutine parked in `asyncio.sleep` cannot observe — so the heaters ran out the rest of max_wait + soak (45 minutes at the default settings) and the printer stayed in `busy_printers`, blocking every other queued item behind a print that was not happening. The routes now signal the scheduler directly and the preheat stage sleeps in slices, so a cancellation landing **during preheat** is noticed within seconds and that dispatch is abandoned. Reproduced on hardware before the fix. Scope: this covers the preheat window only — a cancellation arriving during the FTP upload still starts the print, as it did before, since `_start_print` checks status once at entry and not again.
2. **Preheat unwinds itself when a dispatch aborts.** It records which commands it sent (bed, chamber, airduct) and turns them back off on any non-success exit — failed upload, cancelled item, exception — instead of leaving the printer heating for a job that is not happening.
3. **A chamber-heated print whose file carries no bed temperature no longer skips preheat.** Those items used to bail out with "archive has no bed_temperature metadata" and start against a cold chamber, which is the outcome preheat exists to prevent. They now heat the bed to `queue_keep_warm_bed_temp`. A parsed bed temperature still wins, and a print with no chamber requirement still skips, so no bed temperature is invented for the print itself. Preheat's bed target is transient in any case — the print's own gcode issues its M140/M190 at start.

## Scope / v1 limitations

- Keep-warm applies to queue items **already assigned to a printer**. Model-based items (no `printer_id` yet) do not trigger it. Library-file items do: chamber need comes from the loaded AMS filament rather than the item, and the item's own bed temperature is consulted only to raise the hold above the configured floor.
- The hold drives the **bed only**. It does not send M141 or move the airduct during the wait window, so on X1E/H2x/X2D the chamber is held by bed radiation plus whatever the chamber heater was already doing. A chamber-heater-aware hold is possible follow-up work.
- Chamber history is in-memory and resets on restart — the first print after a restart always runs the full soak.

## Related Issue

Fixes #2726

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#62

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

## Changes Made

- `print_scheduler.py`
  - `_apply_keep_warm()` engagement loop with per-printer `_KeepWarmEntry` state (started / held_target / expired)
  - `_release_keep_warm()` / `_sweep_keep_warm()` — bed-off on timeout and on every exit path, guarded against clobbering a user-set target
  - `_sample_chamber_temps()` — per-tick sampling, TTL pruning, skips disconnected printers, evicts state for deleted printers; called inside the scheduler loop's error boundary
  - `_chamber_soak_remaining()` — history scan with freshness guard, disconnect-gap contiguity, and dip debounce
  - `notify_dispatch_cancelled()` + sliced waits so a cancelled item stops its own preheat
  - `_preheat_pin` registry and `_rollback_preheat_pin()` for aborted dispatches
  - fast path in `_preheat_and_soak()` that still sends bed/airduct/chamber targets, so nothing cools during the 3MF upload
- `schemas/settings.py`, `routes/settings.py` — three new settings (below) with bounds and coercion
- `routes/print_queue.py` — cancel, delete and batch-cancel notify the scheduler
- `frontend/src/api/client.ts`, `pages/SettingsPage.tsx` — toggle plus two number inputs in the Preheat & Heat Soak card
- `frontend/src/i18n/locales/*.ts` — new keys translated in all **13** locale files
- Tests — `test_scheduler_keep_warm.py`, `test_scheduler_chamber_soak.py`, `test_scheduler_preheat_fast_path.py`, plus updates to `test_scheduler_preheat.py`

### New settings

| Key | Default | Range |
|---|---|---|
| `queue_keep_bed_warm` | `false` | bool |
| `queue_keep_warm_bed_temp` | `90` | 40–110 °C |
| `queue_keep_warm_max_minutes` | `120` | 5–480 min |

## Screenshots

**Before** — Preheat & Heat Soak card:

pr2727-before

**After** — new toggle and its two settings:

pr2727-after

## Testing

- [x] I have tested with my printer model: X1C

Verified against a real X1C, including the soak actually being skipped between back-to-back prints (`preheat soak skipped — chamber has been above 50°C for ≥1800s`), the hold releasing when the queue empties, the configurable timeout firing and shutting the bed off, and a chamber-heated file with no bed-temperature metadata now preheating instead of starting cold.

Full gate run locally after rebasing onto current `dev`: `ruff check` + `ruff format --check`, i18n parity across 13 locales, the complete backend pytest suite (8610 passed / 3 skipped), ESLint, `tsc --noEmit`, and the frontend Vitest suite.

## Checklist

## Additional Notes

`queue_keep_bed_warm` defaults to `false`, so nothing changes for existing installs until it is switched on in Settings → Queue → Preheat & Heat Soak. The toggle is disabled in the UI when preheat or plate-clear confirmation is off, and the backend independently re-checks all three gates on every tick.


ticfinack:
Uploading screenshots via comment to get asset URLs — will edit into PR description.

maziggy:
You still don't followed our rules from CONTRIBUTING.md - you don't have an assigned issue. Next time please make sure you follow the rules.

maziggy:
Using AI for assistance is totally fine for me - but to be honest - I don't want to have my repository full of Claude commits. Please remove. Thanks.

maziggy (changes requested):
Blockers

1. i18n fails CI. uk.ts was not touched at all — both keys are missing (parity count: uk 5711 vs en 5713). The other 11 locales received the English string verbatim. node frontend/scripts/check-i18n-parity.mjs exits non-zero on both counts. The description says "all 12 locale files"; there are 13, and they need real translations rather than English placeholders.

2. ruff check fails. backend/tests/unit/test_scheduler_chamber_soak.py:16 — I001 (import block unsorted). ruff format --check would reformat the same file.

3. zh-CN has an unrelated line corrupted. requirePlateClearDescription had its opening “ (U+201C) replaced with ” (U+201D): ...和”将打印板标记为已清理”按钮. Nothing to do with this feature — please revert that line.

4. _sample_chamber_temps() sits outside the scheduler's error boundary. print_scheduler.py:326 is the only statement in the while self._running body that is not inside the try/except Exception. It calls get_all_statuses() -> check_staleness(), which invokes the on_state_change callback and _reset_client_for_reconnect(). If any of that raises, the exception escapes run(), and spawn_background_task has no supervisor — the queue stops dispatching permanently until the container restarts. Move the call inside the existing try.

5. The aux-fan change is undocumented and ungated. print_scheduler.py:2917-2922 sends set_aux_fan(255) on every preheat with a chamber target. It is not gated by queue_keep_bed_warm, not mentioned in the description, and not in the "Changes Made" list. The comment justifies it by a specific third-party accessory ("PandaBreath", a word that appears nowhere else in the codebase), but it applies to every X1C/X1E/H2x/P2S/X2D user who already has preheat enabled. The claim that "the print's own gcode takes over fan control once dispatched, so no cleanup is needed" holds only on the happy path — a preheat timeout, lost state, a failed upload, or a cancelled item all leave the fan pinned at 255 with nothing to reset it. Please either drop this from the PR or give it its own setting and restore the previous speed on every exit path.

Important

6. Keep-warm's gating is UI-only. The backend block at line 880 reads queue_keep_bed_warm and nothing else — it never checks require_plate_clear or preheat_enabled, both of which the description says are required. Compounding it, the toggle is disabled when either is off (SettingsPage.tsx:4596), so a user who enables keep-warm and then turns off plate-clear ends up with a live feature they cannot switch off through the UI.

7. No upper bound on the keep-warm window. The bed is held at the archive's bed temperature for as long as the printer sits in FINISH. If nobody clears the plate over a weekend, the bed stays at 100 C indefinitely, with no printer-card indication and no timeout. This needs a max duration.

8. M140 is re-sent every tick with no idempotence guard. Forty lines below, the airduct code reads current state first, specifically so there is "no MQTT chatter when the flap is already where we want it". state.temperatures["bed_target"] is already parsed (bambu_mqtt.py:3320) — compare against it and skip the publish when it already matches.

9. The fast path returns before M140, airduct and M141 (line 2887). It is reachable with keep-warm off, since a bed coasting down from the previous print still satisfies bed >= bed_target - 2. The bed and chamber are then left with no target set while the 3MF uploads (minutes on a large file), cooling during exactly the window preheat exists to cover. On H2x/P2S it also skips the flap flip that the airduct comment says is needed for convergence. Send the targets, then return.

10. chamber_needed scans every loaded tray, not the next item's filaments. ABS parked in AMS slot 4 will trigger keep-warm for a PLA print in slot 1. _derive_chamber_target has the same max-across-trays semantics, so the copy is faithful to it — but the description ("the next item's filaments map to a non-zero chamber target") describes something different. It is also a 12-line reimplementation of that existing method; self._derive_chamber_target(printer, filament_targets) > 0 would cover it.

Nits

- queue_keep_bed_warm is missing from the bool list at routes/settings.py:190. Nothing breaks today (the response_model=AppSettings coercion turns "false" into False — I verified), but it is the only boolean setting outside that list.
- test_tolerance_boundary_at_threshold_counts_as_below passes for the opposite reason to the one it documents. 48.0 < 48.0 is False, so last_below_ts stays None and the "never dipped" branch returns 0. The # last_below_ts = SOAK seconds ago comment describes a path the test never takes, and since both branches return 0 the test cannot distinguish them.
- No tests for the keep-warm loop, the fast path, or _sample_chamber_temps (TTL pruning, absent chamber key, disconnected printer).
- _sample_chamber_temps samples disconnected printers: get_all_statuses() returns every registered client and check_staleness() never clears temperatures, so an offline printer keeps logging its last hot reading every tick. It self-heals once a real cold sample arrives on reconnect, so the impact is bounded — but please skip not status.connected, and consider requiring sample contiguity in the "no dip in window" branch instead of crediting a gap as time at temperature.
- .github/pr-screenshots/2727/*.png adds 103 KB of binaries to the repo permanently, and that directory does not exist on dev. Please paste the images straight into the PR body instead.
- dict[int, deque] would be better as dict[int, deque[tuple[float, float]]], and _chamber_history is never evicted when a printer is deleted.

ticfinack:
Thanks for the thorough review — that was genuinely useful, and sorry for the slow turnaround. Branch is rebased onto current `dev` and squashed into a single commit, so the history is rewritten (which is why the merge commit you pushed is no longer there — all of that `dev` content is included via the rebase). Description is rewritten too.

**Blockers**

- **i18n** — both keys now exist in all **13** locale files with real translations rather than English placeholders, `uk.ts` included. `node frontend/scripts/check-i18n-parity.mjs` exits 0.
- **ruff** — `ruff check backend/` and `ruff format --check backend/` both clean on the pinned 0.15.20.
- **zh-CN** — the corrupted quote on `requirePlateClearDescription` is not present; that line is untouched.
- **Error boundary** — `_sample_chamber_temps()` is now the first statement inside the existing `try` in `run()`, so it can no longer take the scheduler down.
- **Aux fan** — dropped from this PR entirely. You were right that it changed behaviour for every chamber-preheat user to serve one accessory. If there's interest later it can come back as its own opt-in setting.

**Important**

- **Gating** — the backend now requires `queue_keep_bed_warm` AND `require_plate_clear` AND `preheat_enabled`, re-checked every tick, so the UI can't leave a live feature the user can't switch off.
- **Upper bound** — new `queue_keep_warm_max_minutes` (default 120, range 5–480). On timeout the bed goes to 0 and the hold latches so it can't re-arm. It also releases on every other exit: item deleted, queue emptied, or a gate toggled off mid-hold. The release only fires when firmware still reports the exact target keep-warm set, so a temperature you changed by hand is left alone.
- **M140 idempotence** — compares against `state.temperatures["bed_target"]` and skips the publish when it already matches.
- **Fast path** — now sends bed, airduct and chamber before returning, and registers each one for rollback.
- **`chamber_needed`** — replaced with `self._derive_chamber_target(printer, filament_targets) > 0`. You were also right that the description misdescribed it: it derives from the **loaded AMS filaments**, same semantics as preheat, and the description now says so.

**Nits** — `queue_keep_bed_warm` added to the bool list in `routes/settings.py`; the tolerance-boundary test rewritten so the two branches produce different values instead of both returning 0; tests added for the keep-warm loop, the fast path and `_sample_chamber_temps` (TTL pruning, absent chamber key, disconnected printer, `None` status, eviction); disconnected printers are skipped and the no-dip branch now requires sample contiguity; the screenshot binaries are gone from the branch and the images live in the description instead; `_chamber_history` is typed `dict[int, deque[tuple[float, float]]]` and evicted when a printer disappears.

**Three things I found testing this on my own X1C afterwards**, all fixed here and called out in the description since they affect preheat generally rather than just the new toggle:

1. Cancelling or deleting a queued item didn't stop a preheat already running for it — the routes only write `status`, which a coroutine parked in `asyncio.sleep` can't see, so the heaters ran out the rest of max_wait + soak (45 min at my settings) and the printer stayed in `busy_printers`, blocking everything behind it. The routes now signal the scheduler and the waits are sliced, so it aborts within seconds.
2. An aborted dispatch left the heaters on. Preheat now records what it sent and unwinds it on any non-success exit.
3. A chamber-heated print whose file carries no bed temperature skipped preheat entirely and started against a cold chamber. It now heats the bed to the configured chamber-heating temperature; a parsed bed temperature still wins, and a print with no chamber requirement still skips.

Your point about the unbounded window pushed me to check the thermal side properly rather than guess, so the soak-history logic is now derived from measurements off my X1C (cooling 55 °C → below 48 °C takes 23–73 minutes, ~0.2 °C/min): a stale history credits nothing, and a sub-target reading only counts as real cooling once it outlasts a grace period, since a plate swap briefly opens the chamber during exactly the window this feature runs in.

Locally green after the rebase: ruff check + format, i18n parity, the full backend suite (8610 passed / 3 skipped), ESLint, `tsc --noEmit`, and the frontend Vitest suite.

Also noted on the process — issue first, wait to be assigned, then the PR. Understood for next time, and sorry for the noise on that.

maziggy (changes requested):
A few things left.

Important

1. The settings copy contradicts the behaviour, in all 13 locales. en.ts:2213 says
   "maintain the bed at the next item's target temperature", but the hold is
   max(item_bed, queue_keep_warm_bed_temp) at :3646-3647 -- so an ASA file that
   says 100 C is held at 100, while one that says 80, or carries no bed metadata
   at all, is held at 90, hotter than the copy promises. The wiki PR gets this
   exactly right ("runs at the Keep-warm bed temperature ... raised to the print's
   own value whenever that is higher"); the settings page needs the same sentence,
   and the other 12 locales mirror the English.

2. Library-file items do trigger keep-warm, which both the description and the
   wiki say they do not. Nothing in _apply_keep_warm requires an archive -- it is
   consulted only for a higher bed temperature (:3646) -- so a library-file item
   assigned to a printer, with chamber-needing filament loaded, holds the bed at
   90. I would drop the claim rather than add the gate: chamber need comes from
   the AMS load, not the item, so the current behaviour is the more useful one.
   Please fix it in both places.

3. _cancelled_dispatches never shrinks for items that never dispatched. The routes
   call notify_dispatch_cancelled() unconditionally (print_queue.py:1562, 1862,
   1979), but the only removal is the discard() in _dispatch_one's finally (:1449).
   Cancelling or deleting a pending item -- the common case, and one id per item in
   cancel_batch -- leaves the id in the set for the life of the process. It is not
   a correctness bug today (queue-item ids are never reused, and both revert-to-
   pending paths at :5148 and :5408 sit inside a live dispatch that clears the
   flag), but it is an unbounded set on a process that runs for months. Signal
   only ids the scheduler is actually dispatching, or prune on read.

Nits

- pin.add("airduct") (:3997, :4060) fires whenever the current mode is not heating,
  including None, which means "no mode observed yet" -- and the rollback then forces
  cooling, asserting a state we never saw. Same shape as the aux-fan point last
  round, much smaller blast radius: record the mode you replaced and restore that.
- _sample_chamber_temps evicts _chamber_history and _keep_warm for unregistered
  printers (:3736-3744) but not _preheat_pin.
- _release_keep_warm pops the entry (:3510) before reading status, then returns when
  status is None, leaving the bed hot with nothing tracking it and the timeout no
  longer applying. Same for the _keep_warm eviction, which drops entries without
  releasing.
- The dip-close credit starts at the last below-threshold sample (:3814), so each
  recovered dip over-credits by one sampling interval. The opening side deliberately
  errs the other way and says so; worth a word here too.
- SettingsPage.tsx:4799 inline default ("Chamber-heating bed temperature (C)")
  disagrees with en.ts ("Keep-warm bed temperature (C)"), and t('settings.keepBedWarm')
  is the only one of the six keys without a defaultValue.
- The /certs/ .gitignore entry is unrelated to the feature. Harmless -- nothing is
  tracked under certs/ -- but it is outside what the PR says it does.

One scope note rather than a defect: the cancel flag is only consulted inside
_preheat_sleep, so a cancel landing during the FTP upload still uploads and starts
the print -- _start_print checks status once at entry (:4478) and never again. That
is pre-existing and fine to leave, but "abandons the dispatch" in the description
reads broader than what it does. Worth narrowing the wording to the preheat window.

ticfinack:
All addressed. Also rebased onto current `dev` — the branch had picked up a `dev` merge since the last round, so this is one commit on top of `1748d7ce` rather than a merge chain. Two of the conflicts were real: the budget-reservation work now shares `__init__` and `_dispatch_one`'s `finally` with the keep-warm state, and both sides are kept.

**Important**

1. **Settings copy.** Rewritten in all 13 locales to say what the code does — held at the keep-warm temperature, or the print's own bed temperature when that is higher. I used the wiki's phrasing as you suggested, and the other 12 are real translations rather than the English string.

2. **Library-file items.** You were right, and I took the second option: the claim is gone from both the PR description and the wiki, and the behaviour is unchanged. Chamber need comes from the AMS load, so gating on the item would have made it less useful for no benefit.

3. **`_cancelled_dispatches`.** Now only records ids with a dispatch actually in flight, so it is bounded by the upload pool rather than growing once per cancelled item. The rest are safe to drop for the reason you identified — `_claim_for_dispatch` only claims rows that are still `pending`, and the routes commit a terminal status before signalling. Covered by a test asserting a merely-pending cancel records nothing.

**Nits**

- **Airduct.** Both sites now pin a rollback only when the flap was observed in cooling. `None` means no mode has been reported, and rolling that back to cooling asserted a state we never saw — same reasoning as the aux fan, agreed.
- **`_preheat_pin`** is evicted alongside `_chamber_history` and `_keep_warm`.
- **`_release_keep_warm`** no longer pops before it can act. It went slightly further than the literal note, because the same shape appeared twice: an unreachable printer and a failed publish both now keep the entry, so the timeout keeps applying and a later tick retries instead of leaving the bed hot untracked. Dropping without releasing is now confined to `_sample_chamber_temps`, where the printer has actually left the manager and there is no client to send M140 to. Two tests cover it.
- **Dip close.** Documented at the credit line — it over-credits by up to one sampling interval, the opposite lean to the bracketing above it, and both are bounded by the sample cadence.
- **SettingsPage.** The inline default now matches `en.ts`, and all six keys have a `defaultValue`.
- **`/certs/`** reverted; the PR no longer touches `.gitignore`.

**Scope note** — taken, and the description now says the cancel flag covers the preheat window only, and that a cancellation arriving during the FTP upload still starts the print, as before.

Green after the rebase: `ruff check` + `ruff format --check`, i18n parity across 13 locales, the full backend suite (9281 passed / 3 skipped), ESLint, `tsc --noEmit`, and `npm run test:run`.

maziggy (approved):

PR #2740, opened 2026-08-01
Add per-job ETA to print queue

### LINKED ISSUE #2736: [Feature]: Show per-job ETA in the print queue
## Problem

The print queue shows the estimated duration of each queued job, but not
the estimated clock time at which that individual job would finish if
started now.

The currently printing card already shows an ETA. Showing the same
information beside queued jobs would make the queue easier to scan.

## Proposed behavior

For every pending or staged queue item with a known print duration, show:

ETA = current local time + this job's own print duration

This is intentionally not a cumulative queue forecast. It answers:

"If I start this job now, when will it finish?"

Blocked items without a reliable duration should not show an ETA.

## Implementation

I have tested a local implementation that:

- displays the ETA beside the existing print duration
- includes staged/manual-start jobs
- uses the existing `formatETA` function
- respects the configured 12/24-hour time format
- updates as the queue page refreshes
- passes the frontend production build

Please assign this issue to `mpl1337` if the approach is acceptable.
I will then open a focused PR against `dev` with before/after screenshots.

### Proposed Solution

Display an ETA beside the existing print duration for each pending or staged queue item.

The ETA should be calculated as:

current local time + the individual job's print duration

The existing `formatETA` helper and the user's configured 12/24-hour time format should be reused. Blocked items should not display an ETA.

### Alternatives Considered

A cumulative queue ETA was considered, but staged jobs require manual starts and may not be released immediately. A per-job ETA is therefore clearer and more reliable.

### Feature Category

Print Queue & Scheduling

### Priority

Would improve my workflow

### Mockups or Examples

Image

### Contribution

### Checklist

mpl1337:
## Summary

Adds a per-job ETA to pending and staged print queue items.

For each item with a known print duration, the displayed ETA is calculated as:

`current local time + the individual job's print duration`

This is intentionally not a cumulative queue forecast.

Closes #2736

## Changes

- display an ETA beside the existing print duration
- include staged/manual-start jobs
- reuse the existing `formatETA` helper and configured 12/24-hour format
- omit the ETA for blocked/waiting items
- add unit tests for the queue ETA helper

## Testing

- `npx vitest run src/__tests__/utils/queueEta.test.ts`
- `npm run lint`
- `npm run build`

All completed successfully.

## Documentation

No companion documentation PR is included because this is a small display-only UI enhancement with no new settings, API changes, or workflow changes.

## Screenshots
image

image


maziggy (changes requested):
Blocker: a printing job ends up with two contradictory ETAs.

SortableQueueItem is not pending-only - the Currently Printing section renders with the same component (QueuePage.tsx:2417). The new block is gated only on !item.waiting_reason, so an in-progress print gets a per-job ETA computed from its full duration, right above the progress block at :711 which shows the real ETA from the printer's remaining_time.

Rendered against a 50% done print (print_time_seconds 7200, remaining_time 60):

ETA 01:01 PM   <- new block, now + 2h (full duration)
ETA 12:01 PM   <- existing progress block, now + 60m (correct)
Both use text-bambu-green font-medium, so they look identical. The gate wants to be isPending (already in scope at :642) rather than !waiting_reason.

Scheduled items have the same problem. Line 642-650 already renders the scheduled start on that row ("in 5h", "Overdue", "ASAP"). A job scheduled for tonight shows "in 5h" next to an ETA of now + 2h - a completion time earlier than its start. Either skip when scheduled_time is set, or base the ETA on max(now, scheduled_time). The latter is still per-job rather than a cumulative forecast, so it stays inside the scope you set out in #2736.

Dangling label. The outer guard is item.print_time_seconds &&, which is truthy for a negative value; formatQueueItemETA then returns null and the row renders a bare "ETA" with nothing after it. Worth gating the span on the result rather than on the raw field.

Suggested shape for all three:

{(() => {
  if (!isPending || item.waiting_reason || item.scheduled_time) return null;
  const eta = formatQueueItemETA(item.print_time_seconds, timeFormat, t);
  if (!eta) return null;
  return (
    <span className="text-bambu-green font-medium" title={t('queue.etaIfStartedNow')}>
      ETA {eta}
    </span>
  );
})()}
Tests. The unit test covers a three-line wrapper, and its one positive assertion compares formatQueueItemETA against formatETA itself, so it pins the seconds to minutes conversion but cannot fail on a wrong output. Nothing covers the actual change - which is why all 30 existing QueuePage.test.tsx tests still pass with the double-ETA above. That file already has pending, printing and completed fixtures with print_time_seconds set, so a test asserting one ETA on a pending row, exactly one on a printing row, and none on a waiting row would drop in easily and would have caught this.

Minor:

- The tooltip reuses printers.estimatedCompletion ("Estimated completion time"). Accurate for a live print, overstated for a queued one where the claim is really "if started now". A queue.* key would read better, but it needs real translations in all 13 locales, so only if you want to.
- utils/queueEta.ts is a new module plus a new test file for three lines that would sit naturally next to formatETA in utils/date.ts.
- t?: Parameters<typeof formatETA>[2] is doing a lot of work for (key: string) => string.

Don't worry about the rebuilt static/ bundle or the CHANGELOG entry - I handle both on merge.

mpl1337:
Thanks for the detailed review. I fixed the blocker by limiting the
if-started-now ETA to pending, non-waiting, non-scheduled items with a
valid positive duration.

Active prints now show only the existing live ETA based on remaining_time,
and invalid durations cannot render a dangling ETA label.

I also removed the small wrapper module and added QueuePage rendering tests
covering pending, staged/manual-start, printing, waiting, scheduled, and
invalid-duration rows.

Validation completed successfully with:
- npm run test:run
- npm run lint
- npm run build
- git diff --check
- manual browser verification on a live BamBuddy instance with active and staged jobs

maziggy (changes requested):
Blocker 1 — the "blocked" guard misses the most common queue layout

QueuePage.tsx:433-440 suppresses the ETA when item.waiting_reason is set. But waiting_reason is written in exactly one place — print_scheduler.py:745, inside the elif item.target_model: branch. The fixed-printer_id path never sets it: when the assigned printer is busy it just does busy_printers.add(...); continue (print_scheduler.py:617-618), leaving the row pending with waiting_reason = None.

So jobs pinned to a specific printer — the default when you add a print to a printer — are never marked blocked. I rendered three jobs on one printer, one running and two queued behind it:

pendingEtasShown: [ "ETA 01:00 PM", "ETA 01:00 PM" ]
reality:            B ends ~11:54 UTC, C ends ~12:54 UTC

Both queued jobs claim the same completion time, and both are wrong — the second by roughly two hours. The PR's own comment says "do not show it for active, blocked, scheduled, or invalid queue items"; the blocked case is only covered for model-targeted items.

Blocker 2 — the ETA freezes at first render

formatETA is called during render off new Date(), and nothing re-renders the row while the queue is idle. React Query's structural sharing hands back a referentially identical data when the payload doesn't change, so the poll fires but the component never updates. Measured over 45 simulated minutes:

ETA at T+0    : ETA 01:00 PM
ETA at T+45m  : ETA 01:00 PM      (queue polled 541 times)

Leave the queue page open and the number silently drifts into nonsense. Worth noting the contrast with the live ETA at QueuePage.tsx:719-720, which looks like the same pattern but is safe for an incidental reason — remaining_time changes on every poll, so that row always re-renders.

Important — the tooltip asserts something false

title={t('printers.estimatedCompletion')} renders "Estimated completion time" in all 13 locales. That is precisely what this value is not. The whole premise of the PR is that it's a hypothetical, and the tooltip is the one place that could say so — instead it tells the user the opposite. Needs its own queue-namespaced key, translated properly everywhere.

Minor: require_previous_success items get an ETA even though the scheduler may skip them outright — same class as the exclusions already made. And the hardcoded ETA  string matches existing usage at QueuePage.tsx:720 and PrintersPage.tsx:3730, so I'd leave it.

Issue #2725, opened 2026-07-30
[Feature]: dropdown/split-button "open in local slicer" in 3d preview


pascalheidmann:
### Problem or Use Case

In filemanager I want to open a file to slice it with custom config instead of using the predefined cli server slicer api. In this case I have to download the file instead of just opening it directly.

### Proposed Solution

If api slicer is configured, instead of single button add a dropdown secondary option to open the file in your local slicer. This doesn't even have to rely on your default local slicer but give links to all available slicers, and same secondary option can be applied to local slicers, too (if your users have different slicers installed).

### Feature Category

UI/UX

### Priority

Would improve my workflow

### Mockups or Examples

<img width="205" height="186" alt="Image" src="https://github.com/user-attachments/assets/88697add-2b1f-4bb5-af57-4ee9e9cd3844" />

### Contribution

- [x] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Please note our rules from CONTRIBUTING.md. Thanks!

maziggy:
Thanks.

PR #2763, opened 2026-08-04
Feature/slicer multi button

pascalheidmann:
## Description

Allow to open files in your local / preferred slicer as secondary option next to the slicer api sidecar

## Related Issue

Implements #2725

## Documentation

**Pick one**:
- [ ] Docs PR(s) linked above
- [x] No docs update required — reason: minimal improvement without any documentation

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Filemanager: added switch to open in local slicer if no api slicer is defined
- Filemanager > 3D preview: secondary option to open in both bambu and orca slicer regardless of preferred slicer and api slicer configration

## Screenshots

Variant: no api slicer defined

Variant: with api slicer

## Testing

## Checklist


maziggy (changes requested):
Thanks for this. It is a good fit: ArchivesPage already does exactly this icon-swap-and-branch (useSlicerApi ? <Cog/> : <ExternalLink/>, useSlicerApi ? setShowSliceModal : openInSlicerWithToken), and the File Manager was the one place that never got it. Your version is also better than the original, because you gate the handoff on library:read instead of leaving it ungated, which is the permission the backend actually enforces on the slicer-token endpoint. The split button is a nicer answer to #2725 than a setting would have been.

One heads-up before the notes: CI did not run on this PR. ci.yml triggers only on PRs targeting main, so the seven green checks are all security scanners. I applied your diff to a scratch tree and ran the rest myself: vitest passes (84, including your 11), eslint clean, npm run build clean, i18n parity clean at 5807 leaves across all 13 locales. So nothing is broken. The following are design points.

Blocking:

1. The render gate and the click handler disagree. FileManagerPage.tsx renders the item on (onSlice || onOpenInSlicer) but dispatches on useSlicerApi. A caller that passes only onSlice with the API off renders an enabled item that does nothing when clicked. Both are passed today, so it is latent, but please make the condition mirror the dispatch: (useSlicerApi ? onSlice : onOpenInSlicer).
2. The same permission rule is written twice in two shapes: sliceDisabled inside FileCard (from the useSlicerApi prop) and canSlice in the page (from settings?.use_slicer_api), one the inverse of the other. Those will drift. Please pick one and pass it down.
3. The two file-type rules contradict each other. ModelViewerModal's canOpenInSlicer restricts library handoff to 3mf, but the new File Manager path uses isSliceableFilename, which also allows stl/step/stp. So the card menu offers a desktop handoff for an STL while the 3D preview of that same STL shows "Open in Slicer" greyed out. Please make them agree. My guess is the 3mf-only rule is the over-cautious one, since both slicers accept an STL over the URL handler, but I have not tested it.

Non-blocking, would like:

4. SlicerSplitButton takes a disabled prop that neither call site passes, and the button only renders when the action is already possible. Please drop it.
5. The catch in handleOpenInSlicer hands the slicer an unauthenticated download URL. With auth on and a 403 on the token request, the slicer opens and fetches a 401 and the user sees nothing, so a permission denial looks identical to "no slicer installed". It is the existing house pattern so I am not going to hold the PR over it, but this is a new call site behind a menu item labelled "Slice", where people expect a modal. A toast in the catch would help.
6. slicerDropdownItems is built above handleOpenInSlicer. It works because the reference sits inside a closure, and it compiles and lints, but it reads as a bug. Please move the handler up.
7. Nit: settings?.open_in_slicer || settings?.preferred_slicer || 'bambu_studio' is now the fourth copy. Worth lifting into utils/slicer.ts while you are in here.

Tests: the ones you added are behavioural rather than snapshot, which I appreciate. The gap I would most like closed is the Escape interaction. The e.stopPropagation() in the dropdown's document keydown listener is load-bearing: it stops Escape reaching the modal's own window listener, and it only works because document precedes window in the bubble chain. Nothing tests it, and if either listener ever moves to the same target it breaks silently, because stopPropagation does not stop same-element listeners. Also missing: outside-click close, the list-view button (only the card menu is covered), and the permission change itself, which is the security-relevant part of the PR.

Small one: openMenu grabs card.querySelector('button'), so it takes whichever button comes first. It passes now, but it breaks the moment a button lands ahead of the kebab.

Docs: this one does need them. docs/features/slicer-api.md currently says the sidecar is opt-in and that without it all "Open in Slicer" flows are unchanged, and that the Slice button appears on file cards as a consequence of enabling the sidecar. Both are now inaccurate. docs/features/file-manager.md does not mention the action at all, and the new dropdown in the 3D preview is undocumented. No CHANGELOG needed.

No changelog entry needed from you either.

pascalheidmann:
@maziggy thank you for your extensive feedback! As suggested I simplified the handling of the several potential file types and also added retrospective the same handling to MakerworldPage.tsx as it used the same condition.

One sidenote: when I was revamping the permissions I stumbled upon an issue with the current read permission being more granular and classic read is only in older, pre-configured access levels but in new installation shouldn't be there anymore. Therefore I used the hasAnyPermission, but if someday the old permission is removed this should be cleaned up, too ;)

Hope this now is ready to merge, but any other improvement/suggestion is of course welcome!

PR #2803, opened 2026-08-10
Map each printer against its own AMS, and stop dispatching a job whose filament is not loaded (#2799)

### LINKED ISSUE #2799: [Bug]: AMS mapping computed for one printer is reused on all printers in a multi-printer queue (wrong filament printed)
## Summary

When one job is queued to several printers at once, the AMS mapping computed for one printer is reused verbatim on all the others. Printers whose AMS slot order differs then print with whatever sits in those slot indices — in our case **ASA where the 3MF strictly requires PETG**.

A second, related issue: when the matcher cannot satisfy a required filament it emits `-1` and the job is dispatched anyway, letting the printer pick.

## Environment

- Bambuddy: `ghcr.io/maziggy/bambuddy:latest` (Docker, RPi / arm64)
- Printers: 3× Bambulab P2S, firmware `01.11.02.00`, each with one AMS
- Job queued to P2S-4, P2S-5 and P2S-6 from the same dialog

## What happened

The 3MF (plate 22) requires two filaments:

| 3MF slot | type | color | usage |
|---|---|---|---|
| 1 | PETG | `#76D9F4` (light blue) | 406 g — body |
| 3 | PETG | `#2850E0` (dark blue) | 1.9 g — lettering |

The three printers have the same spools loaded in **different slots**:

| slot | P2S-4 | P2S-5 | P2S-6 |
|---|---|---|---|
| T0 | ASA `#161616` | PETG `#76D9F4` | PETG `#76D9F4` |
| T1 | PETG `#2850E0` | ASA `#161616` | ASA `#161616` |
| T2 | PETG `#76D9F4` | PETG `#2850E0` | PETG `#2850E0` |
| T3 | ASA `#898989` | ASA `#898989` | ASA `#898989` |

All three queue items were created in the same second with an **identical** mapping:

```
item 274 | printer 9  (P2S-4) | ams_mapping [2, -1, 1]
item 275 | printer 10 (P2S-5) | ams_mapping [2, -1, 1]
item 276 | printer 8  (P2S-6) | ams_mapping [2, -1, 1]
```

`[2, -1, 1]` is correct for P2S-4 only. On P2S-5/P2S-6 it resolves to:

- filament 1 → T2 = PETG **dark blue** (wrong shade)
- filament 3 → T1 = **ASA black** (wrong material entirely)

Dispatch log for P2S-6:

```
Sending print command: {"print": {..., "use_ams": true,
  "ams_mapping": [2, -1, 1],
  "ams_mapping2": [{"ams_id": 0, "slot_id": 2}, {"ams_id": 255, "slot_id": 255}, {"ams_id": 0, "slot_id": 1}]}}
PRINT START printer 8: mapping=[2, 65535, 1]
PRINT START printer 8 AMS 0: T0(type=PETG, color=76D9F4FF), T1(type=ASA, color=161616FF),
                             T2(type=PETG, color=2850E0FF), T3(type=ASA, color=898989FF)
```

P2S-4 printed correctly; P2S-5 and P2S-6 produced parts in the wrong blue with ASA lettering. Note that an explicit `ams_mapping` bypasses the printer's own type check, so nothing downstream catches this — the operator only finds out by inspecting the running print.

## Root cause

`backend/app/services/print_scheduler.py`, both dispatch paths (`check_queue`, around lines 383 and 512):

```python
# Compute AMS mapping if not already set
if not item.ams_mapping:
    computed_mapping = await self._compute_ams_mapping_for_printer(db, item.printer_id, item)
```

The per-printer matcher only runs when the mapping is empty. The print dialog stamps one mapping onto every selected printer, so the guard is always false and the mapping is never validated against the target printer's live AMS.

`_match_filaments_to_slots` itself is fine — it matches on `tray_info_idx` → exact color → similar color → type, with type as a hard filter, and is slot-order independent. Replaying this incident against it with each printer's real AMS gives the correct answer for every printer:

```
P2S-4  AMS: T0=ASA/161616, T1=PETG/2850E0, T2=PETG/76D9F4, T3=ASA/898989  -> [2, -1, 1]
P2S-5  AMS: T0=PETG/76D9F4, T1=ASA/161616, T2=PETG/2850E0, T3=ASA/898989  -> [0, -1, 2]
P2S-6  (same as P2S-5)                                                     -> [0, -1, 2]
```

So the matcher would have prevented this had it been allowed to run.

## Second issue: unmatched requirement is dispatched anyway

`_match_filaments_to_slots` (around line 1422):

```python
else:
    comparisons.append({"slot_id": req.get("slot_id", 0), "global_tray_id": -1})
```

A requirement with no matching spool becomes `-1` in the mapping and the job is dispatched, leaving the choice to the printer. Replay with the dark blue spool removed:

```
AMS: T0=PETG/76D9F4, T1=ASA/161616, T2=ASA/898989, T3=ASA/FFFFFF  -> [0, -1, -1]
  filament 3 (PETG #2850E0) -> no match, printed anyway
```

`_block_on_filament_deficit` already implements exactly the right behaviour (promote to `manual_start`, notify, wait for the operator) but only triggers on a grams shortfall, not on a missing type/color.

## Suggested fix

1. Compute the AMS mapping for the printer that will actually run the item, and overwrite a mapping that was computed for a different printer (or at minimum validate the stored mapping against the target printer's AMS and recompute on mismatch). A stored mapping is only trustworthy for the printer it was computed for.
2. When the computed mapping still contains `-1` for a slot that the 3MF requires, hold the item instead of dispatching — same treatment as the filament deficit gate.

Happy to open a PR if the approach looks right. We're running both changes locally as a patch and they behave as expected.

## Note on `type_only_match`

Independently of the above: the final fallback in `_match_filaments_to_slots` matches on type with any color, so a missing exact color silently substitutes a different color of the same type. For a farm that's the same class of surprise as this bug, just less severe. Making that fallback opt-in (or at least surfacing it as a warning on the queue item) would help.

grolmus:
## Description

Queueing one model to several printers sent all of them the same `ams_mapping`. A mapping is a list of global tray IDs, and those only mean something relative to the printer they were resolved against, so the copies landed on printers whose AMS holds the same spools in different slots — the slot index still resolved, nothing looked wrong, and the job printed from whatever sat in that tray. In our farm that put ASA into a slot the 3MF strictly requires PETG. An explicit `ams_mapping` bypasses the firmware's own type check, so nothing downstream catches it.

Three changes, in the scope agreed in #2799:

1. **Frontend** — every selected printer now maps against its own live AMS, and the match badge is derived from the same decision that produces the submitted mapping.
2. **Backend** — a stored mapping is re-validated against the target printer at dispatch and recomputed when it does not fit.
3. **Backend** — a job whose required filament is not loaded is held instead of dispatched.

## Related Issue

Fixes #2799

## Documentation

**Pick one**:
- [x] No docs update required — reason: bug fix, no new setting or user-visible surface. The one new behaviour (a job held for missing filament) reuses the existing `manual_start` + queue-waiting notification path that the filament-deficit gate already documents. Two `CHANGELOG.md` entries are included, and they state the two ceilings you asked to see in the release notes rather than in silence: a held job waits for a human press even once the spool is back, and its reason is English whatever the interface language.

## Type of Change

- [x] Test addition or update

## Changes Made

**`frontend/src/hooks/useMultiPrinterFilamentMapping.ts`** — one `effectiveMappings` per printer decides which overrides legitimately apply to it (its own, the shared ones if they were authored against it, otherwise none), and `finalMapping`, `matchDetails` and `allPrintersReady` all hang off that single value, so the badge and the submitted mapping cannot drift apart. An unknown owner (`null`) counts as nobody.

**`frontend/src/components/PrintModal/index.tsx`** — `getMappingForPrinter` returns `getFinalMapping(printerId)` for a multi-printer selection, and on the single-printer path returns the shared `amsMapping` only when `sharedMappingPrinterId` is unset or is that printer. Both paths now say the same thing rather than one of them resting on the effect that clears `manualMappings`.

**`backend/app/services/print_scheduler.py`**
- `_stored_mapping_conflict()` — for each slot the plate actually prints, checks the mapped tray exists on the target printer and holds the required filament type. Returns a short reason, or `None` when the mapping fits *and* whenever there is not enough evidence to judge, so a recompute only ever follows a positive finding. The cheap existence pass runs on live status alone; the 3MF is opened only if that passes, and the parse is memoised per scheduler pass.
- `_ensure_ams_mapping()` consults it before trusting a resolved stored mapping, clears the mapping it rejected before recomputing, and short-circuits on `skip_filament_check`.
- `_block_on_unmatched_filament()` — promotes to `manual_start` and notifies through the existing `_block_on_filament_deficit` machinery. Keyed off unresolved slots in the *computed* mapping intersected with the plate's requirement list, which inherits the matcher's per-nozzle restriction for free.

As agreed, `type_only_match` is left alone here — separate issue to follow.

## Third review round

**1. The complete miss no longer falls through to the 0700_8012 path.** `_block_on_unmatched_filament` used to bail on `not item.ams_mapping`, so a rejected mapping whose recompute resolved nothing dispatched with no mapping at all. It now reads a missing mapping as "no slot resolved" — but only when live status positively reports loaded trays, which is what separates it from #2589: there the mapping is bogus because the AMS was not known yet, and that cold status has an empty loaded list. Three tests: the hold fires, the cold status dispatches, no status dispatches.

**2. A held model-based job gives its printer back.** Taking the first of the two options: `_block_on_unmatched_filament(..., release_assignment=True)` on the model-based path clears `printer_id`, so the next press of ▶ looks for a printer that can run the job instead of the one that could not. The mapping goes with the assignment — its tray IDs were resolved against the printer being released, and keeping them would hold the job again even on a printer that has the filament, since an unresolved slot is deliberately not a conflict worth recomputing over. The printer-assigned path keeps its printer: that one was the user's choice, not the scheduler's. The notification is still looked up against the printer that held it.

**3. The single-printer path consults the ownership record**, so the rule is stated on both paths rather than resting on the clearing side-effect at `index.tsx`.

**Smaller:** a JSON `true` is no longer read as tray 1 — `bool` is an `int`, so one `_is_tray_id` helper closes it in the conflict check and in the hold. `_stored_mapping_conflict` handed a non-list from a corrupted row now has a test. Both release-note ceilings are in `CHANGELOG.md`, together with the held-job-needs-a-press contract.

## Screenshots

No visual change — `getMappingForPrinter` feeds submission, not rendering. The per-printer panel and its override UI are untouched.

## Testing

- [x] I have tested with my printer model: P2S (×3, one AMS each), alongside X1C and H2D on the same instance

`backend/tests/unit/test_ams_mapping_per_printer_2799.py` — 38 tests built on the real incident: the 3MF needs light-blue PETG in slot 1 and dark-blue PETG in slot 3; P2S-4 holds them in trays 2 and 1, P2S-5 in trays 0 and 2 with ASA in tray 1. `[2, -1, 1]` is correct on the first and feeds ASA into slot 3 on the second.

`frontend/src/__tests__/hooks/useMultiPrinterFilamentMapping.test.tsx` — 6 tests over `printerResults`: per-printer mapping, no lending, kept on its owner, badge derived from the submitted mapping, unknown owner, and a printer that cannot satisfy a slot.

```
pytest backend/tests/       10838 passed, 3 skipped, 1 error (see below)
ruff check backend/         All checks passed!
ruff format --check         807 files already formatted
npx tsc --noEmit            clean
npm run lint                clean
```

The one error is a `KeyError: 'access_token'` in the setup of `tests/integration/test_ownership_permissions.py::test_bulk_delete_skips_non_owned_files`, under `-n 12`. That file passes 86/86 in isolation, and nothing in it is near this change — the same load flakiness you saw in one of your three runs.

On `Layout.test.tsx` "MakerWorld nav link": you were right that it is not a pre-existing failure. On a clean `npm ci` it passes 26/26 in isolation here too. It failed once in a full parallel run and did not recur, so it belongs in the same bucket as the above rather than being carried forward as a known break. Withdrawing that claim from the two earlier rounds.

Field-verified before opening this: replaying the incident 3MF through the real `_match_filaments_to_slots` against each printer's actual AMS gives `[2, -1, 1]` for P2S-4 and `[0, -1, 2]` for P2S-5 and P2S-6 — i.e. the matcher had the right answer all along and only needed to be allowed to run per printer. With the dark-blue spool removed it returns `[0, -1, -1]`, which is the case the hold catches.

Merged `dev` (40 commits) before this push; `PrintModal/index.tsx`, `useFilamentMapping.ts` and `print_scheduler.py` all moved in it, so the merge is re-verified rather than assumed. Unrelated: `dev` carries an empty file named `assert` at the repo root, added in cfecfa36 — looks like a stray shell redirect rather than anything intended.

## Checklist


maziggy (changes requested):
Read all four files end to end and traced both dispatch paths. Locally: ruff clean, the new test file passes -- though it collects 18 tests, not the 21 in the description, worth a
check that nothing was lost in the merge.

The backend design is right, and the part I expected to be wrong is not: every "cannot
judge" path returns None or False, so a recompute or a hold only ever follows a positive
finding, and each of those paths has its own test. Clearing the rejected mapping before
recomputing is the correct order. The `slot_id - 1` indexing matches `mapping[slot_id - 1]` at print_scheduler.py:2964, and using `_canonical_filament_type` on both sides keeps it consistent with the matcher.

Three things I want changed before this goes in, two of them in the frontend half.

## Must fix

**1. The per-printer badge and the mapping actually submitted now disagree.**
`useMultiPrinterFilamentMapping.ts:349` still computes `matchDetails` from
`effectiveMappings`, which for a `useDefault` printer is the shared `defaultMappings`.
Submission now takes `autoMapping`, which by design ignores them. So the panel can show 3/3 exact -- derived from the first printer's manual tray IDs -- while the job goes out mapped differently, and `allPrintersReady` gates on the same stale number. For a
`useDefault` printer `matchDetails` needs the same input `autoMapping` gets: no overrides. `getAutoMapping` was added next to `getFinalMapping` in a2e781be, but the badge path was not touched, which is where the two diverged.

**2. A manual mapping is discarded for the printer it was valid on.** 
The shared `FilamentMapping` panel renders only at `selectedPrinters.length === 1` (index.tsx:1669), so `manualMappings` is always entered against one specific printer. Add a second printer and the panel disappears and index.tsx:828 returns `getAutoMapping` for both -- including the printer those edits were resolved against, where they were correct. Dropping them on the newly added printer is the fix. Dropping them on the original one is a regression, and a silent one, since the panel that showed them is gone by then. Either keep `getFinalMapping` for the printer the shared mappings belong to, or clear `manualMappings` when the selection grows past one so the UI stops implying they still apply.

**3. The hold is nozzle-blind, so it misses its own case on H2D and H2C.**
`_get_missing_filament_types` (print_scheduler.py:1944) scans every AMS tray on the printer with no `nozzle_id` filter, while `_match_filaments_to_slots` applies the per-nozzle restriction as a hard filter when no FTS is installed. A dual-nozzle printer with PETG loaded only on the other nozzle's AMS therefore reports nothing missing, the matcher still returns -1 for that slot, and the item dispatches -- exactly what the gate exists to prevent. You have an H2D on the same instance, so this is reachable on your own farm.

Your objection to keying off -1 is right as stated, but it dissolves if you intersect with
the requirement slot list, which `_stored_mapping_conflict` already does at :2189. Keying the hold off unresolved *required* slots in the computed mapping gets nozzle-awareness for free, because the mapping came from the nozzle-aware matcher.

## Worth fixing

**4.** `_ensure_ams_mapping`'s docstring (:2064) still says a resolved mapping "including manual overrides, or a partially padded one, is left untouched". That is precisely what this PR stops doing.

**5. A partially unresolved mapping is now destroyed rather than preserved** (:2196). `"slot {n} is unresolved"` drops the whole stored mapping, including slots the user resolved by hand, and recomputes all of them. A -1 on a required slot is not evidence the mapping was resolved against another printer -- it is evidence the matcher had nothing, which the new hold handles separately. I would drop that branch.

**6. External spool (>= 254) skips the type check** (:2198)
which leaves this PR's own bug class open on that path. `_build_loaded_filaments` does report the external tray's type (:2582), so the check is available. Two printers with different filament in the external feed and one stamped mapping is the same failure you are fixing for AMS trays. Either check it, or add it to the "one limit worth stating plainly" paragraph, which currently mentions only colour.

**7. `_stored_mapping_conflict` parses the source 3MF on every tick, for every pending
item.** `_get_filament_requirements` opens the zip and parses `slice_info.config` on each call with no cache (filament_requirements.py:28); the old fast path never touched disk. At the 3 s post-dispatch cadence with a long queue that is a new repeated cost, and the conflict path then parses a second time inside  `_compute_ams_mapping_for_printer`. Checking
tray existence first -- status only, no disk -- and parsing only if that passes would cost
nothing.

## Nits

- `waiting_reason="filament_missing"` is a machine token interpolated raw into the
  notification body (notification_service.py:2115); the seeded template example is a human   string. `missing` is in scope, so "needs PETG" is one line. The deficit gate has the same wart -- no need to fix that one here.
- No UI surface for the hold. The deficit gate sets `filament_short`, which QueuePage
  renders as a badge and the Print Anyway route clears. This one sets only `manual_start`, so with notifications off a job stops being automatic with nothing on screen saying why. `waiting_reason` already exists on the model (print_queue.py:29) and in the schema.
- No frontend tests, and `useMultiPrinterFilamentMapping` has no test file at all. Points 1 and 2 are both what a test over `printerResults` would have caught, and the frontend half is the part that prevents the bug at source rather than catching it later.

grolmus:
All seven addressed in 20e3f359, on top of your dev merge. Test count: you were right, it collected 18 — I miscounted when writing the description. It is 24 now, and the description is corrected.

## Must fix

**1 + 2 — badge/submission drift, and the discarded manual mapping.** These turned out to be the same bug, so they get one fix rather than two patches.

The hook now derives one value per printer and everything hangs off it:

```ts
const ownsDefaultMappings =
  defaultMappingsPrinterId == null || printerId === defaultMappingsPrinterId;
const effectiveMappings = !config.useDefault
  ? config.manualMappings
  : ownsDefaultMappings ? defaultMappings : {};
```

`finalMapping`, `matchDetails` and `allPrintersReady` all read that, so the panel cannot promise one mapping and submit another — not because the two paths were re-synced, but because there is only one path left. `getAutoMapping` is gone; `getMappingForPrinter` is now just `getFinalMapping` for the multi-printer case.

For 2 I took the first option rather than clearing the edits. The shared panel only renders at `selectedPrinters.length === 1`, so the owning printer is known exactly — `sharedMappingPrinterId` is recorded in `onManualMappingChange`, and seeded from `queueItem.printer_id` in edit mode, where the stored mapping belongs to the printer the item is queued on. The edits stay on the printer they were correct for and do not follow the selection.

**3 — nozzle-blind hold.** You were right, and it was reachable on our own H2D. The hold no longer looks at loaded types at all; it reads unresolved slots out of the computed mapping, intersected with the plate's requirement list, exactly as you suggested. That inherits the matcher's per-nozzle restriction for free, and my objection to `-1` does dissolve once the intersection is there — the padding case is excluded by construction rather than by hoping. `required_filament_types` is no longer used by this path. `test_nozzle_bound_filament_on_the_wrong_side_holds` covers the H2D case.

## Worth fixing

**4** — docstring rewritten; it now says the mapping is re-checked and recomputed on a misfit, survives when it fits, and always survives "Print Anyway".

**5** — dropped. `-1` on a required slot no longer triggers a recompute, so a partially hand-resolved mapping keeps the slots the user resolved. The hold owns that case, and both methods now say so in their docstrings so the split does not read as an oversight.

**6** — external is type-checked like any other tray now, no `>= 254` early exit. One asymmetry kept deliberately: a mapping naming an external tray we have not heard about returns `None` rather than "not loaded", because an unreported feed is absence of evidence, not a foreign mapping. Two tests: the type check fires on a wrong-type external feed, and an unreported one does not.

**7** — split in two. A cheap pass over live status alone rejects any mapping naming a tray this printer does not have; only if that passes is the 3MF opened for the type check. `test_foreign_tray_is_caught_without_reading_the_3mf` asserts `_get_filament_requirements` is never awaited on that path, so the fast path cannot regress silently.

## Nits

- Notification and queue reason are both human now: `Needs PETG #2850E0`, built with the existing `_describe_filament`.
- UI surface: the hold writes `waiting_reason`, which QueuePage already renders for pending items — so it shows up without a schema change. That turned up a hole I would have introduced: the start route clears `manual_start` and `filament_short` but not `waiting_reason`, and a held item never reaches the gate again, so the row would have read "Waiting" forever. The gate clears its own message on the pass that lets the item through, matched by prefix so it never eats a reason set by the model-based path. Two tests.
- Frontend tests: `useMultiPrinterFilamentMapping.test.tsx`, 5 tests. Two of them are 1 and 2 as regressions — `does not lend one printer's shared mapping to another` and `reports a match badge derived from the mapping it will actually submit` both fail on the previous commit.

## Verification

```
backend/tests/unit                     6791 passed, 3 skipped
test_ams_mapping_per_printer_2799.py     24 passed
ruff check / format                    clean
tsc --noEmit / eslint                  clean
frontend npm run test:run              2988 passed, 1 failed
```

Same single pre-existing frontend failure as before — `Layout.test.tsx` "MakerWorld nav link", fails identically on a clean dev with this branch's commits stashed.

## Unrelated

Your reply about the colour matcher — nearest-wins, the four copies, Prefer Lowest as tie-break — landed as a comment on #2805 (the plate-clear trigger) rather than #2804. Answering it on #2804 so the thread stays findable. #2805's own two questions are still open whenever you get to them.

grolmus:
One more in `fea125c8`, found reviewing my own branch rather than reported — it reintroduced the bug through a path I had not considered.

`ownsDefaultMappings` treated a `null` owner as "every printer owns them":

```ts
defaultMappingsPrinterId == null || printerId === defaultMappingsPrinterId
```

That reads fine when nobody has edited anything, because the shared record is empty and applying it is a no-op. It is wrong in edit mode. `PrintQueueItem.printer_id` is nullable, and a model-based item has it null while its stored `ams_mapping` still seeds `manualMappings` — so opening a model-based queue item and selecting several printers handed one item's tray IDs to all of them, which is exactly what this PR exists to stop.

Now an explicit match, with `null` meaning nobody:

```ts
defaultMappingsPrinterId != null && printerId === defaultMappingsPrinterId
```

Unknown provenance falls back to per-printer auto-matching, which is the safe direction — a mapping we cannot attribute is one we cannot validate. `does not hand shared mappings to anyone when their owner is unknown` covers it, and fails on the previous commit.

grolmus:
Traced every path that writes `PrintQueueItem.ams_mapping` and every path that dispatches one, to convince myself the gate is actually closed rather than closed on the route I happened to be looking at. Three things worth reporting.

**The dispatch side is closed.** `printer_manager.start_print` has exactly one caller, `_start_print`, and the only two ways into `dispatch_ids` (`print_scheduler.py:1040`, `:1275`) both sit behind `_ensure_ams_mapping`. `webhook_start_print` clears `manual_start` and hands back to the scheduler rather than dispatching. `printers.py:690`/`:765` looked like a dispatch on a grep but is the status serialiser reporting the printer's own current mapping. So the only bypass is `skip_filament_check`, which is the one you asked for.

**A second live instance of this bug, in the VP queue path.** `virtual_printer/manager.py:1148` stamps the slicer's mapping onto the queue item while `printer_id=self.target_printer_id` on the line above — which is `None` for a model-based VP. So an "Any P2S" virtual printer writes the slicer's tray IDs onto an item the scheduler will hand to whichever P2S is free. Gated behind the per-VP `save_ams_mapping` opt-in and stood down by `queue_force_color_match`, so it needs those two settings to combine, but nothing in that path knows the mapping is unattributable. It is the same failure as #2799 arriving from a different direction, and `_stored_mapping_conflict` catches it at dispatch. Happy to add a test for that shape if you want it pinned here rather than left to the general check.

**Provenance already exists, just not on the queue item.** You said nothing records which printer a stored mapping came from — true for `PrintQueueItem`, but `archive_print()` takes `slicer_ams_mapping_printer_id` and stores it beside the mapping (`archive.py:1288`), and the reprint fallback refuses to reuse a saved mapping unless `saved["printer_id"] == data.printer_id` (`print_queue.py:1073`). So the rule this PR generalises is already enforced by hand in the one place someone thought about it. Not arguing for a provenance column — validation covers strictly more, including the stale-spool case — just noting the precedent in case it is useful for how this gets described.

That reprint comment also says, of re-verifying a stored mapping against live AMS at dispatch: *"#1308 covers [it]; that check is a separate PR and, once merged, will also catch a stale slot inherited through this fallback."* This is that check, so #1308 looks like it closes with this too if you agree.

Two paths I confirmed are fine rather than merely untested: batch clones copy `printer_id` alongside `ams_mapping` (`print_batch.py:54-58`), so a clone stays on the printer its mapping was resolved for; and `_resolve_variant` (`:1183`) folds a model-scoped variant mapping on *before* `_ensure_ams_mapping` at `:1258`, so it is validated rather than trusted.

grolmus:
Pinned in `df17ded4` — `TestModelBasedVirtualPrinterMapping`, two tests.

The first is the failure: a queue item fresh from an "Any P2S" VP, `printer_id` still `None` and the slicer's `[2, -1, 1]` on the row, checked against a P2S whose tray 1 holds ASA. It recomputes to `[0, -1, 2]`, and asserts the judgement was made against the printer passed in rather than the row's absent one — that argument is the whole difference on this path, since the model-based caller has a printer before the item does. Fails on the previous commit with `2 != 0`; I verified that by stubbing the conflict check to `None` rather than assuming it.

The second keeps the other half honest: a VP mapping that does fit the chosen printer is left alone, so the slicer's deliberate slot pick is not thrown away by a check that only knows how to recompute.

26 tests in the file now, 6621 in `backend/tests/unit`, ruff clean.

maziggy (changes requested):
Second pass on 3cb2411b6. All seven items and the three nits from the last round are addressed, and 1 + 2 came back better than what I asked for: deriving one effectiveMappings per printer and hanging finalMapping, matchDetails and allPrintersReady off it makes badge/submission drift structurally impossible rather than fixing the two symptoms separately. getAutoMapping is gone rather than left dangling. The ownsDefaultMappings null-owner case you found on your own branch is the right call - treating an unknown owner as nobody is the only safe reading. Item 3 keyed off the computed mapping intersected with the requirement list inherits nozzle-awareness from the matcher for free, which was the point. Every fail-safe path has its own test.

One blocker, from the merge rather than from your work.

The branch does not currently run

print_scheduler.py:2585-2586 calls _canonical_filament_type, which no longer exists anywhere in the backend. dev extracted it to backend/app/utils/filament_types.canonical_filament_type, and the merge in 3cb2411b6 updated every pre-existing call site but not the two added in 3028b9012. The import is already present at line 59, so the fix is dropping the leading underscore on both lines.

ruff check backend/    F821 x2 (print_scheduler.py:2585, :2586)
pytest tests/          10 failed, 10184 passed

All ten failures are in test_ams_mapping_per_printer_2799.py, all the same NameError, and nothing else on the branch is affected. It raises on the type-check pass, which is reached whenever the mapped trays all exist on the target printer - the ordinary dispatch path - so it is not a corner case.

Two things while you are in there. canonical_filament_type accepts None and uppercases internally, so (req.get("type") or "").upper() can simply be req.get("type"). And please do not collapse want and have and want != have into the neighbouring filament_types_compatible: it returns True for two empties and False for one empty, so a requirement with no declared type would start reading as a conflict, which loses exactly the fail-safe this method is built around.

Worth fixing

The external-feed bail-out abandons the rest of the mapping. In the cheap pass an unrecognised >= 254 tray returns None for the whole method, so [254, -1, 7] with a foreign tray 7 is waved through - slot 3 is never examined. continue keeps the "absence of evidence" treatment for that slot alone while still judging the others, and the type pass is already safe there since by_tray.get(254) returns None and skips. test_unreported_external_feed_is_not_a_conflict uses [254, -1, 255], where both entries are external, so it cannot catch this.

The clear-on-recovery branch in _block_on_unmatched_filament is unreachable. On the printer-assigned path the interlock block at print_scheduler.py:1093 nulls any reason that is not its own before dispatch is reached; on the model-based path waiting_reason is set to None at assignment. Either way the startswith check sees None and never fires - the branch is exercised only by its own unit test. The more valuable edit is on the other side: that interlock block states "the interlock is the only thing that writes a waiting_reason on this branch", and this PR makes that false. Worth correcting so the next reader does not build on it.

Two 3MF parses per dispatch attempt in the happy path. _stored_mapping_conflict parses once the cheap pass succeeds, then _block_on_unmatched_filament parses again, each a DB lookup plus a zip open and XML parse with no cache. Correcting my own wording from last round: this is per dispatch attempt, not per pending item per tick - both calls sit behind the printer-idle gate, which I had wrong. It still means a one-file fan-out to 20 idle printers pays roughly 40 parses of the same file in one tick. A per-tick memo keyed on (archive_id or library_file_id, plate_id) would remove nearly all of it.

Nits

- test_foreign_tray_is_caught_without_reading_the_3mf repeats my incorrect "runs for every pending item on every tick" in its docstring. My error, but worth fixing in place rather than preserving.
- The description says 24 tests; the file has 26.
- useMultiPrinterFilamentMapping.test.tsx mutates statusByPrinter[10] in the last test without restoring it. Harmless in the current order, a trap for whoever adds a test after it.
- The hold's waiting_reason is English assembled in the backend, so the queue row is untranslated. Same as the existing model-based and interlock reasons, so not a regression from this PR - just noting the ceiling on that nit.

grolmus:
Blocker fixed, both "worth fixing" taken, all four nits done.

## Blocker

`canonical_filament_type` on both lines, and the coercion dropped as you suggested since it takes `None` and uppercases internally:

```python
want = canonical_filament_type(req.get("type"))
have = canonical_filament_type(loaded_tray.get("type"))
```

Left `want and have and want != have` alone. You are right about why: `filament_types_compatible` returns True for two empties and False for one, so a requirement with no declared type would start reading as a conflict — the one shape this method must never invent evidence for.

Worth saying plainly: the branch was broken and I pushed it that way. `ruff` and the suite would both have caught it, and I ran neither after your merge landed.

## Worth fixing

**External-feed bail-out.** `continue` rather than `return None`, so an unrecognised `>= 254` is absence of evidence about that slot only and the rest of the mapping is still judged. `test_an_external_feed_does_not_excuse_the_rest_of_the_mapping` uses `[254, -1, 7]` — the shape the old test could not reach, since both its entries were external.

**The clear-on-recovery branch.** Removed rather than kept, along with its two tests: the interlock block clears every reason written on this branch, so mine was dead code held up by tests written against it. The `_FILAMENT_HOLD_REASON_PREFIX` constant existed only so that branch could recognise its own message, so it is gone too and the string is inline.

Your comment at `:1119` is corrected, and I tried to make it say the thing that is actually load-bearing — that a held item never reaches that block, which is *why* its reason survives on screen while it waits, and that the hold re-evaluates on the same pass once it does:

```
# This is the single place a stale waiting_reason is cleared on this branch,
# for both the things that write one: the interlock itself, and the
# unmatched-filament hold at dispatch (#2799). A held item never reaches
# here — it is skipped on manual_start above, which is what keeps its reason
# on screen while it waits — so by the time one does arrive the hold has been
# lifted and its reason is stale. The hold re-evaluates on this same pass and
# writes a fresh one if the filament is still missing.
```

**Two 3MF parses.** Per-pass memo on `_get_filament_requirements`, keyed `(archive_id, library_file_id, plate_id)` and cleared at the top of `check_queue`. Both keys rather than `archive_id or library_file_id`, so a library file cannot collide with an archive of the same id.

The part worth reviewing is that it hands out a copy: `_apply_filament_overrides` rewrites those dicts in place, so a shared list would leak one item's overrides into the next item printing the same plate — a nastier bug than the one being fixed. Misses are cached too, so a missing file does not cost a DB lookup per gate per attempt. Five tests, including one that the memo does not outlive the pass.

Thanks for the correction on when these run — it changes the shape of the fix. Per dispatch attempt behind the idle gate is a much smaller number than per pending item per tick, but a fan-out to twenty idle printers is exactly the case this PR is about, so it was still worth doing.

## Nits

- Docstring no longer repeats the retracted "every pending item on every tick".
- Description says 30 now, and I will keep it in step rather than let it drift twice.
- `afterEach` restores `statusByPrinter`, with a note saying what it protects.
- Agreed on the untranslated reason, and agreed it is the existing ceiling rather than something to solve here.

## Verification

Ran what I should have run last time:

```
ruff check backend/       All checks passed!
ruff format --check       776 files already formatted
pytest backend/tests      10264 passed, 3 skipped
npm run test:run          3097 passed, 1 failed
tsc --noEmit / eslint     clean
```

The one frontend failure is `Layout.test.tsx` "MakerWorld nav link". I checked it properly this time rather than asserting it: run alone it fails 2/2 on the branch with my changes stashed and 3/3 with them applied, so it is pre-existing and not order-dependent. `SpoolFormModal.test.tsx` also appeared in one full run and not another; it passes in isolation either way, so that one is genuinely flaky under the parallel runner.

maziggy (changes requested):
Checks on the branch: ruff + format clean, 10466 backend tests pass (unit + integration, -n 30), frontend 3151 pass. Two notes on their test claims: the first of my three full frontend runs had 2 failures that didn't recur in runs 2 and 3 (load flake, same shape I saw on another branch today), and Layout.test.tsx "MakerWorld nav link" — which they've reported twice as pre-existing — passes here 26/26 in isolation and in all three full runs, so that one is local to their machine.

Draft comment for #2803:

Third pass. All four round-two items check out, and I traced the two that could have been fixed cosmetically rather than actually:

- canonical_filament_type on both sides at :2616-2617, imported at module level, and   `want and have and want != have` kept for the reason you gave.
- The external bail-out is `continue`, and [254, -1, 7] reaches the shape the old test   could not.
- The clear-on-recovery branch and its prefix constant are gone, and the comment at :1119 is now load-bearing rather than decorative: a held item is skipped on manual_start above, so its reason survives on screen, and the block writes None for everything else before the gates re-run. That is what makes your removal safe rather than merely tidy.
- The memo hands out dict(r) copies, and the thing that would have made shallow copies insufficient - _apply_filament_overrides touching something nested - is not the case: it only rewrites type, color and tray_info_idx. All four _get_filament_requirements call sites are inside check_queue, so nothing outside a pass can read a stale memo either.

Three things this round.

1. The complete-miss case falls out of the new hold and into the old ugly path. When a
stored mapping is rejected and the recompute resolves nothing, _ensure_ams_mapping clears the mapping and returns None on any printer that has an AMS -
_unmappable_without_ams_message only speaks when the printer positively reports `ams: []`. _block_on_unmatched_filament then bails on `not item.ams_mapping`, so the item dispatches with no mapping at all and takes the 0700_8012 route: several megabytes uploaded, firmware rejection, retries burned, "never started the print". That is precisely the situation your hold exists for, and this PR is what makes it reachable - before it, a conflicting stored mapping was trusted and never got here.

Your docstring reasons about it as the #2589 path, and for #2589 the docstring is right:
there the mapping is bogus because the AMS was not known yet, and holding would be holding on absence of evidence. But the two are separable with the evidence you already gather - hold only when live status reported loaded trays and the computed mapping still resolved nothing. #2589's cold status has an empty loaded list, so it stays untouched. Worth a test in the shape of the others: conflicting stored mapping, printer with trays but none matching, expect held with "Needs ..." rather than dispatched.

2. A held model-based item keeps the printer that held it. `item.printer_id = printer_id`
is committed at :1474, before the gates, so an "any P2S" job that hits the hold is pinned
to the one P2S that could not run it and the queue will not try another. That is the same
shape as _block_on_filament_deficit, so it is not a rule you invented - but this PR makes
it far more reachable, since any AMS change between assignment and dispatch now lands here. Worth a decision either way: release the assignment when holding on the model path, or say plainly that a held model job stays put until someone presses the start button.

3. The single-printer path still does not consult the ownership record.  getMappingForPrinter returns the shared amsMapping whenever one printer is selected, and that value folds in manualMappings without checking sharedMappingPrinterId. I could not build a live case: the effect at index.tsx:693 clears manualMappings whenever the selection differs from the initial one, which covers it. But that leaves the rule enforced explicitly on one path and resting on a clearing side-effect on the other, which is a thin thing to rely on the next time that effect is edited. Returning amsMapping only when sharedMappingPrinterId is null or matches effectivePrinterId makes both paths say the same thing.

Smaller:

- isinstance(tray, int) accepts JSON true, which reads as tray 1. Only reachable through a hand-written API payload, and one `not isinstance(tray, bool)` closes it.
- waiting_reason and the notification are English-only. Agreed it is the existing ceiling,   but it renders on the queue row, so it wants a line in the release notes rather than   silence.
- A held item needs a human press even once the right spool is back in. Same contract as the deficit gate, so consistent - just the sort of thing an unattended farm notices, and   it belongs in the same release-note line.
- Nothing covers _stored_mapping_conflict being handed a non-list from a corrupted row. The isinstance guard is right; it is just untested.

On verification: ruff and format clean here, 10466 backend tests pass, frontend 3151 pass. Your Layout.test.tsx "MakerWorld nav link" failure does not reproduce on this machine - 26/26 in isolation and clean in three full runs - so it is worth another look on yours rather than being carried forward as pre-existing. I did see two failures in one of my three full runs which did not recur, which matches the flakiness you noticed in
SpoolFormModal: the parallel runner here is not entirely stable under load, and that is
mine to chase, not yours.

PR #2823, opened 2026-08-12
Pick the nearest eligible filament colour instead of the first one in tray order (#2804)

### LINKED ISSUE #2804: [Bug]: Filament colour matching depends on AMS slot order and can pick a visibly wrong spool
Split out of #2799 as agreed — the `type_only_match` point, plus two neighbouring problems in the same matcher found while working on that fix. All three change which spool a job prints from when the exact colour is not loaded, so they belong together.

Same logic exists twice and both copies would need the change: `PrintScheduler._match_filaments_to_slots` / `_colors_are_similar` in `backend/app/services/print_scheduler.py`, and `computeMappingWithOverrides` / `colorsAreSimilar` in `frontend/src/hooks/useMultiPrinterFilamentMapping.ts` + `frontend/src/utils/amsHelpers.ts`.

## 1. The similar-colour bucket takes the first candidate, not the nearest

```python
elif self._colors_are_similar(f_color, req_color):
    if not similar_match:
        similar_match = f
```

Candidates are walked in tray order, so the winner is whichever *tray index* comes first among those inside the threshold — not the closest colour. With two spools of the required type both within tolerance, which one prints is decided by AMS slot order.

Concretely: requirement `#3A7BD5`, AMS holds `#2850E0` in tray 1 and `#3B7AD2` in tray 3. Both pass, tray 1 wins, and the visibly-correct spool in tray 3 is ignored. Reordering the spools changes the output with no other change.

Ranking the bucket by distance and taking the minimum would be a small change and makes the result independent of slot order — which is the property #2799 was about in the first place.

## 2. `_colors_are_similar` is a per-channel box, not a distance

```python
return abs(r1 - r2) <= threshold and abs(g1 - g2) <= threshold and abs(b1 - b2) <= threshold
```

With `threshold=40` this accepts a corner-to-corner difference of √(3·40²) ≈ 69 while rejecting a single-channel difference of 41. So `#000000` vs `#282828` (visibly different greys) is "similar", and `#0000FF` vs `#0029FF` (nearly identical blues) is not.

If the bucket becomes nearest-wins, the natural metric is a proper distance — Euclidean in RGB is the cheap option, CIE76 in Lab the accurate one if a conversion helper is acceptable. Keeping the box as the *admission* test and adding distance only for *ranking* would already fix the ordering dependence without changing which spools are eligible, if you would rather not touch eligibility at all.

## 3. `type_only_match` silently prints the wrong colour

The final fallback matches on type with any colour, so a requirement whose colour is not loaded at all prints in whatever same-type spool is free, with no signal. From #2799: making it opt-in changes behaviour for every existing user, so the suggestion there was to surface it on the queue item instead. That still seems right — the job goes out, but the queue row says it is running in a colour the file did not ask for.

The same treatment would suit case 1: record which bucket won (`idx` / `exact_color` / `similar_color` / `type_only`), so "why did it pick that spool" is answerable after the fact rather than only under the `prefer_lowest` logging that exists today.

## Practical note

Transparency is a non-issue and I want to record it since it looks like one: `_normalize_color_for_compare` truncates to 6 hex chars, so the alpha the slicer writes for a transparent filament is dropped on both sides and a transparent `#76D9F4` matches an AMS `76D9F4FF` exactly. Worth keeping that truncation in mind if the comparison is ever reworked — an alpha-aware distance would regress it.

Happy to take this one too if the direction is agreed. My preference would be minimum-distance ranking within the existing eligibility rules (1 + the ranking half of 2), leaving 3 as a separate visibility change, but I would rather follow your call on scope than guess.

grolmus:
## Description

When no loaded spool matches a required colour exactly, the matcher falls back to spools inside a per-channel tolerance and took the first of those in tray order. The winner therefore depended on which slot a spool happened to sit in, and moving spools between slots changed the outcome with nothing else changed.

Eligible spools are now ranked by RGB distance and the nearest wins. Eligibility is untouched — the same per-channel box as before — so no spool becomes usable or unusable, only better or worse ranked among those already eligible.

Scope is the one agreed in #2804: minimum-distance ranking within the existing eligibility rules, plus the unconditional bucket log. The `type_only` visibility change is left for its own issue.

## Related Issue

Fixes #2804

## Documentation

**Pick one**:
- [x] No docs update required — reason: no new setting or UI surface. The behaviour change is described in the CHANGELOG entry, which spells out that it changes which spool some prints use and how it interacts with Prefer Lowest.

## Type of Change

- [x] Test addition or update

## Changes Made

**One ranking rule, shared.** `findNearestSimilar(candidates, requiredColor, getColor)` and `colorDistance()` land in `frontend/src/utils/amsHelpers.ts`, and all three frontend matchers call it: `autoMatchFilament` (same file), `computeAmsMapping` in `useFilamentMapping.ts` — including its own `tray_info_idx` branch — and both `computeMappingWithOverrides` and `computeMatchDetails` in `useMultiPrinterFilamentMapping.ts`. The backend gets the same rule as `PrintScheduler._color_distance`, applied in both colour-comparison branches of `_match_filaments_to_slots`. That is the four copies you listed, now agreeing by construction rather than by four people remembering to edit four places.

**Prefer Lowest becomes the tie-break, as you called it.** `findNearestSimilar` keeps the earliest candidate on an equal distance (strict `<`), and callers already pass their candidates in the order they established — slot order, or the prefer-lowest sort. So colour distance decides first, and the existing sort key decides between spools that are equally close, with no separate plumbing. Tests pin both halves: nearest wins regardless of order, and a tie follows the incoming order.

**The bucket log is unconditional** (`[ams-match] picked gtid=… via …`), and carries the colour distance when the winner was a near-match. It was only emitted under `prefer_lowest` before. A `similar_color` win is now a ranked choice rather than a positional accident, so being able to see which rule won — and by how much — is what makes the change diagnosable after the fact. Renamed from `[prefer-lowest]` since it is no longer scoped to that preference; nothing asserts on the old prefix.

**`slicePresetPicker.ts`** is untouched, as you asked. It uses `colorsAreSimilar` as a `+2` term in a score it already maximises, not to pick a spool, and `colorsAreSimilar` itself is unchanged — so its behaviour is identical.

**Alpha stays dropped** on both sides. Noted in both docstrings so a future rework does not "fix" it: the alpha a slicer writes for a transparent filament is not a colour the user chose, and counting it would stop a transparent filament matching itself.

**CHANGELOG** entry under Changed, stating plainly that this changes which spool some prints use and what happens to Prefer Lowest.

## Screenshots

No visual change. The per-printer match badge can now report a different spool than before, but only because the underlying pick changed — no component or layout was touched.

## Testing

- [x] I have tested with my printer model: P2S (×3), X1C (×3), H2D

Your example is the fixture on both sides — required `#3A7BD5`, purple `#6253AD` in tray 1 (40/40/40 off, admitted, distance ≈69) and `#3B7AD2` in tray 3 (distance ≈3). Thanks for correcting the one in the issue; `#2850E0` is 43 off on green and never was a candidate.

`backend/tests/unit/test_nearest_colour_match_2804.py` — 12 tests: distance is euclidean not per-channel, alpha ignored, unusable input returns `None` rather than a number, nearest wins, **the answer does not change when the tray order is reversed** (the bug in one line), exact still outranks near, eligibility unchanged so 41-off stays type-only, ties follow caller order, nearest applies inside a shared `tray_info_idx`, type remains a hard filter, and two slots wanting the same colour consume different spools nearest-first.

`frontend/src/__tests__/utils/nearestColourMatch.test.ts` — 10 tests over the shared helper, mirroring the above plus "skips a candidate with no usable colour instead of throwing".

```
backend/tests/unit                     6779 passed, 3 skipped
test_nearest_colour_match_2804.py        12 passed
ruff check / format                    clean
tsc --noEmit / eslint                  clean
frontend npm run test:run              2992 passed, 2 failed
```

Neither frontend failure is from this branch:

- `Layout.test.tsx` "MakerWorld nav link" — fails identically on a clean `dev` with this commit stashed.
- `SpoolFormModal.test.tsx` — passes in isolation both with and without this commit, and did not recur on a second full run. Flaky under the parallel runner rather than a regression.

## Checklist


maziggy (approved):

maziggy:
Very nice! I think that'my very first PR which gets merged after the first review. Keep it going :)

Issue #1148, opened 2026-04-27
[Feature]: Display HA Binary Sensors


bsaunder:
### Problem or Use Case

My printer is inside an enclosure with an exhaust fan to ensure all of the fumes are extracted away outside. I have Contact Sensors on the doors for the enclosure that are linked to Home Assistant. It would be nice to be able to see their state on the Printer Card to know if I left the enclosure open before I start a print remotely. 

### Proposed Solution

Add the ability to include binary_sensor types from Home Assistant and then display them on the Printer Card inline or under the existing HA switches. Configuration of the binary_sensor could likely use the same modal as configuration of the HA switches. 

Bonus points if we can get notifications based on the state, but that is obviously a lot more work. 

### Alternatives Considered

There is no way to read in sensor data and display it, so therefore there is no workaround in Bambuddy. Current workaround is simply to utilize the Bambu Integration in HA and use an Automation there to send an alert. 

### Feature Category

Smart Plugs

### Priority

Nice to have

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Before we commit to building this, I'd like to gauge community interest. If you'd find this feature useful, please give this issue a thumbs up (👍) reaction so we can prioritize accordingly.

maziggy:
Available/Fixed in branch dev and available with the next release or daily build.

May I ask you guys to test and provide feedback? Thx.

https://wiki.bambuddy.cool/features/smart-plugs/?h=home+assistant+sensors#home-assistant-sensors

If you find Bambuddy useful, please consider giving it a ⭐ on [GitHub](https://github.com/maziggy/bambuddy) — it helps others discover the project!

malkinskir:
I've tested it and works as expected

<img width="684" height="418" alt="Image" src="https://github.com/user-attachments/assets/d8f5f402-6f2c-4436-8c49-13ef503fe8cd" />

<img width="2138" height="1160" alt="Image" src="https://github.com/user-attachments/assets/b92c985c-f32c-4f51-8cd1-b3a3b7b60f5e" />

PR #2827, opened 2026-08-13
[Feature]: Bind Home Assistant sensors to storage locations (dryboxes/bins)

### LINKED ISSUE #2824: [Feature]: Bind Home Assistant sensors to storage locations (dryboxes/bins)
### Problem or Use Case

Bambuddy already lets a printer's card show a bound Home Assistant sensor (a door contact, a thermometer). A lot of us also keep filament in Home-Assistant-monitored dryboxes or storage bins — temperature, humidity, sometimes a battery-powered sensor — but there's currently no way to bind those sensors to a Bambuddy storage location. Keeping an eye on drybox conditions means tabbing over to Home Assistant instead of staying inside Bambuddy.

### Proposed Solution

Extend the existing printer-sensor binding feature to storage locations:

Bind a Home Assistant temperature/humidity/battery sensor to a storage location, the same way a sensor can already be bound to a printer.
Show the live reading on the filament card and in the inventory table for any spool stored at that location.
Configurable alert thresholds per sensor category (above/below), with optional notifications and value colorization against those thresholds.
When the first sensor is bound for a location, offer to auto-bind matching temperature/humidity/battery siblings that share the same entity-ID prefix, each individually selectable via checkbox

I have build it for my own, but I think its a good improvement.

### Alternatives Considered

Users can keep watching Home Assistant separately, or note drybox readings by hand, but that works against the point of having one dashboard for prints and filament.

### Feature Category

Spool Inventory

### Priority

Would improve my workflow

### Mockups or Examples

Image

Image

Image

Image

Image

Image

Image

### Contribution

### Checklist

MagicMelody84:
## Description

Adds Home Assistant sensor bindings for storage locations (dryboxes, filament
bins), mirroring the existing printer-sensor binding feature. Bound
temperature/humidity/battery sensors show live readings on the filament card
and in the inventory table, with configurable alert thresholds, notifications,
and threshold colorization.

## Related Issue

Fixes #2824

## Documentation

- Wiki: maziggy/bambuddy-wiki#68

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- **Backend**: new `location_ha_sensors` table/model/schema mirroring
  `printer_ha_sensors` (minus print-blocking, which doesn't apply to a storage
  bin), a `LocationHASensorManager` background poller with its own cache, a
  `location_ha_sensor_alert` notification template, and a
  `/location-ha-sensors` REST API (CRUD, entity picker, per-location
  readings). The poll/refresh cadence is a new user setting
  (`location_sensor_poll_interval`, 60s minimum, 120s default) instead of a
  fixed interval, and the poller runs once immediately on startup so a
  restart doesn't leave the UI blank until the first cycle completes.
- **Settings**: a new "Sensors" tab groups the existing printer-sensor
  section with the new "Home Assistant Sensors (Storage Locations)" section
  (with a connected-sensor count badge, matching Smart Plugs). Binding the first sensor
 for a location offers to auto-bind matching temperature/humidity/battery siblings that share
 the same entity-ID prefix, each individually selectable via checkbox (checked by default).
A "Location Sensor Options" dialog sets default alert thresholds per
  category, whether values are colorized against those thresholds
  (configurable above/below/optimal colors), and can bulk-apply the current
  defaults to every existing sensor.
- **Inventory**: the filament card shows a footer with the spool's storage
  location and any bound sensor readings when available; the table view
  gains sortable Temperature/Humidity/Battery columns with the same
  threshold colorization, formatted to two decimal places.
- **Locations modal**: shows how many sensors are bound to each location, and
  a location can be created inline while adding a sensor.

## Testing

- [ ] I have tested with my printer model: — n/a, this change doesn't touch printer communication
- [x] `./test_frontend.sh` — TypeScript, ESLint, Vitest, and i18n parity all pass
- [x] `./test_backend.sh --full` — Ruff + full pytest suite passes
- [x] `./test_docker.sh` — full Docker build, unit tests, and integration tests pass

## Checklist

- [] I have commented my code where necessary

## Additional Notes

Sensor-category matching (temperature/humidity/battery) uses the Home Assistant
 `entity_id` suffix convention (`..._temperature`, `..._humidity`, `..._battery`) to find
 and offer related sensors automatically when the first one is bound to a location — 
a naming-convention match, not a device-registry lookup, since Home Assistant's state 
API doesn't expose which physical device an entity belongs to.

Only one sensor per category is allowed per location — binding a second
sensor of the same category prompts to replace the existing one.


maziggy:
I don't have a problem with using AI for assistance but to be honest, I want to avoid Claude Code commits in the repo. Please remove. Thanks.

maziggy (changes requested):
Thanks for this, and for mirroring the printer sensor feature so closely - the alert-edge semantics, the fail-quiet behaviour when HA is unreachable, and the cascade delete are all handled right. I ran the checks here too: ruff clean, 10195 backend tests pass, your 28 new frontend tests pass, i18n parity holds across all 13 locales, eslint and build clean.

A few things I would like addressed before this goes in.

1. Sensor names are the raw entity ID. LocationHASensorModal.tsx:170 sets name: entityId.slice(0, 100) and there is no name field in the modal. The printer modal has an editable name prefilled from entity.friendly_name (HASensorModal.tsx:55,113). As it stands, the settings card, the filament card and the notification body all read "sensor.drybox_1_temperature is 65 %". The picker already has friendly_name available.

2. Per-location polling runs unconditionally. InventoryPage.tsx:1157 fires one request per distinct storage location, and SpoolLocationFooter (:2776) fires a second one per location in card view - so 2 queries per location, repeating at the poll interval. Nothing gates them on whether any sensor exists or whether Home Assistant is configured, and the new table columns default to hidden. An install with 20 bins and no HA does 40 round trips per page load and 40 more every 120s to receive empty arrays. SettingsPage.tsx:533 is the same shape, polling the sensor list on every tab just to keep a badge count fresh. Since the locationHaSensors list is already fetched, gating on it being non-empty (or adding one bulk readings endpoint) collapses this to a single request.

3. Options -> Save has no error path. LocationSensorOptionsModal.tsx:173 awaits api.updateSettings with no try/catch. PUT /settings/ needs SETTINGS_UPDATE (admin) but the Options button is not admin-gated, so a non-admin pressing Save gets an unhandled rejection: no toast, no error, modal stays open, and the localStorage half of persistDefaults has already written. The Reset path right next to it does this correctly with onError. It would also help to only PATCH the interval when it actually changed, so non-admins can still save the colour preferences.

4. The permission choice diverges from the sibling feature. ha_sensors.py reuses SMART_PLUGS_* with a comment explaining why; these routes use INVENTORY_* with no note. That has an API-key consequence: SMART_PLUGS_CREATE/UPDATE/DELETE are on the denylist as administrative, while INVENTORY_CREATE/UPDATE/DELETE are allowlisted under can_manage_inventory. So an inventory-scoped API key can create, edit and delete Home Assistant bindings, which the printer feature deliberately refuses. Please either use SMART_PLUGS_* for consistency or add a comment stating the decision.

5. The docstrings and comments did not come across with the code. Diffing the new files against their printer originals, the code is near-identical and every explanatory comment is gone: why /entities must precede /{sensor_id}, why the post-save refresh is best-effort, why PATCH re-validates against a merged row, why _last_alerting is kept separate from _readings. Those carry the reasoning for decisions that are not obvious from the code, and I would like them in the new files too.

Smaller things, happy to take these or not:

- The Settings cards render sensor.last_state from the database rather than the readings endpoint, so they have no notion of reachable. An unreachable sensor keeps showing its last value, colorized as optimal, indefinitely. The inventory card handles this correctly with haSensors.unavailable.
- SettingsPage.tsx:544 deletes sequentially and onError does not invalidate, so a failure partway leaves already-deleted sensors on screen. onSettled would cover it.
- BINARY_LABELS and ICONS are now copied verbatim from PrinterHASensorRow.tsx into both InventoryPage.tsx and SettingsPage.tsx, and the threshold logic exists in locationSensorDefaults.ts, SettingsPage.tsx and the backend. Adding a device class is a three-file edit that will drift - worth extracting.
- evaluate/describe_state/SensorReading are reused for LocationHASensor but still annotated PrinterHASensor.
- test_manager_has_no_blocked_locations_concept asserts not hasattr on methods nobody wrote, so it cannot fail. The @pytest.mark.asyncio decorators are redundant (asyncio_mode = auto), and the file could use the _2824 suffix to match the convention.
- Six display preferences sit in localStorage while the poll interval is a server setting, in the same dialog. Worth a thought about whether the colours should follow the interval.
- toFixed(2) gives "87.00 %" for battery and humidity.
- The German block mixes Sie ("Binden Sie ... Ihre Lagerorte") and du ("Verknuepfe", "Waehle aus") within the same section.
- One-sensor-per-category is enforced only in the UI while the API accepts duplicates, and the table's .find() then picks an arbitrary one. Documented in sectionDescription, so mostly a note.
- findSiblingEntities matches on entity-ID naming rather than the HA device. The description says "found on the same Home Assistant device", which is more than /api/states can actually support - worth rewording in the PR and the wiki page.

MagicMelody84:
> Thanks for this, and for mirroring the printer sensor feature so closely - the alert-edge semantics, the fail-quiet behaviour when HA is unreachable, and the cascade delete are all handled right. I ran the checks here too: ruff clean, 10195 backend tests pass, your 28 new frontend tests pass, i18n parity holds across all 13 locales, eslint and build clean.
> 
> A few things I would like addressed before this goes in.
> 
> 1. Sensor names are the raw entity ID. LocationHASensorModal.tsx:170 sets name: entityId.slice(0, 100) and there is no name field in the modal. The printer modal has an editable name prefilled from entity.friendly_name (HASensorModal.tsx:55,113). As it stands, the settings card, the filament card and the notification body all read "sensor.drybox_1_temperature is 65 %". The picker already has friendly_name available.
> 2. Per-location polling runs unconditionally. InventoryPage.tsx:1157 fires one request per distinct storage location, and SpoolLocationFooter (:2776) fires a second one per location in card view - so 2 queries per location, repeating at the poll interval. Nothing gates them on whether any sensor exists or whether Home Assistant is configured, and the new table columns default to hidden. An install with 20 bins and no HA does 40 round trips per page load and 40 more every 120s to receive empty arrays. SettingsPage.tsx:533 is the same shape, polling the sensor list on every tab just to keep a badge count fresh. Since the locationHaSensors list is already fetched, gating on it being non-empty (or adding one bulk readings endpoint) collapses this to a single request.
> 3. Options -> Save has no error path. LocationSensorOptionsModal.tsx:173 awaits api.updateSettings with no try/catch. PUT /settings/ needs SETTINGS_UPDATE (admin) but the Options button is not admin-gated, so a non-admin pressing Save gets an unhandled rejection: no toast, no error, modal stays open, and the localStorage half of persistDefaults has already written. The Reset path right next to it does this correctly with onError. It would also help to only PATCH the interval when it actually changed, so non-admins can still save the colour preferences.
> 4. The permission choice diverges from the sibling feature. ha_sensors.py reuses SMART_PLUGS_* with a comment explaining why; these routes use INVENTORY_* with no note. That has an API-key consequence: SMART_PLUGS_CREATE/UPDATE/DELETE are on the denylist as administrative, while INVENTORY_CREATE/UPDATE/DELETE are allowlisted under can_manage_inventory. So an inventory-scoped API key can create, edit and delete Home Assistant bindings, which the printer feature deliberately refuses. Please either use SMART_PLUGS_* for consistency or add a comment stating the decision.
> 5. The docstrings and comments did not come across with the code. Diffing the new files against their printer originals, the code is near-identical and every explanatory comment is gone: why /entities must precede /{sensor_id}, why the post-save refresh is best-effort, why PATCH re-validates against a merged row, why _last_alerting is kept separate from _readings. Those carry the reasoning for decisions that are not obvious from the code, and I would like them in the new files too.
> 
> Smaller things, happy to take these or not:
> 
> * The Settings cards render sensor.last_state from the database rather than the readings endpoint, so they have no notion of reachable. An unreachable sensor keeps showing its last value, colorized as optimal, indefinitely. The inventory card handles this correctly with haSensors.unavailable.
> * SettingsPage.tsx:544 deletes sequentially and onError does not invalidate, so a failure partway leaves already-deleted sensors on screen. onSettled would cover it.
> * BINARY_LABELS and ICONS are now copied verbatim from PrinterHASensorRow.tsx into both InventoryPage.tsx and SettingsPage.tsx, and the threshold logic exists in locationSensorDefaults.ts, SettingsPage.tsx and the backend. Adding a device class is a three-file edit that will drift - worth extracting.
> * evaluate/describe_state/SensorReading are reused for LocationHASensor but still annotated PrinterHASensor.
> * test_manager_has_no_blocked_locations_concept asserts not hasattr on methods nobody wrote, so it cannot fail. The @pytest.mark.asyncio decorators are redundant (asyncio_mode = auto), and the file could use the _2824 suffix to match the convention.
> * Six display preferences sit in localStorage while the poll interval is a server setting, in the same dialog. Worth a thought about whether the colours should follow the interval.
> * toFixed(2) gives "87.00 %" for battery and humidity.
> * The German block mixes Sie ("Binden Sie ... Ihre Lagerorte") and du ("Verknuepfe", "Waehle aus") within the same section.
> * One-sensor-per-category is enforced only in the UI while the API accepts duplicates, and the table's .find() then picks an arbitrary one. Documented in sectionDescription, so mostly a note.
> * findSiblingEntities matches on entity-ID naming rather than the HA device. The description says "found on the same Home Assistant device", which is more than /api/states can actually support - worth rewording in the PR and the wiki page.

Thanks for the thorough pass. Went through everything:

**1. Sensor names** — Added the name field, editable, prefilled from `entity.friendly_name` (same pattern as `HASensorModal.tsx:55,113`). It also now follows the entity picker when you switch entities, as long as you haven't typed a custom name yourself. Auto-added siblings default to their own `friendly_name` too instead of their entity id. One deliberate exception: the Settings sensor overview list still shows the entity id, not the name — that's where you'd actually want to confirm which Home Assistant entity a row is bound to. The friendly name shows up on hover there, and everywhere a reading is displayed (filament card, inventory table, notifications) it's the name, not the id.

**2. Unconditional polling** — Fetch the sensor list once (ungated, invalidated on create/edit/delete rather than polled) and use it to skip the readings query entirely for locations with nothing bound. `SpoolLocationFooter` reads the same query key, so it's a cache hit rather than a second request. Also dropped the `refetchInterval` from the Settings badge-count query — it only backs a count that's already kept fresh by the same invalidation, it never needed to poll.

**3. Save error path** — Moved onto a mutation with `onError`/`onSuccess`, matching the Reset button's existing pattern: error toast + modal stays open on failure, success toast + close otherwise. The poll interval is now only PATCHed when it actually changed, so a non-admin can still save colour/threshold prefs without ever touching `SETTINGS_UPDATE`.

**4. Permission choice** — Switched to `SMART_PLUGS_*`, with a comment spelling out why (and the API-key consequence you flagged, spelled out explicitly so it doesn't get "fixed" back the other way later).

**5. Missing comments** — Ported the reasoning back into all four files (model, schema, manager, routes), adapted for locations: route ordering, best-effort refresh, merged-row re-validation, the `_last_alerting`/`_readings` split, and the immediate-poll-on-start behavior that doesn't have a printer-side equivalent to copy from.

Smaller items:

- Settings cards now read the live readings endpoint instead of `last_state`, so an unreachable sensor shows "Unavailable" instead of a stale colorized value forever.
- Bulk delete invalidates in `onSettled`, so a partial failure doesn't leave already-deleted rows on screen.
- Extracted `BINARY_LABELS`/`ICONS`/icon-selection into `utils/haSensorDisplay.ts`, used by both the printer and location code now — one place to add a device class instead of three-to-five.
- `evaluate()`/`describe_state()` take a structural `Protocol` instead of the concrete `PrinterHASensor` type.
- Dropped the no-op `test_manager_has_no_blocked_locations_concept` test, dropped the redundant `@pytest.mark.asyncio` decorators (`asyncio_mode = auto` already covers it), renamed the file to the `_2824` suffix.
- Fixed the German Sie/du mismatch in `sectionDescription`.
- Reworded the wiki's "found on the same Home Assistant device" claim — it's a naming-convention match on the entity-ID prefix, not a device-registry lookup, since `/api/states` doesn't expose that.

Left for later, deliberately:

- `toFixed(2)` — tried the trailing-zero trim, then reverted it. Fixed two decimals is actually a deliberate choice: it keeps values roughly aligned width-wise (temperature, humidity, battery all read at a consistent decimal width instead of jumping around). Not attached to it though — happy to trim trailing zeros again if you'd still rather have that.
- Enforcing one-sensor-per-category server-side (currently UI-only, as you noted, and documented as such).

maziggy:

MagicMelody84:
sorry was late yesterday, done

maziggy (changes requested):
Second-round review done. Everything ran clean here: ruff + ruff format --check clean, 10462 backend tests pass (-n 30, unit + integration), eslint and npm run build clean, 3184 frontend tests pass. One caveat: the first full vitest run failed one test in ConfigureAmsSlotModal.test.tsx (untouched by this PR); it passed on two re-runs and in isolation, so it looks like a pre-existing flake under parallel load, not a PR regression.

Of the round-1 items, 1, 4 and 5 and most of the smaller ones are genuinely fixed; 2 and 3 are half-fixed — the contributor's "cache hit" claim about the footer holds for the sensor list but not for the readings, which still use a different query key. Five new things came in with the fixes.

Draft comment for #2827:

Second pass. Checks first: ruff and ruff format clean, 10462 backend tests pass, eslint
and build clean, 3184 frontend tests pass.

Confirmed fixed: 1 (names, and following the picker is a nice touch), 4 (permissions,
with the comment), 5 (comments ported, the Protocol, the shared haSensorDisplay.ts), plus last_state -> readings endpoint, onSettled on the bulk delete, the no-op test dropped and the file renamed, and the German section no longer mixing registers.

Still open from last time:

A. Item 2 is half done. The readings query is now gated on a location actually having a
sensor, which was the main thing. But SpoolLocationFooter does not share the fetch - it
shares the sensor list, not the readings. InventoryPage.tsx:1186 uses the key ending
'all' (show_on_card=false), the footer at :2751 uses 'cardOnly' (show_on_card=true), so
in card view every location with sensors still costs two requests per interval. And in
table view the 'all' query runs even when temperature, humidity and battery are all
hidden, which is their default. Gating it on view mode plus at least one of those columns
being visible, and letting the footer filter the 'all' payload (needs show_on_card in the
reading schema), collapses this to one.

B. Item 3's error path is right now, but the write order is not. persistDefaults
(LocationSensorOptionsModal.tsx:163) writes all six localStorage preferences first and
PATCHes after, so a non-admin still gets the error toast after the colours have already
been saved. Server call first, localStorage after, and the toast tells the truth.

New:

C. Location alerts ride on the printer sensor toggle. notification_service.py:1880 looks
providers up under on_ha_sensor_alert with printer_id None. Every other event in that file has its own column - this is the only reuse. So someone who enabled Sensor Alert on a provider scoped to a single printer now also receives drybox alerts, with no way to have one without the other. Its own column and toggle would be the consistent thing; if you would rather not add a column, please say so in a comment and reword the toggle.

D. The reset confirm closes the Options dialog. The ConfirmModal at
LocationSensorOptionsModal.tsx:380 is rendered inside the overlay div whose onClick is onClose (:235), and ConfirmModal's own overlay does not stop propagation - so dismissing the confirm by clicking beside it also closes Options and drops unsaved edits. LocationHASensorModal gets this right by rendering its confirms as siblings in a fragment.

E. The overwrite path can lose a binding. overwriteMutation (LocationHASensorModal.tsx:250) deletes the existing sensor and only then calls saveMutation; if that create fails, the old binding is gone with nothing in its place. Nothing guards close-while-pending either - the Escape handler at :86 and the backdrop at :316 close unconditionally, where LocationsModal does guard it - so an Escape landing between the delete and the create does the same. PATCHing the existing row onto the new entity avoids both.

F. Display preferences are read once, per component. InventoryPage.tsx:661 and every
SpoolLocationFooter read localStorage in useState initializers, so changing the colours in Settings does nothing to an already-open Inventory page until a reload - while SettingsPage re-reads them when the modal closes. Same preference, two behaviours. A small hook, or passing them down from the page, fixes that and drops four localStorage reads per card.

Smaller:

- haSensorDisplay.ts:53 adds battery to the icon map, which also changes the printer row (battery used to fall through to Gauge). Fine by me, just noting it touches #1148.
- The two features format differently now: the printer row prints 23.4, the location   surfaces toFixed(2), so 23.40 and 87.00 %. Your call on trailing zeros, but with the   display helpers shared, one formatter is the natural place for it.
- The auto-add prompt's onCancel (LocationHASensorModal.tsx:594) still saves the primary sensor, which is the right behaviour, but the button says Cancel. A cancelText along the lines of "Only this one" would match what it does.
- LocationsModal with startCreating and no onPickLocation renders null after a successful save while open stays true. One caller today, so a guard or a note on the prop is enough.
- on_location_ha_sensor_alert takes location_id and never uses it, and it is the only   method in that file without a docstring.
- de.ts runs 363 Sie against 25 du, so the section is now consistent with itself but not   with the file.
- No regression test for the polling gate, which was the main fix this round.

One note rather than a request: moving the printer sensor section out of Plugs into a new Sensors tab is the right call - sensors are things you read, plugs are things you switch, and I would rather they stopped sharing a tab. So the section goes from the Plugs tab entirely, which this PR already does.

Two things follow from that, both mine to handle, listed here so nothing gets done twice:
the release notes need a line saying the printer sensors moved, and the wiki documents the feature inside features/smart-plugs.md ("Go to Settings > Smart Plugs and scroll to Home Assistant Sensors"), which stops being true on merge. That section wants to become its own page covering both the printer bindings and your storage-location ones. Nothing for you to change in the PR for either.

MagicMelody84:
> Second-round review done. Everything ran clean here: ruff + ruff format --check clean, 10462 backend tests pass (-n 30, unit + integration), eslint and npm run build clean, 3184 frontend tests pass. One caveat: the first full vitest run failed one test in ConfigureAmsSlotModal.test.tsx (untouched by this PR); it passed on two re-runs and in isolation, so it looks like a pre-existing flake under parallel load, not a PR regression.
> 
> Of the round-1 items, 1, 4 and 5 and most of the smaller ones are genuinely fixed; 2 and 3 are half-fixed — the contributor's "cache hit" claim about the footer holds for the sensor list but not for the readings, which still use a different query key. Five new things came in with the fixes.
> 
> Draft comment for #2827:
> 
> Second pass. Checks first: ruff and ruff format clean, 10462 backend tests pass, eslint and build clean, 3184 frontend tests pass.
> 
> Confirmed fixed: 1 (names, and following the picker is a nice touch), 4 (permissions, with the comment), 5 (comments ported, the Protocol, the shared haSensorDisplay.ts), plus last_state -> readings endpoint, onSettled on the bulk delete, the no-op test dropped and the file renamed, and the German section no longer mixing registers.
> 
> Still open from last time:
> 
> A. Item 2 is half done. The readings query is now gated on a location actually having a sensor, which was the main thing. But SpoolLocationFooter does not share the fetch - it shares the sensor list, not the readings. InventoryPage.tsx:1186 uses the key ending 'all' (show_on_card=false), the footer at :2751 uses 'cardOnly' (show_on_card=true), so in card view every location with sensors still costs two requests per interval. And in table view the 'all' query runs even when temperature, humidity and battery are all hidden, which is their default. Gating it on view mode plus at least one of those columns being visible, and letting the footer filter the 'all' payload (needs show_on_card in the reading schema), collapses this to one.
> 
> B. Item 3's error path is right now, but the write order is not. persistDefaults (LocationSensorOptionsModal.tsx:163) writes all six localStorage preferences first and PATCHes after, so a non-admin still gets the error toast after the colours have already been saved. Server call first, localStorage after, and the toast tells the truth.
> 
> New:
> 
> C. Location alerts ride on the printer sensor toggle. notification_service.py:1880 looks providers up under on_ha_sensor_alert with printer_id None. Every other event in that file has its own column - this is the only reuse. So someone who enabled Sensor Alert on a provider scoped to a single printer now also receives drybox alerts, with no way to have one without the other. Its own column and toggle would be the consistent thing; if you would rather not add a column, please say so in a comment and reword the toggle.
> 
> D. The reset confirm closes the Options dialog. The ConfirmModal at LocationSensorOptionsModal.tsx:380 is rendered inside the overlay div whose onClick is onClose (:235), and ConfirmModal's own overlay does not stop propagation - so dismissing the confirm by clicking beside it also closes Options and drops unsaved edits. LocationHASensorModal gets this right by rendering its confirms as siblings in a fragment.
> 
> E. The overwrite path can lose a binding. overwriteMutation (LocationHASensorModal.tsx:250) deletes the existing sensor and only then calls saveMutation; if that create fails, the old binding is gone with nothing in its place. Nothing guards close-while-pending either - the Escape handler at :86 and the backdrop at :316 close unconditionally, where LocationsModal does guard it - so an Escape landing between the delete and the create does the same. PATCHing the existing row onto the new entity avoids both.
> 
> F. Display preferences are read once, per component. InventoryPage.tsx:661 and every SpoolLocationFooter read localStorage in useState initializers, so changing the colours in Settings does nothing to an already-open Inventory page until a reload - while SettingsPage re-reads them when the modal closes. Same preference, two behaviours. A small hook, or passing them down from the page, fixes that and drops four localStorage reads per card.
> 
> Smaller:
> 
> * haSensorDisplay.ts:53 adds battery to the icon map, which also changes the printer row (battery used to fall through to Gauge). Fine by me, just noting it touches #1148.
> * The two features format differently now: the printer row prints 23.4, the location   surfaces toFixed(2), so 23.40 and 87.00 %. Your call on trailing zeros, but with the   display helpers shared, one formatter is the natural place for it.
> * The auto-add prompt's onCancel (LocationHASensorModal.tsx:594) still saves the primary sensor, which is the right behaviour, but the button says Cancel. A cancelText along the lines of "Only this one" would match what it does.
> * LocationsModal with startCreating and no onPickLocation renders null after a successful save while open stays true. One caller today, so a guard or a note on the prop is enough.
> * on_location_ha_sensor_alert takes location_id and never uses it, and it is the only   method in that file without a docstring.
> * de.ts runs 363 Sie against 25 du, so the section is now consistent with itself but not   with the file.
> * No regression test for the polling gate, which was the main fix this round.
> 
> One note rather than a request: moving the printer sensor section out of Plugs into a new Sensors tab is the right call - sensors are things you read, plugs are things you switch, and I would rather they stopped sharing a tab. So the section goes from the Plugs tab entirely, which this PR already does.
> 
> Two things follow from that, both mine to handle, listed here so nothing gets done twice: the release notes need a line saying the printer sensors moved, and the wiki documents the feature inside features/smart-plugs.md ("Go to Settings > Smart Plugs and scroll to Home Assistant Sensors"), which stops being true on merge. That section wants to become its own page covering both the printer bindings and your storage-location ones. Nothing for you to change in the PR for either.

Thanks again for the second pass — went through all of it:

**1. Two requests for one location** — Table view and the card footer used to fetch under different query keys ('all' vs 'cardOnly'), so a location with sensors cost two requests per poll interval in card view, and the table's own request fired even with all three sensor columns hidden (the default). Both sites now share one key and one unfiltered fetch; the footer filters to `show_on_card` itself instead of asking the server for a second, differently-scoped copy. The fetch is also gated on actually needing it — card view always, table view only when a sensor column is visible.

**2. Write order in Options Save** — `persistDefaults` now does the interval PATCH first and the six localStorage writes after, so a rejected PATCH leaves every local preference untouched instead of the error toast following a save that partially happened.

**3. Reset-confirm closing the whole Options dialog** — `ConfirmModal` was rendered inside the Options overlay's `onClick={onClose}` div; its own overlay doesn't stop propagation, so dismissing it by clicking beside it bubbled up and closed Options too. Moved it to a sibling in a fragment, matching how `LocationHASensorModal` already does it.

**4. Overwrite path could lose a binding** — `overwriteMutation` now PATCHes the existing row onto the new entity instead of delete-then-create, so a failed create can no longer leave a location with no sensor bound and nothing to show for it. Also added the same close-while-pending guard `LocationsModal` already has (Escape, backdrop, and the X button all check `isPending` now) — an Escape landing mid-PATCH doesn't drop the user out from under an in-flight request.

**5. Location alerts riding on the printer toggle** — `on_location_ha_sensor_alert` now has its own `NotificationProvider` column instead of reusing `on_ha_sensor_alert`. A provider scoped to one printer's sensor alerts no longer silently also receives every drybox alert; there's a separate "Storage Location Sensor Alert" toggle in Add/Edit Notification now.

**6. Missing regression test for the polling gate** — added a test that a location entirely absent from `getLocationHASensors()` is never queried at all, in either view — not merely disabled, genuinely excluded from the request list.

**7. de.ts Sie/du** — my first fix made the section internally consistent (all du) but not with the file, which runs overwhelmingly Sie. Reverted the four du-form strings in the `locationHaSensors` block back to Sie to match the file's own convention.

**8. `on_location_ha_sensor_alert`'s unused param + missing docstring** — dropped `location_id` (never used — `NotificationLog` has no location column to log it against, unlike `printer_id`) and added a docstring matching `on_ha_sensor_alert`'s.

**9. `LocationsModal(startCreating)` rendering null while `open` stayed true** — a successful create with no `onPickLocation` used to fall through to the "close the inline editor" branch, which doesn't call `onClose`, leaving `editorOpen=false` / `open=true` — nothing left to render, but the parent never told to close. `startCreating` mode now always closes on a successful create, `onPickLocation` is called only if it was given.

**10. Auto-add prompt's Cancel button** — it still saves the primary sensor (right behavior, skips only the siblings), so the label was misleading. `cancelText` is now "Only this one".

**11. toFixed(2) vs raw value now that the display helpers are shared** — extracted `describeHASensorReading` into `haSensorDisplay.ts`; both the printer row and both location call sites use it now instead of three near-identical copies (two of which were byte-for-byte the same). Behavior is unchanged on both sides: printer stays raw, location stays fixed to two decimals for the column-alignment reason from before — the divergence is now a documented `decimals` option on one function instead of copy-paste drift.

**12. Battery joining the shared icon map affects the printer row too** — true, and confirmed: the printer's own map never had a `battery` entry (checked upstream), so a battery-class printer sensor fell back to the generic Gauge icon before this. Decided to keep the Battery icon for printer sensors too rather than carve out an exception — there's no principled reason a battery-class reading should render differently depending on which feature bound it, and the old fallback reads more like an omission than a considered choice. Documented in a comment and pinned with a test.

Also, separately from anything flagged: added a small hook (`useLocationSensorColorPrefs`) so the four location-sensor colour preferences update live across already-open pages (InventoryPage, SettingsPage) instead of only on next load — they used to be read once in a `useState` initializer per component, including once per `SpoolLocationFooter` card.

Every item above has a regression test, and each one was verified to actually fail without its fix before I called it done.

**More Changes**

**13. Blank Sensor Alert template preview** — `location_ha_sensor_alert` was missing from `SAMPLE_DATA` in `notification_template.py`, so its live preview in Settings rendered every placeholder empty (just ": is "). Added the missing entry, plus a regression test that walks `EVENT_VARIABLES`/`SAMPLE_DATA` for every event type so this class of bug can't recur silently for a future event.

**14. Printer template rename** — Renamed the `ha_sensor_alert` template from "Home Assistant Sensor Alert" to "Printer Sensor Alert" (migration only touches rows still on the old default name, so a custom rename is preserved). Once "Storage Location Sensor Alert" existed next to it, the old name no longer said which one was the printer's.

**15. Matching UI label rename** — The toggle/badge label everywhere outside the template editor (event settings list, provider badges, add-provider modal) still just said "Sensor Alert" — not updated when the template itself was renamed. Renamed to "Printer Sensor Alert" to match, across all 13 locales.

**Found but not fixed yet**

**16. Four toggles that display as off regardless of the real value**
 — `on_ha_sensor_alert`, `on_location_ha_sensor_alert`, `on_stock_reorder_alert`, and `on_stock_break_alert` are all missing from `_provider_to_dict`, so **no GET response ever returns them** — every toggle and badge for these four always renders off in the UI no matter what's actually in the database.

**17. Notification {timestamp}**
Notification {timestamp} ignores the configured date_format setting — not location-specific, applies to every event type (not merged, just found) — Noticed via the "Storage Location Sensor Alert" template preview, but it's systemic, not particular to location sensors:

Preview (notification_templates.py:163): renders SAMPLE_DATA[event_type]["timestamp"], a fixed string like "2024-01-15 14:30", identical hardcoded value for essentially all ~30 event types.
Real sends (notification_service.py:252): variables["timestamp"] = datetime.now().strftime("%Y-%m-%d %H:%M") — server local time, fixed YYYY-MM-DD HH:MM format, independent of anything user-configurable.
There's an existing, real, user-configurable date_format setting (settings.py:443) that the frontend respects everywhere else via formatDateTime/formatDateOnly/formatDateInput in utils/date.ts. The notification timestamp is the one place that's completely disconnected from it — both the preview sample and every real notification for every event type (print complete, printer offline, etc., not just the two sensor-alert ones) always render the timestamp in that one fixed format regardless of what's configured under Settings. 

Purely a finding, nothing touched.

**17. Some comments**
If the Location Storange Sensors are finished, I would improve the Printer Sensors to the same Multiview. 
I have many other ideas for some improvements, but not in this PR :)

maziggy:

maziggy (changes requested):
Thanks for this - it is well above the usual bar for a change this size. Auth on every route
with a written rationale for reusing SMART_PLUGS_* over INVENTORY_* (so an inventory-scoped
API key cannot create HA bindings) is exactly the reasoning I want to see. The PATCH handler
re-validating the merged row against the Create schema, the _AlertableSensor Protocol that
widens annotations without touching the printer path, and the was_alerting is None cold-cache
guard against restart alert-storms are all good calls.

I ran the suites: backend 10758 pass, frontend 3296 across 245 files, eslint clean, i18n
parity at 6224 leaves in all 13 locales, production build succeeds. Five things before this
can go in.

1. Blocking: the notification toggle can never be enabled, and misreports itself when it is.

backend/app/api/routes/notifications.py is not in the diff and needs to be. Both directions
are hand-maintained field-by-field maps - _provider_to_dict() at line 31 and the
NotificationProvider(...) constructor in create_notification_provider() at line 117 - and
on_location_ha_sensor_alert is in neither. I checked this at runtime rather than by reading:

  CREATE on_location_ha_sensor_alert -> False
  PATCH  on_location_ha_sensor_alert -> False
  GET    on_location_ha_sensor_alert -> False
  DB     on_location_ha_sensor_alert -> True

NotificationProviderResponse inherits NotificationProviderBase, so FastAPI fills the missing
key with the schema default and serialises false unconditionally. PATCH persists correctly
because the update handler is a generic setattr loop.

So ticking the box in the create modal is dropped, the card's switch snaps back off and stays
off, and a provider toggled via PATCH does receive alerts while its UI says the trigger is
disabled.

Two notes with it. on_ha_sensor_alert from #1148 has the identical defect today - grep for it
in that file returns nothing, and since on_location_ha_sensor_alert contains that string, one
grep covers both. Please fix both here. And test_location_sensor_alert_provider_scope_2824.py
cannot catch this: it builds NotificationProvider objects through the ORM and never exercises
the route. An API-level test is needed.

2. CI will fail on formatting.

backend/app/utils/natural_sort.py wraps its return across three lines; ruff 0.15.20 (the pin in
requirements-dev.txt) collapses it to one, and ci.yml:54 runs ruff format --check backend/.
Worth knowing that ./test_backend.sh is literally `ruff check && ruff format --check`, so that
checkbox in the description cannot have passed. Nothing has flagged it because only the seven
security workflows have run on 8984ea5 - the main CI workflow has not triggered on this SHA.

3. The static bundle was not rebuilt.

static/ is tracked here. This changes 20+ frontend files and ships no static/ changes, so
anyone running the branch without building gets a UI with none of the feature in it. I rebuilt
to confirm the hash moves off the committed index-e8_8xCtK.js.

4. A transient DB error kills the poller permanently.

In location_ha_sensor_manager.py:

  try:
      await asyncio.sleep(await self._get_poll_interval())
      break

_get_poll_interval() opens a session and queries Settings, and guards only the int() call with
(TypeError, ValueError). A pool exhaustion or a restarting database escapes the loop, the task
ends, and self._task is still set so start() will not revive it - location sensors stay frozen
until the process restarts. The same error one line earlier inside poll_once() is caught by
except Exception and survives, which is what makes it worth fixing. The printer sibling does
not have this because its sleep path does no I/O. Widening that second except to Exception is
enough.

5. "One sensor per category per location" is enforced only in the UI.

The API's uniqueness check is on (location_id, entity_id), so POST accepts two temperature
sensors on one location and the inventory column then .find()s whichever comes first. Either
enforce it server-side or drop the claim from the description.

Smaller things:

- locationSensorDefaults.ts keeps the per-category alert defaults in localStorage alongside the
  colour preferences, but the dialog editing them sits in Settings among server-backed
  settings. The colour prefs are fine as per-user. The alert-threshold defaults mean two admins
  binding sensors from different browsers seed different alert rules, and none of it reaches
  backup/restore. Worth reconsidering for the thresholds specifically.
- moisture is handled inconsistently: categoryFor() maps it to humidity, but
  findSiblingEntities() then tests endsWith('humidity'), which is false for a _moisture entity,
  so auto-bind silently finds no siblings. Separately the inventory humidity column accepts
  moisture, which is a binary device class, in a column formatted for percentages.
- list_locations switching from ORDER BY name to a Python natural sort is correct and nicely
  commented, but it is scope beyond the issue - flagging it so it is a decision rather than a
  surprise.
- N sensor-bearing locations means N polled queries. Cache-served and properly gated on view
  mode and visible columns, so not a blocker, but a ?location_ids= batch form would collapse
  them if you want it later.

maziggy:

MagicMelody84:
Thanks — all four addressed, plus a security pass over the feature that turned up a few things of its own. Closing the loop on the open items from last round first, since my last write-up left three of them hanging.

**Carried over from round 3**

- **The notification toggle that could never be enabled — fixed and in.** `on_ha_sensor_alert` and `on_location_ha_sensor_alert` are wired into both `_provider_to_dict` and the create route's constructor, with API-level tests covering create → read, PATCH → read and the list route. The open design question behind the revert (should the toggle be auto-derived from whether any sensor wants notifications?) is settled as **no**: deriving it is not provider-specific, so the only way to implement "automatic and non-editable" was to stamp the value onto every provider row — which would have started sending sensor alerts to channels set up for something else, with no way to opt back out. They stay ordinary per-provider toggles, as they were in #1148.

- **`on_stock_reorder_alert` / `on_stock_break_alert` — still open, deliberately.** Same defect as above, and worse: they are absent from the Pydantic schema entirely, so there is no way to set them through the API at all, even though `notification_service.py` already gates real stock-alert sends on those exact columns. That means stock alerts cannot have fired for anyone. Untouched here because it is outside this PR's subject — flagging it so it is a known gap rather than a silent one. 

**This round**

**1. Alert defaults in localStorage** — agreed on the split you drew. The three alert fields (`alertAbove`, `alertBelow`, `notifyOnAlert`) now live in a server-backed `location_sensor_alert_defaults` setting, next to the existing `location_sensor_poll_interval`, as a JSON string in the same shape as `preheat_filament_targets`. Empty = built-in defaults, so nothing needs migrating. The colour preferences stay per-browser as you suggested.

`showOnCard` deliberately stayed local, and I want to flag that as a decision rather than an oversight: `show_on_card` is a per-sensor column with its own checkbox in the form, so what localStorage holds is only the pre-selection. It is also a genuine per-person working preference — someone curating filament cards tightly wants new sensors hidden, someone bringing up a rack of dryboxes wants them shown, and unlike a threshold there is no correct answer to settle it with. Keeping it local also keeps the split legible: everything presentation-side in this dialog (the four colour prefs) is local, everything the installation runs on is shared. The Options dialog's "Reset" does still apply it installation-wide, which is intentional — that button exists so you can change your mind once instead of editing twelve sensors by hand.

Writing this up surfaced a bug in my own change: the effect that seeds the form from the server overwrote whatever was being typed the moment the settings query resolved — a field cleared and retyped came back as "3035". Fixed with a touched-guard so it seeds once and never over an edit in progress. Worth knowing that `pollInterval` in the same dialog has the identical race; I left it alone as out-of-scope, happy to fix it if you want it in this PR.

**2. `moisture`** — both symptoms were real, and the cause was one mapping doing two jobs. `categoryFor()` gates the entity picker *and* assigns the threshold/uniqueness category, so `moisture` had to be called "humidity" just to appear in the picker at all. That is what put a binary wet/dry sensor into a percent-formatted column and let a leak detector block a real hygrometer.

Rather than patch the two symptoms, `moisture` is no longer a location category at all — it drops out of the picker with it. A storage location wants a hygrometer; a leak detector is a printer-sensor case, and that path already handles it correctly as binary wet/dry via the shared display helpers, unchanged here. The `endsWith('humidity')` sibling bug disappears with the mapping. Also worth noting the API would have rejected the seeded thresholds anyway — the schema refuses `alert_above`/`alert_below` for `kind="binary"` — so the category was promising something it could never deliver.

**3. `list_locations` natural sort** — fair to ask for it to be a decision. Keeping it, deliberately: the sort is what makes the location dropdown in the sensor dialog usable ("Drybox 2" before "Drybox 10"), which is squarely this feature's surface. Reverting it would leave the dialog this PR adds worse than it needs to be. Called out in the PR description now rather than left as a surprise.

**4. N polled queries** — agreed it is not a blocker, and agreed a `?location_ids=` batch form is the right shape. Leaving it for a follow-up rather than growing the API surface in a PR already at round four. Noted so it does not get lost.

---

**Security / robustness pass**

Went over the feature adversarially rather than only re-running the suite. What it found:

- **`last_state` could sink an entire poll batch.** A numeric entity that starts reporting long free text (an enum, an error string) exceeds the `String(64)` column. SQLite stores it, PostgreSQL rejects the row — and because the poller commits its whole pass at once, one such sensor took down every sensor's update on every tick, and could swallow a same-tick alert notification with it. The poller now truncates what it persists, and compares against the truncated form so an unchanged-but-long state does not read as a change on every poll.

- **NaN/Infinity thresholds were accepted.** Pydantic's lax mode coerces the strings `"nan"`/`"inf"` into real floats. A NaN threshold satisfied the "notify needs an alert condition" rule yet every comparison against it is false — a notification that can never fire — and it slipped past the below-vs-above ordering check the same way. Now `allow_inf_nan=False` on the write paths. Reads stay tolerant, so a row that got one in before cannot 500 the whole list.

- **The duplicate-binding check was read-then-insert.** Two concurrent creates could both pass it. Backed by a unique index on `(location_id, entity_id)`; the route maps the resulting `IntegrityError` to the same 400 as the pre-check. The migration collapses any already-raced-in duplicates to the oldest row before building the index, since `CREATE UNIQUE INDEX` refuses to build over duplicates and would otherwise abort startup.

- **`entity_id` was bounded by its pattern but not its length.** `[a-z0-9_]+` is unbounded while the column is `String(255)`, so a direct API caller could store a 400-character id — accepted by both POST and PATCH. Same failure mode as `last_state`: fine on SQLite, `DataError` on PostgreSQL, and since the routes only map `IntegrityError` it would surface as a 500 rather than a 422. Now `max_length=255` on the write paths, reads left tolerant for the same reason as above. **This one is inherited — `printer_ha_sensor.py` has the identical gap**, since the location schema was copied from it, so it is fixed on both.

- Also capped `location_sensor_alert_defaults` at 2000 chars on write (the real payload is under 300 — this only stops a stray client parking megabytes in the settings table), and the frontend now slices an oversized Home Assistant unit to the `String(16)` column the way it already did for the name, since that field is snapshotted rather than typed and must not come back as a 422 on something the form never showed.

Checked and found sound, for completeness: mass assignment (`id`/`last_state` in the payload are ignored — Pydantic drops them), SQL injection and path traversal in `entity_id` (rejected by the pattern), name/`sort_order`/`location_id` bounds, cascade delete of a location with sensors attached, the readings endpoint for a non-existent location (empty list, no leak), and XSS (no raw HTML anywhere in these components). Permissions are `SMART_PLUGS_*`, which sit in `_APIKEY_DENIED_PERMISSIONS` — API keys cannot create or edit bindings, matching the printer sibling.

Every item above has a regression test, each verified to fail without its fix.
If you found more, give me round5 :)
For all HA Sensor related stuff, i will look into it after this is finished.

maziggy (approved):

Issue #2793, opened 2026-08-08
[Feature]: Support model sites besides MakerWorld (Thingiverse, Printables, ...)


pascalheidmann:
### Problem or Use Case

Makerworld integration is great but I would love to be able to import files from thingiverse, printables, you name it, as alternative. Right now importing means you have to download & upload in bambuddy again.

### Proposed Solution

Haven't looked at the code yet but splitting up the part that actually fetches things from makerworld and then handling all the stuff around if not already in place could allow for a module based solution where downloading files from other providers should be quite straight forward.

In case of Thingiverse they have an official REST-API in place: https://www.thingiverse.com/developers/swagger, Printables has not officially an GraphQL endpoint.

Implementation wise this would propably require a dedicated login for some platforms and UI changes to the Makerworld page but propably one can reduce those to a minimum as they all offer a list of files with some metadata and normally a preview image.

### Feature Category

Other

### Priority

Nice to have

### Contribution

- [x] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Please note our rules from CONTRIBUTING.md. Thanks!

Issue #1777, opened 2026-06-18
[Bug]: MakerWorld Import fails with "requires a Bambu Cloud login" when called via API key, even with a valid Bambu Cloud session


Mx772:
### Product

Bambuddy

### Area

API / integrations (Home Assistant, MQTT export, Obico, ...)

### Bug Description

Did some digging, I think this is what is going on here. 

`POST /api/v1/makerworld/import` (and `/makerworld/resolve`) fail with:
```
{"detail":"Downloading files from MakerWorld requires a Bambu Cloud login"}
```
when the request is authenticated via an API key (`X-API-Key` header), even though the user has an active Bambu Cloud login linked under Settings -> Bambu Cloud and the same download works fine from the web UI (session-authenticated).

### Because....
In `backend/app/api/routes/makerworld.py`, both `resolve_url` and `import_instance` take:
```python
current_user: User | None = RequirePermissionIfAuthEnabled(Permission.MAKERWORLD_VIEW),
```
Per `require_permission_if_auth_enabled()` in `backend/app/core/auth.py`, when the request is authenticated via API key, the dependency validates the key's scopes but **always returns `None`** for `current_user`:
```python
if x_api_key:
    api_key = await _validate_api_key(db, x_api_key)
    if api_key:
        _check_apikey_permissions(api_key, perm_strings)
        return None  # <-- no identity attached
```
Both routes then call `await _build_service(db, current_user)`, which does `get_stored_token(db, current_user)` to fetch the linked Bambu Cloud token. With `current_user is None`, this always returns no token, so `MakerWorldService` is built with `auth_token=None`, and `get_profile_download()` raises the "requires a Bambu Cloud login" error — regardless of the actual account's Bambu Cloud status.

This is the same problem #1182 fixed for cloud slicer presets, via `_user_from_api_key()`:
```python
async def _user_from_api_key(db: AsyncSession, api_key: APIKey) -> User | None:
    """Resolve the owner of a validated API key... Cloud routes... read the
    returned User to look up per-user state like cloud_token."""
```
but the MakerWorld routes were never updated to call it — they only use the dependency-injected `current_user`, which is always `None` for API-key callers.

### Should Be Fixed By....
In `resolve_url` and `import_instance`, when `current_user is None` but the request carried a valid API key, resolve identity via `_user_from_api_key(db, api_key)` before calling `_build_service`, mirroring the pattern used for cloud presets in #1182.

### Background
[...]

PR #2845, opened 2026-08-15
(Refactor): modularize import ("Makerworld tab")

pascalheidmann:
## Description

As first step for achieving #2793 this PR splits out makerworld code into a small provider strategy pattern wrapper.

## Related Issue

Refactor as preparation for #2793

## Documentation

**Pick one**:
- [ ] Docs PR(s) linked above
- [x] No docs update required — reason: refactor only

## Type of Change

- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- api flow for makerworld is now: `Makerworld API` -> `Model provider registry` -> `Makerworld` (or other provider when implemented) instead of directly API -> Makerworld service

## Testing

## Checklist

## Additional Notes

I do not programm much with python these days anymore so I'm not that proficient with its specifics ;)


maziggy (changes requested):
Thanks for this - the direction is right, and the extraction itself is careful work.

First thing worth knowing: **no CI has run on this branch.** Workflow runs from forks sit at `action_required` until a maintainer approves them, so nothing has ever checked this PR. That is on us, not you. Two commands reproduce everything
below locally:

    python -m ruff check backend/          # 7 errors
    python -m pytest backend/tests -q      # 15 failed

### Blockers

1. **`MakerWorldService` is no longer imported** in `routes/makerworld.py`. The old import block went away and nothing replaced it, so `:142` in `proxy_thumbnail` raises `NameError` - every thumbnail on the page 500s. The return annotation at `:71` is the same name (harmless, it is lazy).

2. **`source_type` is undefined in `import_instance`.** It is assigned at `:203` inside `resolve_url` and read at `:440` inside `import_instance`. Ruff flags both ends (F841 + F821).

3. **`body.url` does not exist on `MakerWorldImportRequest`.** `:332` and `:379`
   pass it to `_build_service` / `find_for_url`, but the schema carries `model_id` / `profile_id` / `instance_id` / `folder_id`. This is the one I would think about before the others: the registry routes on URLs, but import identifies a model by numeric id, so there is no URL to route on. Either the request gains a `source_type` and the route uses `registry.get(source_type)`, or import re-derives the provider some other way. Every future endpoint hits this same wall, so it is worth settling now.

4. **`provider` can be `None` and is dereferenced.** `:202` handles it   (`provider.source_type if provider else ...`) and then `:205` calls    `provider.parse_url(...)` unguarded; same at `:264` and `:379`. Your own    `test_find_for_url_returns_none_for_foreign_hosts` pins that this happens. A pasted Thingiverse link used to be a clean 400; it is now a 500. `_build_service` gets this right at `:78-82` - that guard just needs hoisting.

5. **Auth-disabled installs stop recording a rejected Bambu token.** `makerworld/provider.py:85` skips `on_auth_failure` when `user_id is None`, but `mark_cloud_token_invalid(None)` is a real path: `cloud.py:214-222` branches on exactly that to write the global flag used when auth is disabled, which is the default single-user setup. Those installs do have a token (`get_stored_token(db, None)` reads global settings), so this leaves the import button enabled against a dead credential. dev passed the callback unconditionally.

Also: `test_makerworld_routes.py:70` calls `_canonical_url(1400373, provider=...)` while the shipped signature is `_canonical_url(provider, model_id, profile_id=None)`, so those three tests `TypeError`.

### Design, once it runs

- `resolve`, `get_download`, `download` and `get_status` have no callers. The routes still call `get_design` / `get_profile_download` / `download_3mf` directly and `/status` is untouched, so there are two implementations of the same flow that will drift. Either wire the endpoints onto the interface here, or land those methods with their first caller.
- `build_service` takes `api_key_owner` but `_build_service` never passes it - the route pre-collapses `current_user or api_key_cloud_owner`. Worth picking one place to resolve identity.
- `makerworld/auth.py` imports from `api/routes/cloud.py`, so the service package now depends on the route layer. dev avoided that by injecting the callback from the route. No cycle today, but it is a trap for provider #2.
- Small stuff: `SOURCE_TYPE` at `:66` is never assigned or read; `find_for_url` runs twice on the same URL in `resolve_url`; the `recent_imports` query change is a behaviour change that would be easier to review alongside the provider that needs it.

### What is good

I diffed every moved function by AST: `_get_json`, `_headers`, `_note_auth_failure`, `_s3_ssl_context`, `_download_s3_urllib`, `_extract_upstream_error`, `download_3mf`, `fetch_thumbnail`, `get_design`, `get_design_instances`, `get_profile`, `get_profile_download` and `close` are all byte-identical. The SSRF allowlist, the size caps and the certifi-pinned S3 TLS context came through untouched - that is the part that would have been easy
to break.

`url.py` also preserves the canonical dedupe shape exactly, so existing `source_url` rows keep matching, and having the MakerWorld errors subclass both hierarchies means nothing downstream had to change. Both good calls.

Happy to look again once the five blockers are fixed - and tell me what you think on (3), since it shapes the rest.

maziggy (changes requested):
Second pass. All five blockers are fixed, and so is everything under "Design, once it runs".

Verified locally on 8192a43: ruff check clean, ruff format clean, 10788 tests pass (-n 30).
CI still has not run - only GitGuardian has reported, so the workflow run needs approving.

Blockers, one by one: MakerWorldService is imported again and proxy_thumbnail resolves;
source_type is a real schema field with a "makerworld" default, so existing callers are
untouched; _provider_for_url turns an unroutable link into a clean 400 before anything is
dereferenced, and the test pins it; and provider.py passes on_auth_failure unconditionally
with user_id=None flowing through to the global flag, which test_anonymous_user_still_gets_
auth_callback holds in place. On (3) you picked the option I would have picked - naming the
provider on the request is the honest answer when there is no URL to route on.

The credential extraction is the best part of this round. bambu_cloud_credentials.py breaks
the services -> routes dependency properly rather than papering over it, and re-exporting
from cloud.py means preset_resolver, github_backup and the eleven test modules that import
those names never notice. I re-diffed the moved functions by AST again: still byte-identical.

What is left, in order:

1. thumbnail_hosts() is documented as the SSRF seam and nothing calls it. base.py tells
   implementers "Must restrict the upstream host to ModelProvider.thumbnail_hosts", but
   fetch_thumbnail checks MAKERWORLD_CDN_HOSTS imported from url.py. Same tuple today, so
   nothing is exposed - but the contract is enforced by coincidence, and provider #2 can
   implement thumbnail_hosts() and have it silently ignored. Worth wiring while it is
   still one provider.

2. get_download lost its test coverage in the move. It is the most intricate piece the
   provider owns - alphanumeric modelId resolution, the two-tier profile fallback, three
   502 guards - and nothing executes it now. The old route tests ran that code incidentally
   because _default_design() supplied modelId and instances; the new tests stub it out with
   _download_info(). A TestGetDownload alongside TestGetProfileDownload would cover it.

3. TestCanonicalUrl now asserts against _MockProvider.canonical_url, which reimplements the
   real logic, so it pins the copy rather than the dedupe shape its docstring claims. Using
   makerworld_provider makes it real again (test_model_provider_interface already covers the
   shape, so the mock could go entirely).

4. _provider_for_source passes str(KeyError) into the detail, and KeyError.__str__ is the
   repr of its argument, so the body comes back as "\"No model provider registered for
   source_type 'thingiverse'\"" - quotes included. Building the message in the route fixes it.

5. Small stuff:
   - cloud.py's re-export block carries "# noqa: E402,F401" but only F401 is needed. I moved
     it to the top and imported backend.app.main cleanly - the placement is the only reason
     for the E402.
   - Declared and never read: default_folder_name, view_permission, import_permission,
     can_download on ModelProvider, ProviderAuthConfig.credential_fields, and
     ProviderResolvedModel.already_imported_library_ids (documented as "filled in by the
     route layer", which the route never does). Two of these the route could adopt now -
     the hardcoded "MakerWorld" folder name and the hardcoded Permission.MAKERWORLD_* on the
     decorators - and adopting them would prove the descriptor rather than describe it.
   - get_download's instance_id parameter is unused, and reintroduces a field the schema
     documents as "Retained for backwards compatibility; no longer used".
   - The SSRF allowlist is split: MAKERWORLD_CDN_HOSTS lives in url.py ("URL parsing and
     canonicalisation") while _ALLOWED_DOWNLOAD_SUFFIXES lives in http.py, whose docstring
     claims the CDN guard helpers. download_3mf imports one from each.
   - _map_service_error's fallback still reads "MakerWorld error: {exc}" - the last message
     in that file that names the provider.

maziggy (changes requested):
Thanks for this - it is a well-judged split, and the interface is better
documented than most of what is already in the repo. The provider/service
separation is the right seam, and resolving the provider before the folder
block so an unknown source_type 400s instead of auto-creating a folder is a
real bug fixed in passing.

What I checked: the bambu_cloud_credentials extraction is faithful (logic
identical, cloud.py re-exports so nothing downstream breaks); source_type
defaults to "makerworld" so the frontend needs no change; the SSRF guards
survived the split correctly, with the thumbnail allowlist now
provider-driven and the 3MF download allowlist deliberately left as its own
module constant; owner_id attribution for API-key callers (#1777) is
preserved; and no test coverage was lost - the five route tests that
disappear moved with the code they cover, net +11. 226 tests in the touched
areas pass, 11225 collect clean, ruff clean.

One thing to fix before merge, then three notes you can take or leave.

Fix: in import_instance, lines 281 and 288 read the default folder off the
makerworld_provider singleton instead of the provider you just resolved on
line 259. Everywhere else in that function correctly uses the resolved
line 259. Everywhere else in that function correctly uses the resolved
object - provider.source_type on 301 and 365, provider.canonical_url on 314.
It is latent today because MakerWorld is the only registered provider, so
the two are the same object and no test can see the difference. It goes live
with the second provider: a Thingiverse import would land in the MakerWorld
folder. Two-word change.

Note 1: the route derives sign_in_expired from `status.auth_error is not
None` (line 189), but the interface documents auth_error as a
human-readable reason. A provider that sets it for a network error or a rate
limit would report "your sign-in expired". An explicit credential_rejected
bool on ProviderStatus would make that reading honest instead of
conventional.

Note 2: thumbnail_hosts() is a declared seam with the SSRF contract stated
on fetch_thumbnail, but the 3MF download guard has no counterpart. A new
provider gets no structural hint it needs one. A download_hosts() would make
the pair symmetric.

Note 3: find_for_url returns the first match in dict insertion order with no
documented tie-break, and default_folder_name is typed str | None but goes
straight into LibraryFolder(name=...), so a provider that leaves it unset
creates a NULL-named folder. Both are fine with one provider; worth a
docstring sentence and a guard respectively.

No CHANGELOG needed on this one.

pascalheidmann:
@maziggy can you have another look at this?

maziggy (approved):

PR #2853, opened 2026-08-16
Add printer video downloads and range selection

### LINKED ISSUE #2850: [Feature]: Download printer videos from print history and improve multi-file selection
# [Feature]: Download printer videos from print history and improve multi-file selection

## Problem or Use Case

Timelapse files and continuous IP-camera chunks can remain on a printer after a print, but print history does not provide a direct way to find and download the files associated with that run. The printer file browser also buffers multi-file ZIP downloads in both server and browser memory, which can fail with IP-camera chunks that are commonly hundreds of megabytes each. Selecting a long contiguous range currently requires clicking every file individually.

## Proposed Solution

- Add a Print videos action to archive cards and list rows.
- Find an attached timelapse, a matching printer-side timelapse, and IP-camera chunks whose timestamps overlap the print window.
- Let users choose printer-side videos and download them in a disk-backed ZIP through a short-lived, single-use token.
- Use the same disk-backed browser-native download flow for multi-selection in the printer file browser.
- Add row selection and Shift-click range selection based on the visible search/sort order.
- Keep FTP file browsing available when MQTT live status is offline, since the services are independent.

## Alternatives Considered

- Continue using `fetch()` and `response.blob()`: rejected because it buffers the complete ZIP in browser memory and the old server path also held source files and the ZIP in memory.
- Attach or copy every matching video into the archive automatically: rejected because IP-camera recordings can be very large and listing/downloading should remain an explicit, read-only user action.
- Select every matching IP-camera chunk automatically: rejected to avoid accidental multi-gigabyte downloads.

## Feature Category

Print Archiving

## Priority

Would improve my workflow

## Contribution

I would be willing to help implement this feature.

## Checklist

- [x] Searched open and closed issues for printer file downloads, timelapse downloads, IP-camera chunks, and multi-select downloads; no matching request was found.

logikal:
## Description

Adds printer-side video discovery and reliable, cancellable multi-file downloads. Print history can now find attached or printer-side timelapses and matching IP-camera chunks, while the printer file browser supports row clicks and Shift-click range selection.

Large selections are prepared asynchronously, one file at a time, under the app data volume. The browser polls progress and receives a short-lived, single-use, printer-bound download token only after preparation succeeds. Video entries use stored Zip64 entries; G-code, 3MF, and other files are deflated. A 10 GiB cap, actual-byte accounting, free-space reserve, and cross-worker staging lock bound disk use.

## Related Issue

Fixes #2850

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#69

Website: not required because this change does not alter installation, upgrades, public claims, or the website feature list.

**Pick one**:
- [x] Docs PR(s) linked above
- [ ] No docs update required

## Type of Change

- [ ] Breaking change
- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [x] Test addition or update

## Changes Made

- Add archive media discovery for attached timelapses, matching printer timelapses, and print-window IP-camera chunks, including alternate storage directories and explicit unavailable warnings.
- Add asynchronous, cancellable, data-volume-backed printer download jobs with progress, partial-failure reporting, a 30-minute bound, a 10 GiB cap, actual FTP-byte enforcement, reserved-space checks, and cross-worker cancellation/capacity coordination.
- Use browser-native attachment downloads without buffering the response in JavaScript or hidden iframes.
- Reject truncated FTP transfers using fresh server size metadata; fall back to the client hint only when the printer does not support `SIZE`.
- Enforce JWT/API-key permissions and printer allowlists consistently, while keeping attached archive media available without printer-file access.
- Preserve legacy paths-only ZIP requests, duplicate basenames, relative-path rejection, empty-result behavior, and safe bounded Unicode filenames.
- Fix multi-file downloads and add visible-order Shift-click range selection that always selects the range; prune hidden/stale selections after navigation and filtering.
- Keep printer storage browsing available when MQTT status is offline.
- Add startup/periodic staging cleanup and await active preparation tasks during application shutdown.
- Add backend/frontend regression tests and real translations for all 13 locales.

## Screenshots

| Before | After |
| --- | --- |
| Offline printer cards do not offer storage access.<br>Before: offline printer cards without file browsing | Offline printer cards retain the Browse printer files drive action in the footer.<br>After: offline printer cards with the file browser action in each footer |
| The archive has no print-video download action.<br>Before: print archive without video download actions | Print history discovers matching timelapses and IP-camera chunks.<br>After: Print videos dialog with printer timelapse and IP-camera chunks |

Shift-click selects the visible range and enables one multi-file download:

Three contiguous printer files selected with Shift-click

## Testing

- [x] I have tested with my printer models: X1C and H2D
- [x] Host backend — 10,501 tests passed with two workers; Ruff lint/format passed
- [x] Host frontend — 236 files / 3,177 tests passed under Node 22; TypeScript, ESLint, production build, and i18n parity passed
- [x] Docker backend — 10,466 passed / 22 environment-specific skips with two workers
- [x] Docker frontend — 236 files / 3,177 tests passed with two workers
- [x] Docker integration — 2,884 tests passed against an isolated healthy container; production image import/static/health/API/root checks passed
- [x] Security — Bandit found no medium/high issues; pip-audit and npm audit found zero known vulnerabilities
- [x] `pre-commit run --all-files` — all hooks passed
- [x] Companion wiki — strict MkDocs build passed and PR updated
- [x] Post-merge affected suites — 104 FTP, 420 remaining backend, and 126 frontend tests passed; lint/typecheck/i18n/build/pre-commit passed
- [x] LAN integration — the offline-MQTT H2D root listed over `192.168.1.11:8001`; a two-file asynchronous ZIP reported 2/2, downloaded with both entries intact, and removed its staged bundle after consumption

## Checklist

## Additional Notes

- Token-minting/status/cancel routes remain behind normal authentication and permission middleware. Only native `/dl/` consumers are middleware-public, and they independently require single-use resource-bound tokens.
- Preparation happens before token minting, so long FTP work does not consume token lifetime and the UI can report storage, timeout, cancellation, and partial-result errors before navigation.
- IP-camera chunks are not automatically selected when multiple matches exist because individual chunks can be hundreds of megabytes.
- FTP LIST mtimes matched UTC-naive archive timestamps on tested X1C and H2D firmware; LIST provides no timezone signal for reliable conversion on firmware that reports local time.
- The production data directory was never modified. Development and Docker validation used isolated app data, containers, networks, volumes, and ports.
- The branch is merged with current upstream `dev` through `7e77bf58`.


maziggy (changes requested):
Thanks - the disk-backed staging is the right fix for the /ipcam case, and the token flow
is implemented carefully (single-use, resource-bound, and the allowlist test entry is a
good habit). I ran everything here: ruff clean, 10440 backend tests, eslint and tsc clean,
3149 frontend tests, build clean. Findings below, worst first.

1. ZIP_STORED is applied to everything. Right for video, but the file browser's usual case
is gcode and 3mf, which deflate five to ten times. Multi-file gcode downloads now move a lot more bytes over the LAN and onto disk than they did before. Choose per entry: stored for VIDEO_SUFFIXES, deflated otherwise.

2. Staging writes to the system temp dir. build_printer_files_zip uses tempfile.mkdtemp(), which on the appliance image and under systemd PrivateTmp is a tmpfs - RAM. That reintroduces the exact amplification this PR removes, on the hardware least able to take it, and nothing checks free space before staging up to 1000 paths of ~250 MB. Stage under the app's own data directory (archive_dir/temp is already used that way) and fail early when the volume cannot hold the selection.

3. The public pattern is wider than its comment. "/files/download-zip/" is matched with
`pattern in path`, so it also matches /printers/{id}/files/download-zip/token and the
token-minting route skips the gateway too. Not a bypass - I traced require_permission_if_auth_enabled and it independently verifies the JWT signature, jti
revocation, user active state and _is_token_fresh, so a revoked or stale token still cannot mint one - but the middleware layer is lost for no benefit. Either name the mint route so it does not contain the pattern (.../files/zip-token) or match the consume route exactly.

4. Download failures are invisible. The form posts into a hidden iframe, so an expired
token (403), "No files could be downloaded" (404) or a 500 renders inside the iframe while the toast has already said the download started. ArchiveMediaDownloadModal.downloadSelected has no catch at all, so a failing token request is an unhandled rejection and the only visible effect is the spinner resetting. Catch the token step and toast at minimum.

5. Nothing bounds or reports the long request. The ZIP is fully built before FileResponse
begins, at timeout=600 per file and up to 1000 paths, so one request can hold a worker for a very long time with no progress, no cancel and no size ceiling. The UI already knows each file's size - a cap on total selected bytes would be cheap.

6. Partial failures vanish. build_printer_files_zip returns a success count and both callers
discard it. Select twelve chunks, get nine, no indication. The number is already there.

7. /archives/{id}/printer-media lists printer storage under archive permissions. It requires
ARCHIVES_READ_*, then LISTs /timelapse and /ipcam. Someone with archive read but without printers:files learns file names, sizes and timestamps they cannot otherwise browse (the download itself is still blocked by the token route). Add the printer-files permission or record the decision in the docstring.

8. The same endpoint can hang for a minute. Two list_files_async calls at the 30s default, with an offline printer being the normal case for an older archive. The codebase already has ftps_handshake_blocked for this - the 3MF probe uses it - so pass a shorter timeout and skip when the handshake is in cool-off.

Smaller:

- match_ipcam_chunks treats FTP mtimes as UTC-naive. That matches archive timestamps on the models you tested, but LIST-parsed mtimes are local time on some firmware, where the -1/+10 minute window would match nothing. Worth saying in the docstring which models it was measured against.
- downloadLocalTimelapse clicks an <a> that is never appended to the document, which is unreliable outside Chromium; the other download helpers in client.ts append and remove.
- The iframe is removed on a 24 hour timer, so a session of repeated downloads accumulates hidden iframes. Removing on its load event does the same job.
- Shift-click deselects the range when the clicked row was already selected, where file   managers always select. Fine either way, but the tooltip should say which it does.
- The offline card banner duplicates the footer button, which this PR already enables   unconditionally. On a fleet with several offline printers that is a lot of repeated   explanation - I would keep the enabled button and drop the banner.
- The autouse fixture in test_scheduled_drying_routes hides status cached by another module rather than fixing the leak. Fine as a stopgap; please note in the docstring that the cache crossing xdist modules is the actual bug, so it does not get read as intended   behaviour.
- Two tests I would like: a token minted for printer A rejected on printer B (that nonce   binding is the security property), and build_printer_files_zip skipping relative and NUL-bearing paths.

logikal:
Addressed all findings from the review in `e3f55d30`:

- ZIP compression is now selected per entry: video stays stored; G-code, 3MF, and other files are deflated.
- Printer ZIPs stage under the app data volume, with a 10 GiB selection cap, free-space preflight, 256 MiB reserve, actual-byte enforcement, and stale-bundle cleanup.
- The protected preparation route is now `/files/zip-token`; only the exact token-consuming route is middleware-public.
- Preparation failures and partial downloads are surfaced before the browser download starts; hidden iframes and local-download anchors are cleaned up.
- Archive printer-media discovery also requires `printers:files`, observes the FTP handshake cool-off, and uses short listing timeouts.
- Shift-click always selects the visible range, the duplicate offline banner was removed, and the FTP timestamp/xdist-cache assumptions are documented.
- Added regression coverage for cross-printer token rejection, relative/NUL path rejection, compression choice, staging/free-space limits, permissions, FTP cool-off/timeouts, exact route matching, and frontend partial/error behavior.

Validation is green on the exact review commit:

- [CI](https://github.com/logikal/bambuddy/actions/runs/32049965851)
- [Security Audit](https://github.com/logikal/bambuddy/actions/runs/32049985575)
- Local `./test_docker.sh`: all 11 gates, including 10,412 backend tests, 3,151 frontend tests, and 2,856 integration tests
- Local `pre-commit run --all-files` and strict companion-wiki build

The PR description and [companion wiki PR](https://github.com/maziggy/bambuddy-wiki/pull/69) have been updated as well.

maziggy (changes requested):
Round 2. All eight majors, the seven smaller points and both requested tests are genuinely addressed - I checked each against the code rather than the summary, and the cross-printer test correctly asserts the token survives the wrong-printer attempt, which was the property I cared about. Verified locally: ruff check and format clean, 10477 backend tests (unit plus integration), eslint clean, build clean, 118 frontend tests in the touched suites, i18n parity at 6162 keys across 13 locales with the new strings really translated.

One thing I would fix before merge, then three to decide on.

**The ZIP is assembled on the event loop.** `build_printer_files_zip` is async and the FTP transfer is correctly off-loop (`download_file_async` goes through `run_in_executor`), but `archive.write(...)` is a plain blocking call in the middle of it - as are `_prune_stale_bundles` and the `shutil` calls. Measured on x86:

```
realistic G-code, DEFLATE   25 MiB/s (ratio 0.31)  -> 200 MiB blocks for 8.0 s
video, STORED (page cache)  ~1 GiB/s               -> 200 MiB blocks for 0.2 s
```

A ten-file G-code selection freezes every other request - MQTT handling, the UI, other downloads - for the better part of a minute, and a Pi is several times slower. `await asyncio.to_thread(archive.write, ...)` per entry fixes it. The deflate ratio also confirms the per-entry compression was worth doing.

**The archive button is not gated on the permission the endpoint now requires.** `/archives/{id}/printer-media` correctly gained `printers:files`, but the new Film button in `ArchivesPage.tsx` is only disabled on `!archive.timelapse_path && (!archive.printer_id || !archive.started_at)`. A user with archive permissions but without `printers:files` sees an enabled button and gets a red "Could not check the printer for videos." `hasPermission` is already in scope at `ArchivesPage.tsx:313` and used by the neighbouring buttons. The sharper version: the endpoint 403s before returning `local_timelapse`, which needs no printer access at all, so that user also loses the attached timelapse the modal exists to hand them. Checking the permission inside and degrading - local file plus a warning - beats refusing the whole endpoint.

**Abandoned bundles sit for 24 hours.** `SLICER_TOKEN_EXPIRE_MINUTES` is 5, so a prepared bundle is unusable after five minutes, but `_STALE_BUNDLE_SECONDS` is 86400 and `_prune_stale_bundles` only runs when someone else starts a download. Prepare 10 GiB, close the tab, and the data volume carries it for a day - on the appliance, potentially until the disk fills. An hour would be generous, and pruning at startup closes the "nobody downloads again" case. Unlinking a file a FileResponse is still streaming is safe on POSIX, so a short cutoff cannot cut off an in-flight download.

**The legacy JSON route is now a breaking change.** Its docstring says "Kept for API clients", but `PrinterFilesDownloadRequest` makes `sizes` mandatory with one entry per path, so an existing client posting `{"paths": [...]}` gets a 422. The wiki PR documents the new field, which helps, but `sizes: dict[str, int] = Field(default_factory=dict)` would keep those clients working - the enforcement that matters is the actual-byte accounting in the loop, not the preflight.

Smaller:

- `PUBLIC_API_REGEXES` hardcodes `/api/v1`, but routers mount at `settings.api_prefix` (`config.py:105`, env-overridable). Change the prefix and the exemption stops matching and the download 401s with no clue why. Build the pattern from the setting.
- The preflight trusts client-declared `sizes`; the real cap only trips after the offending file is already on disk, so an under-reported size can overshoot by one file. Acceptable - worth a comment saying so.
- Two concurrent preparations pass the free-space check independently. The per-file recheck limits the damage, so this is a note rather than a request.
- `ArchiveMediaDownloadModal`: the auto-select-single-file effect keys on `remoteFiles`, whose identity changes on every refetch (60s staleTime, refetch on focus). Deselect the single file, tab away, come back, and it is selected again.
- Both new modal mounts carry `{/* Timelapse Viewer Modal */}`. The file now has three identical comments, two of them describing the wrong modal.
- `selectAllFiles` still re-implements the search filter instead of using the new `visibleFiles` memo; shift-range and Select All will drift the first time that filter changes.

logikal:
Addressed all Round 2 findings in `dadc80ff`:

- Offloaded ZIP writes, pruning, disk checks, directory creation, stats, unlinks, closes, and cleanup from the event loop.
- Kept attached timelapses available to archive readers without `printers:files`; printer discovery is skipped with a permission warning, and remote-only buttons are gated in both archive layouts.
- Reduced abandoned-bundle retention to one hour and added startup pruning.
- Restored compatibility for legacy `{"paths": [...]}` bulk-download requests while retaining optional size preflight hints and actual-byte enforcement.
- Derived the public token-consumer regex from the configured API prefix and documented preflight under-report/concurrency behavior.
- Kept manual single-file deselection stable across query refetches, corrected the modal comments, and made Select All reuse `visibleFiles`.
- Added regression tests for each changed behavior and updated the companion wiki PR.

Validation is green on the exact review commit:

- [CI](https://github.com/logikal/bambuddy/actions/runs/32150691301): all lint, audit, frontend/backend, production build, Docker, and 2,856-test integration jobs passed.
- [Security Audit](https://github.com/logikal/bambuddy/actions/runs/32150688678): Bandit, dependency audits, Trivy image/IaC scans, and SARIF uploads passed.
- Local: Ruff, pre-commit, 10,446 backend tests (22 skipped), 3,171 frontend tests, TypeScript, ESLint, i18n parity, production image/health/API/static checks, and strict MkDocs build all passed.

The PR description and [companion wiki PR](https://github.com/maziggy/bambuddy-wiki/pull/69) are updated as well.

logikal:
Addressed the complete third-party review set and re-audited both maintainer review rounds on `1376212e`. Finding-by-finding:

1. **Long requests, RAM amplification, cancellation, and proxy timeouts** — preparation is now an asynchronous disk-backed job (`POST`, poll, cancel), bounded to 30 minutes and 10 GiB. FTP cancellation is cooperative even inside the worker thread; cross-worker cancel markers and persisted job status make polling/cancellation work across app workers. Shutdown now cancels and awaits active jobs.

2. **Staging capacity and concurrency** — staging is under `archive_dir/temp`, never system tmpfs. Client sizes are early hints only; fresh FTP `SIZE`, callback byte counting, per-write free-space checks, a 256 MiB reserve, and a cross-process lock enforce actual usage and prevent concurrent workers spending the same free-space observation. Startup, periodic, and request-time cleanup remove abandoned bundles/status files.

3. **Compression regression** — video suffixes use `ZIP_STORED`; G-code, 3MF, and other entries use `ZIP_DEFLATED`.

4. **Authentication and authorization** — JWT and API-key paths now share printer-aware permission/allowlist enforcement. Cross-printer token reuse is rejected without consuming the token. Archive discovery returns an attached local timelapse to archive readers while skipping remote printer listing with a warning when `printers:files` is absent. Local archive media also uses an auth-bound, single-use token instead of a reusable public URL.

5. **Middleware/public-route scope** — token minting, job status, and cancellation stay behind normal middleware. Only native `/dl/` consumers use the existing exact public convention, then validate their resource-bound token independently; the broad substring/config-prefix regex is gone.

6. **Invisible and partial failures** — the UI polls preparation, displays progress, supports cancel, catches token/job failures, and reports requested/successful/failed counts. Failed/expired downloads no longer disappear into an iframe. Native `<a>` downloads are appended and removed, including local timelapses.

7. **FTP integrity and availability** — downloads stage directly to disk, use fresh server size as authority, fall back to the client hint only if `SIZE` is unsupported, reject short reads, enforce actual-byte limits during `RETR`, and wait for the FTP worker before unlinking on cancellation. Listing distinguishes unavailable storage from an empty directory and observes the FTPS cool-off/short timeout.

8. **Archive discovery correctness** — database sessions end before FTP I/O. Discovery checks alternate timelapse/IP-camera directories, excludes files already present at the print-start baseline, bounds unfinished print windows, and documents the measured X1C/H2D timestamp assumption. Tests cover camera-only layouts, baseline filtering, unavailable printers, and stale unfinished archives.

9. **Browser/file-picker behavior** — browser-native navigation removes blob/iframe buffering and iframe accumulation. Shift-click always selects the visible contiguous range; Select All shares the same visible list; selections hidden by filtering/navigation are removed; a manually deselected single archive file is not reselected on refetch. Offline UI keeps the footer action without duplicating a banner.

10. **Compatibility and filenames** — the legacy synchronous ZIP route still accepts `{"paths": [...]}`, preserves duplicate basenames and empty/all-failed behavior, and rejects relative/NUL paths. Download filenames strip path/control characters and are bounded safely for long Unicode names.

11. **Tests, fixtures, translations, docs** — centralized printer-manager cleanup removes the actual xdist cross-module leak rather than hiding it in scheduled-drying tests. New tests cover token binding, malformed paths, compression, capacity serialization, cancellation/shutdown, short transfers, permissions, legacy contracts, progress, partial failures, selection behavior, and filename bounds. All 13 locale trees have real translations. Wiki PR #69 now documents the asynchronous job lifecycle.

Validation on this branch:

- Host: 10,501 backend tests; 3,177 frontend tests; Ruff, ESLint, TypeScript, build, 13-locale parity, and pre-commit clean.
- Docker: 10,466 backend passed / 22 environment skips; 3,177 frontend passed; 2,884 integration passed; production image and health/API/static checks clean.
- Security: Bandit no medium/high findings; pip-audit and npm audit clean.
- Post-upstream-merge: 104 FTP + 420 other affected backend tests and 126 affected frontend tests passed; lint/typecheck/i18n/build/pre-commit clean.
- Real LAN smoke: with persistent MQTT disabled in isolated dev, the H2D filesystem listed over `192.168.1.11:8001`; two selected G-code files prepared asynchronously, reported 2/2, downloaded as an intact ZIP with both entries, and left no staged bundle.
- Companion wiki strict build passed and the update was pushed to maziggy/bambuddy-wiki#69 (`4c48842`).

Production `/zpool1/docker_configs/bambuddy` was not read from or modified during this round. Docker checks used uniquely named test images/containers and an isolated project network/volume; those resources were removed afterward. Production and `bambuddy-dev` remained running.

logikal:
Final-head automation is green on `1376212e8a04a33b41473a00554644c6dd01bdd0`:

- [CI](https://github.com/logikal/bambuddy/actions/runs/32186757582): backend and Docker backend matrices, frontend tests/lint/typecheck/build, dependency audits, production image checks, and the integration suite all passed.
- [Security Audit](https://github.com/logikal/bambuddy/actions/runs/32186527130): Bandit, pip-audit, npm audit, and Trivy container/IaC scans all passed.

This is in addition to the local host, Docker, isolated-container, and real LAN/printer checks documented above.

maziggy:

logikal:

Rebased!

PR #2854, opened 2026-08-16
Show K-profile value on AMS slot card

### LINKED ISSUE #2532: [Feature]: Add K-Profile to loaded filament
### Problem or Use Case

Right now, when a filament is loaded, you have to hover over it to see the K-profile. 

### Proposed Solution

It would be great if we could have the K-profile shown under the filament that is loaded as it is in bambu studio.

### Alternatives Considered

_No response_

### Feature Category

K-Profiles

### Priority

Nice to have

### Mockups or Examples

The image below is what I am trying to suggest.

Image

### Contribution

### Checklist

gyrene2083:
Adds an always-visible K-value line under the material name on AMS, AMS-HT, and external/dual-nozzle slots, matching Bambu Studio's own display. Closes #2532.

## Description

Currently, the K-profile (pressure advance) value for a loaded spool is only visible by hovering over an AMS slot. This adds a small, always-visible line under the material name showing the K-value directly, matching how Bambu Studio's own UI displays it. Covers standard AMS slots, AMS-HT slots, and the external/dual-nozzle spool holders.

## Related Issue

Fixes #2532

## Documentation

**Pick one**:
- [ ] Docs PR(s) linked above
- [X] No docs update required — reason: wiki screenshot update will follow separately

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- Added an always-visible K-value display line to the standard AMS slot card
- Added the same display to the AMS-HT (high-temp) slot card
- Added the same display to the external/dual-nozzle spool holder card
- Added a new test file verifying the K-value shows on loaded slots and is correctly hidden on empty slots

## Screenshots
Screenshot 2026-08-16 at 11 20 17

Screenshot 2026-08-16 at 11 20 44

## Testing

- [X] I have tested with my printer model: H2C

## Checklist

## Additional Notes

I tested this a few different ways before submitting. All the existing tests still pass, and I added two new ones just for this change — one checking the K-value shows up when a slot has filament loaded, one checking it stays hidden when a slot is empty. I also ran the project's code-quality checks and everything came back clean. Beyond that, I ran it locally against my own H2C printer so I could actually watch it work with real AMS, AMS-HT, and external slot data, not just test data.


maziggy (changes requested):
Thanks for putting this together -- the feature itself is in the right place and the gating logic is consistent across all three slot types. Two things need to change before I can merge.

**1. The test file is in the wrong directory, and that is what fails CI.**

Frontend tests live in `frontend/src/__tests__/` (CONTRIBUTING.md, Testing section).
Your file sits in `frontend/src/pages/`, and all five TypeScript errors come from that one fact: `tsconfig.app.json` includes `src` but excludes `src/__tests__`, so `tsc -b` (the first half of `npm run build`) type-checks any test file outside that folder. There the relative imports do not resolve and the jest-dom matcher types are not loaded, hence both the "Cannot find module" and the "toBeInTheDocument" errors.

Worth knowing: where the file currently is, vitest collects zero tests from it --

    npx vitest run src/pages/PrintersPageKProfileDisplay.test.tsx
    -> (0 test), Failed to resolve import "../../pages/PrintersPage"

The imports are already written for the `__tests__/pages/` location, so moving the file is the whole fix: I copied it there unchanged and both tests pass. Please also re-check the claim in the PR description that the checks came back clean -- with the file in its current place they cannot have.

**2. A loaded slot with no K data now shows a fabricated "K 0.020".**

`formatKValue()` is `(k ?? 0.020).toFixed(3)`, and `AMSTray.k` really is nullable: the backend only fills it from `cali_idx` when the printer's profile map has that index, and nulls it when a slot is cleared. Gating on `filamentData` excludes empty slots, but not a loaded slot whose K is simply unknown -- which is common on X1C, where the value comes through the cali_idx lookup.

I ran your component against a loaded PETG slot with `k` missing, and again with `k: null`. Both render "K 0.020". Behind a hover, next to a "K Factor" caption, that default was harmless; as a permanent line on the card it reads as the slot's measured pressure advance. Please gate on `tray.k != null` instead (or have `formatKValue` return null and skip the row). Your own comment makes exactly this argument for empty slots -- it just needs to cover this case too.

**Two smaller things:**

- The label is hardcoded as "K", while the hover card right above it uses   `t('ams.kFactor')`. A German user sees "K 0.024" on the card and "K-Faktor 0.024"
  in the popup. Reusing the key, or adding a short one, keeps those in step.
- The tests only cover the standard AMS slot; the AMS-HT and external/dual-nozzle
  edits are untested. A loaded-slot-with-null-K case would be the valuable one to add, since it is what catches point 2.

For the record, what I checked and found fine: the gating condition matches across all three sites, the slot row is a CSS grid with stretch so a taller loaded slot does not misalign its empty neighbours, `--pc-t8` scales with card size like the existing label, and eslint is clean on both files. No CHANGELOG entry needed from you.

gyrene2083:
Just pushed a fix for everything you flagged, thank you for the detailed review, it was really helpful.

I moved the test file's to `__tests__/pages/` now, so that should clear up the CI failures. And you were completely right about the fabricated K value — I gated it on tray.k != null instead of just filamentData, so a loaded slot with no reported K stays blank instead of showing that misleading 0.020 default. I actually reverted the fix locally first and confirmed my new test failed without it, just to make sure the test was really catching the bug and not passing by accident.

Also swapped the hardcoded "K" for t('ams.kFactor') so it matches the hover popup across languages, and added test coverage for the AMS-HT and external/dual-nozzle cases plus the null-K scenario specifically, since that's the one that would've caught this originally.

Let me know if anything else needs adjusting.

maziggy (changes requested):
Thanks - all four points from the first round are properly fixed, and I checked each one rather than taking the summary's word for it. The test file is in `__tests__/pages/` now and `tsc -b` exits 0 with all five tests collected (it was collecting zero before). The null-K gate is right at all three sites, including the external/dual-nozzle one where you used `!isEmpty && extTray.k != null` instead - that variation is correct, not a slip, since `extFilamentData` is built unconditionally there and `filamentData &&` would have been a no-op. And your null-K test does fail without the gate: with `filamentData` alone the row renders "K Factor 0.020" and the `queryByText(/^K Factor/)` assertion trips. Good instinct reverting the fix to confirm that.

One new thing, and it falls out of the label fix.

**The localized label eats the value at narrow slot widths.** The slot grid is `minmax(3.5rem, 1fr)` (`PrintersPage.tsx:5151` and `:5163`) and the new line is 8px `font-mono` with `truncate`. Rendered in Chromium at the real widths:

```
slot width   en             de             fr             ja          old "K"
56px (floor) K Factor…      K-Faktor…      Facteur …      K値 0.024   K 0.024
70px         K Factor 0.…   K-Faktor 0.…   Facteur K 0…   ok          ok
88px         ok             ok             ok             ok          ok
```

A four-slot AMS needs roughly 350px of card width before the number survives in en/de/fr, so on a phone, in the dense fleet view, or in a narrow window, the label survives and the value - the entire point of the feature - is what gets clipped. Japanese is fine because `K値` is short.

My first-round note offered "reusing the key, or adding a short one". The measurement says take the short one: an `ams.kFactorShort` key ('K' in most locales, `K値` unchanged for ja) on the card, with `title={t('ams.kFactor')}` on the div so the full localized name is still there on hover. That keeps the languages in step without spending the width on the label.

One question rather than a request: `k != null` lets a firmware-reported `0` through as "K Factor 0.000". The backend's own profile map already treats zero as absent - `if kp.slot_id is not None and kp.k_value` (`printer_manager.py:1305`) - so if any firmware sends `"k": 0` for an uncalibrated slot, the misleading-reading problem comes back in a different shape and a truthy check would close it. If 0 is a legitimate report on some model, leave it. You have the captures for that; I don't.

Minor: the value now appears both on the card and in the hover card directly above it. I assume that is deliberate.

gyrene2083:
Thanks for the review, I pushed another round of fixes.

For the width/truncation issue - went with your second suggestion and added a short label (kFactorShort) instead of reusing the full one. The full localized name is still there, just as a hover tooltip now via the title attribute. Updated all 13 locale files so it's not just hardcoded English.

On the k != null question - you were right, switched it to a truthy check on tray.k across all three spots. Traced it through printer_manager.py and confirmed the backend already treats a K-profile of exactly 0 as "no calibration" when building the lookup map, so this just matches that existing behavior instead of letting a real 0 slip through as a fake reading.

Also found something on my own while testing this across a few browsers - Safari was truncating the value even at normal widths, Chrome/Firefox were fine. Turned out to be font-mono resolving to different actual fonts per browser (SF Mono in Safari vs Menlo in Chromium), so the same text took up different amounts of space depending on browser. Switched to tabular-nums instead, which is already used elsewhere in this file for the same kind of tight numeric label, and it renders the same everywhere since it's not pulling in a separate font family. Confirmed working in Brave, Firefox, and Safari now.

Added tests for the k:0 case and the short-label/title behavior too, so we're at 6 tests now.

Let me know if anything's off.

maziggy (approved):

Issue #2816, opened 2026-08-11
[Feature]: Improve File Manager - Configurable Thumbnail rendering, rotation and conversion


NaegeliJ:
### Problem or Use Case

I imported some STL Files from my Local Explorer and all the STL show up in Green Color and are Barley recognisable: 

<img width="191" height="197" alt="Image" src="https://github.com/user-attachments/assets/e5b668c0-f951-462d-8812-f2fef44e4725" />

This makes the Files look generic and are hard to Recognise. There is no way to adjust any of the Configuration like the Thumbnail Color and Rotation.

### Proposed Solution

I would propose these Improvments for File Handeling:

1. Make the Render Colour Configurable to allow for a Custom Default Option for STL
2. Rotate the STL to allow for a Spezific Angle to be shown in the Preview
3. STL to 3MF Conversion (via the existing slicer sidecar)

### Alternatives Considered

Workaround today: open the STL in a desktop slicer and export 3MF manually

Alternatives:
Organising my Files with another 3D Assest Manager like [Manyfold](https://github.com/manyfold3d/manyfold) maybe Integrate Extensions to allready Popular File Manager?

### Feature Category

File Manager

### Priority

Would improve my workflow

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Before we commit to building this, I'd like to gauge community interest. If you'd find this feature useful, please give this issue a thumbs up (👍) reaction so we can prioritize accordingly.

sadontsev:
I hit the same thing and went looking for the cause before reaching for configuration. It turns out to be neither the colour nor the angle: **the render is unlit.**

Both thumbnail renderers build the mesh the same way:

- `backend/app/services/stl_thumbnail.py:159`
- `backend/app/services/plate_thumbnail.py:182`

```python
Poly3DCollection(
    triangles,
    facecolors=BAMBU_GREEN,
    edgecolors=BAMBU_GREEN,
    linewidths=0.1,
    alpha=0.9,
)
```

`Poly3DCollection` defaults to `shade=False`, so every triangle gets exactly the same colour regardless of which way it faces. A solid silhouette is the guaranteed output rather than a bad-luck one — and because `edgecolors` is the same green as `facecolors`, the wireframe adds no relief either.

[...]

PR #2861, opened 2026-08-17
Light the generated thumbnails so one model differs from another (#2816)

sadontsev:
## Description

Both thumbnail renderers built their mesh with no light source, and `Poly3DCollection` defaults to `shade=False`, so every triangle was filled with the identical `BAMBU_GREEN` regardless of its normal. A flat silhouette was the guaranteed output — a model's own outline, with nothing to tell it apart from any other file. Because `edgecolors` is the same green as `facecolors`, the wireframe added no relief either.

This passes `shade=True` and a shared `LightSource` in both renderers. Same palette, same camera, same framing, same output sizes — only the surface picks up relief.

## Related Issue

Refs #2816

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#71 (updated alongside the review fixes)

**Pick one**:
- [x] Docs PR(s) linked above

The wiki's File Manager page documents the renderer in a *Technical Details* table; that table said nothing about lighting, which was accurate while the renders were unlit. The companion PR adds it.

No screenshots needed refreshing, and I checked rather than assumed: `file-manager.md` embeds no images at all, and `assets/archives.png` (used by `archiving.md` and `features/index.md`) shows archive cards whose thumbnails are all desktop-Studio renders — already grey and shaded — so the backend change does not affect it. The table's `Color` row also stays correct: shading modulates the Bambu green rather than replacing it.

## Type of Change

## Changes Made

- `backend/app/services/stl_thumbnail.py` — add `shade=True` + `lightsource=` to the `Poly3DCollection`, with `LIGHT_AZIMUTH_DEG` / `LIGHT_ALTITUDE_DEG` module constants alongside the existing palette constants.
- `backend/app/services/plate_thumbnail.py` — same, mirroring `stl_thumbnail.py`'s constants as that module already does for the palette. The `LightSource` is built once in `_render_model_thumbnails` and passed into `_render_at_size`, following the existing pattern of injecting the lazily-imported matplotlib objects, so the 512px and 128px renders are lit identically.
- Tests for both renderers asserting the output is actually shaded.

No dependency change: `shade`/`lightsource` arrived in matplotlib 3.7 and `requirements.txt` already pins `matplotlib>=3.8.0`.

## Screenshots

Same STL, rendered through `generate_stl_thumbnail` on this branch vs. with only the two service files reverted.

| Before | After |
|---|---|
| before | after |

The ribbing, the stepped profile and the recess on top only become visible once the mesh is lit.

## Testing

Local run (Python 3.14, matplotlib 3.11.1, trimesh 5.0.0):

```
ruff check backend/          All checks passed!
ruff format --check backend/ 788 files already formatted
pytest backend/tests/unit/         7552 passed
pytest backend/tests/integration/  2850 passed
pre-commit run --files <the four changed files>   all hooks passed
```

**The new tests fail without the fix**, which is the point of them — reverting just the two service files and keeping the tests gives:

```
FAILED test_stl_thumbnail.py::TestStlThumbnailService::test_generated_thumbnail_is_shaded_not_flat
FAILED test_stl_thumbnail.py::TestStlThumbnailConstants::test_light_direction_is_a_valid_direction
FAILED test_plate_thumbnail.py::TestInjectPlateThumbnails::test_injected_thumbnail_is_shaded_not_flat
E       AssertionError: assert 1 >= 3
```

The assertion counts distinct colours covering the model's surface, quantised to 5 bits per channel so PNG dithering doesn't scatter a flat fill into false variety. An unlit cube renders as exactly **1** tone; lit, its three visible faces give **3**.

## Checklist

## Review response

All four points addressed, plus two the review did not reach. Every fix has a test that fails without it, verified by reverting each in isolation.

**The light was behind the model.** Confirmed exactly as you described: dot product -0.34, both visible faces on the identical 0.475. Moved to 315. Two things worth adding — the comment claiming "front-left, toward the camera" described a light the code did not have, and the render still *improved* at 225 because the top face is bright at any azimuth, which is why a screenshot did not catch it. The companion wiki PR said "front-left" too and now describes the light by its effect instead.

**Two things your four points would have left in place**, both found while fixing:

`fix_winding` alone is not enough. It makes the winding agree but may settle on either orientation, and on a half-inverted icosphere it picks inward — volume -519, and all 1280 normals opposite the correct mesh. Consistent, and consistently lit from inside. `fix_inversion` after it flips on the sign of the volume and needs no scipy either. The winding test caught it as a mean render delta of 3.94, which is how I found it.

`shade=True` crashed on meshes with no shadeable face — a regression the shading itself introduced. `_shade_colors` returns the colour argument unchanged when every normal is degenerate, and for a colour *string* that is a 0-d array, so `to_rgba_array` raises `len() of unsized object`. Zero-area facets, collinear facets and a 3MF with an empty `<triangles/>` all rendered flat before and failed after. `batch_generate_stl_thumbnails` walks a whole folder with no `MIN_USABLE_STL_BYTES` pre-skip, so each would have counted as a failure in the UI and put a traceback in the log — the noise the demoted logging in `stl_thumbnail.py` exists to prevent. `_shade_kwargs` decides shadeability up front instead, ~6 ms on 227k faces.

**On the test.** You were right that `>= 3` was coupled to alpha. Fixing the azimuth fixes it, and I also corrected the "PNG dithering" claim — PNG is lossless, the spread is Agg antialiasing plus alpha compositing, as you said. Worth recording in the helper: the count measures large flat tone regions, which only coincides with "is it shaded" for a flat-faced model. Unshaded at alpha=0.9 a cube gives 1 and a cylinder 1, but a sphere and a torus already give 3 — so the cube fixture is load-bearing, and that is now written down.

**On the invariant.** My first attempt asserted only that the light was not behind the camera, and that is not sufficient: `az=45` — "put the light where the camera is" — scores the *highest* dot product of any azimuth (+0.94) and lights both visible faces identically. It now also requires the two visible faces to differ, so both failure modes are held.

**Perf.** `vertices_scaled[faces]` as you suggested. End to end on 82k faces: 0.592s before any of this, 0.698s with shading on the list form, **0.327s now** — 45% faster than the original unlit code while doing strictly more work.

**Also done:** helper moved to the existing `conftest.py` as a fixture, off the deprecated `Image.getdata()` (removed in Pillow 14, and `pyproject` silences DeprecationWarning so it would have surfaced as an `AttributeError` in CI), and `Fixes` changed to `Refs` above — #2816 asks for three things this PR does not implement, so merging should not close it.

models

Local on the rebased branch: ruff clean, **7603 unit** and **2850 integration** passing.

## Additional Notes

One trap is worth recording, and there is a comment at both call sites: `shade=True` raises inside matplotlib if `edgecolors="none"` —

```
ValueError: operands could not be broadcast together with shapes (227412,1) (0,4)
```

— because it shades the edge colours alongside the faces and the empty array fails to broadcast. The current code passes a real `edgecolors`, so this is safe exactly as written; it would only bite a later change that drops the edge colour while keeping the shading.

I kept the change deliberately narrow. #2816 also asks for configurable colour and rotation and STL→3MF conversion; those are separate and I have not touched them. My reasoning for doing this part first is that a configurable colour would produce a solid *blue* silhouette — what is missing from these renders is shape information, not hue.


maziggy (changes requested):
Thanks for this - the diagnosis is right, the matplotlib usage is right, and the tests are real ones rather than decoration. I ran the branch: ruff clean, full backend suite 10469 passed. Four points, the first two being the same root cause.

**1. The light is behind the model.** `LightSource(azdeg=225, altdeg=45).direction` is `[-0.5, -0.5, +0.707]`; the camera at `view_init(elev=25, azim=45)` sits at `[+0.64, +0.64, +0.42]`. Negative dot product. The resulting shade factors:

```
+x, +y  (the two faces you can see)   0.475, 0.475
+z top                                0.898
-x, -y  (hidden)                      0.825
```

So the two visible side faces are lit identically - the front vertical edge of a cube has no contrast - and the brightest surfaces in the scene are the ones facing away. The comment ("Front-left ... keeps the lit face toward the default view_init camera") and the wiki PR ("Directional light from the front-left") describe a light the code does not have.

`azdeg=315` gives `[-0.5, +0.5, +0.707]`, dot with the camera +0.30, and lights left 0.825 / right 0.475 / top 0.898. One constant, and both sentences become true as written. `135/60` is the mirror and works equally well.

**2. The shading test passes for a different reason than it claims.** The three tones a lit cube actually produces here are:

```
(0, 87, 33)  35%   side face over a back side face
(0, 81, 30)  29%   side face over the bottom face
(0,153, 58)  27%   top face
```

The first two are the same face colour, separated only by what bleeds through `alpha=0.9`. Re-render at `alpha=1.0` and the count drops to 2, so `>= 3` fails on a correctly shaded image - the assertion is coupled to the alpha value, not to the shading. Fixing point 1 fixes this too: at 315/45 the cube gives 3 tones at alpha 1.0 and 4 at 0.9, for the reason the docstring states. (Also, PNG is lossless - there is no dithering; the spread comes from Agg antialiasing and the alpha compositing. Quantising is still the right call.)

**3. STLs with inconsistent winding now render as camouflage.** matplotlib derives normals from vertex order, so inverted triangles shade backwards, and `trimesh.load(..., force="mesh")` does not repair winding (`is_winding_consistent` is still False after a round-trip). Such a file used to render flat but clean; now it renders broken, and the File Manager takes arbitrary user STLs. A gate would cover it:

```python
if not mesh.is_winding_consistent:
    trimesh.repair.fix_winding(mesh)
```

Measured on 82k faces: the check costs 28 ms, the repair 3.55 s and only runs on meshes that would otherwise look wrong. Use `trimesh.repair.fix_winding`, not `mesh.fix_normals()` - the latter reaches `body_count` -> `scipy.csgraph` and scipy is not a dependency; `fix_winding` goes through networkx, which is already pinned.

**4. Perf, optional.** 82k faces: 0.46 s unshaded -> 0.67 s shaded, nearly all of it `_generate_normals` walking the list-of-lists (0.205 s; the same call on an ndarray is 0.006 s). `poly3d = vertices_scaled[faces]` is the identical data, accepted by `Poly3DCollection`, and erases that cost - and speeds up the existing path. `plate_thumbnail` pays it twice per plate, once per size.

Minor: `_distinct_surface_tones` is duplicated verbatim in both test files and `backend/tests/unit/services/conftest.py` already exists; and `test_light_direction_is_a_valid_direction` only fails today because the constant would not exist. The gap worth covering instead is that the two modules agree - `plate_thumbnail._LIGHT_AZIMUTH_DEG == stl_thumbnail.LIGHT_AZIMUTH_DEG` - since "plate card and library thumbnail look alike" is the justification for the copy, and drift between them is what would break it silently. Importing them the way `_configure_matplotlib_cache` already is would do the same job.

sadontsev:
Thanks — all four are right, and each reproduced exactly as you set it out. Pushed as `ace7189a`, rebased onto your `dev` merge.

Your numbers matched to the digit, so I will not restate them. What is worth your time is **two things the four points would have left in place**, both found while fixing:

**1. `fix_winding` alone is not enough.** It makes the winding agree but is free to settle on either orientation, and on a half-inverted icosphere it picks *inward*: volume −519.09, and all 1280 face normals opposite the correct mesh. Consistent, and consistently lit from inside. My winding test caught it as a mean render delta of 3.94 against the correctly-wound render. `trimesh.repair.fix_inversion` after it flips on the sign of the volume, and needs no scipy either.

**2. `shade=True` crashes on a mesh with no shadeable face** — a regression the shading itself introduced, and the one I would most have regretted shipping. `_shade_colors` returns the colour argument unchanged when every normal is degenerate, and for a colour *string* that is a 0-d `<U7` array, so `to_rgba_array` calls `len()` on it:

```
TypeError: len() of unsized object
```

Zero-area facets, collinear facets, and a legal 3MF with an empty `<triangles/>` all rendered flat before and failed outright after:

| input | pre-PR | first PR | now |
|---|---|---|---|
| zero-area triangle | PNG 948 B | `TypeError` | PNG 948 B |
| collinear triangle | PNG 1670 B | `TypeError` | PNG 983 B |
| 50 zero-area facets | PNG 948 B | `TypeError` | PNG 948 B |

The part that decided it for me is the blast radius: `batch_generate_stl_thumbnails` walks a whole folder and, unlike the other call sites, has no `MIN_USABLE_STL_BYTES` pre-skip. Every stub STL would have counted as a failure in the UI *and* put a traceback in the journal — exactly the noise the demoted logging at `stl_thumbnail.py:117-124` was added to prevent. `_shade_kwargs` decides shadeability up front rather than catching the TypeError, so a flat render stays a real outcome instead of an error path. ~6 ms on 227k faces.

**On your point 1**, two footnotes. The comment claiming "front-left … toward the camera" described a light the code did not have — I wrote the intent and never checked the geometry. And the render still looked better at 225, because the top face is bright at any azimuth; that is why the before/after screenshot did not catch it. The companion wiki PR said "front-left" too and now describes the light by its effect.

**On the invariant I added for it** — my first attempt asserted only `light · camera > 0`, and that is *not* sufficient. `az=45` — "put the light where the camera is", the most natural next edit — scores the **highest** dot product of any azimuth (+0.94) and lights both visible faces identically. The test now also requires the two visible faces to differ, so both failure modes are held:

| reverted to | assertion that fires |
|---|---|
| `az=225` | `assert -0.342 > 0` — "the light is behind the model" |
| `az=45` | `assert 0.0 > 0.05` — "both visible faces are lit the same" |

**On your point 2**, agreed and corrected, including the PNG-dithering claim — the spread is Agg antialiasing plus the alpha compositing. One thing worth recording in the helper: it counts large flat tone regions, which only coincides with "is it shaded" for a flat-faced model. Unshaded at alpha=0.9, a cube gives 1 and a cylinder 1, but a **sphere and a torus already give 3** — so the cube fixture is load-bearing, and that is now in the docstring rather than left to luck.

**On your point 4**, measured end to end through `generate_stl_thumbnail` on 82k faces: 0.592s before any of this, 0.698s with shading on the list form, **0.327s now**. Faster than the original unlit code while doing strictly more work.

Minor items all done: helper moved to the existing `conftest.py` as a fixture, the constants imported rather than mirrored so drift is impossible instead of merely tested for, off the deprecated `Image.getdata()`, and `VIEW_ELEV_DEG`/`VIEW_AZIM_DEG` named so the test asserts against the real camera rather than a hardcoded 25/45 that could silently diverge.

I also changed `Fixes #2816` to `Refs #2816` — #2816 asks for configurable colour, rotation and STL→3MF, none of which this implements, and `Fixes` would have closed all three on merge. Happy to move this under its own issue if you would rather.

models

ruff clean; **7603 unit** and **2850 integration** passing locally on the rebased branch.

sadontsev:
@maziggy appreciate re-review here. 
Pinging only in case it was missed by accident, since it's been a few days. 

Thank you.

sadontsev:
Perhaps the only issue I spotted so far with this light is on cases when there is a smaller object after a larger object, so it ends up in a shadow. 

Maybe it also worth considering either placing an individual light for each object or just placing light from every direction. 

Also, subjectively, lighter colours enable better contrast, details. 

image

maziggy:
I'm working on issues/PRs in the order they came in.

sadontsev:
> I'm working on issues/PRs in the order they came in.

Noted, sorry. 
Won't bother again

maziggy (approved):

maziggy:
Re-reviewed. All six hold, and I mutation-tested rather than trusting the green:
reverting the azimuth to 225 fails the dot assert, setting it to 45 fails the
face-difference assert, forcing the winding gate false fails the winding test,
and removing the degenerate guard reproduces the TypeError verbatim. Your
_shade_kwargs normal formula is identical to _generate_normals on the
ndarray-of-triangles path, and the finite-and-nonzero test is the exact
complement of mask.any() in _shade_colors. The tone test is no longer coupled to
alpha either: at 1.0 the cube gives 3 shaded against 1 unshaded.

One thing the round left open, and I have pushed the fix rather than sending it
back. fix_inversion only acts on a watertight mesh -- it returns early
otherwise, since a volume measured across holes says nothing about which way is
out. Your fixture is watertight, so it passes. Punch holes in the same icosphere
and fix_winding settles inward unopposed: 0 of 1200 faces oriented like the
correct mesh, mean render delta 4.56 against your own threshold of 1.0. That
shape -- inconsistent winding and not watertight -- is the ordinary broken STL.
With no usable volume the orientation now comes from whether the faces point
away from the centroid: 1200 of 1200, delta 0.00, numpy only. The test is
parametrised over both topologies. I also folded the repair into one shared
_repair_winding, the way you did for the light and the shade kwargs, and moved
the plate_thumbnail call above the vertex read -- scaled was being indexed with
post-repair faces, which is fine today and stops being fine the moment a repair
moves a vertex.

On the shadow: it is not one. _shade_colors works per face from the normal with
no occlusion, and it normalises on a fixed -1..1 into 0.3..1 scale rather than a
scene-relative one, so a large object cannot darken a small one. I rendered a
big box with a small box behind it and the tone bands are the same as either
rendered alone, around 80, 136-143 and 152-156, with no extra dark band. What
you are seeing is a part turned so the 0.475 faces are the ones you see, and
per-object lights would not change it.

Your instinct about contrast is right, though, just at a different knob. 315/45
fixed the side-to-side collapse but moved the problem to the top edge: the lit
side at 0.825 and the top at 0.897 are 0.072 apart. Sweeping azimuth and
altitude over the three visible faces, 315/45 ranks 469th of 890 valid positions
by worst-pair separation, where the best is about 0.237 -- 285/55 gives 0.456 /
0.702 / 0.937. I am still looking at renders before changing it, so the constant
stays as you have it for now.

Refs #2816 rather than Fixes is the right call: colour, rotation and STL to 3MF
are all still unbuilt, so the issue stays open.

Issue #1473, opened 2026-05-21
[Bug]: `print_missing_spool_assignment` notification fires on every Spoolman-mode print (only checks legacy `spool_assignment` table, ignores `spoolman_slot_assignments`)


ojimpo:
## Bug description

`print_missing_spool_assignment` notification fires on every print start, even when all used trays are correctly bound to Spoolman spools via `spoolman_slot_assignments`.

## Bambuddy version

v0.2.4.2 (current latest stable, as of 2026-05-21)

## Setup

- Spoolman mode enabled (bindings stored in `spoolman_slot_assignments`, not legacy `spool_assignment`)
- AMS 2 Pro, 4 trays loaded with Bambu RFID-tagged spools, all bound via Bambuddy "Assign Spool" UI
- Confirmed `spoolman_slot_assignments` populated:
  ```
  id=6  printer_id=1  ams_id=0    tray_id=2  spoolman_spool_id=1
  id=7  printer_id=1  ams_id=0    tray_id=1  spoolman_spool_id=10
  id=8  printer_id=1  ams_id=0    tray_id=0  spoolman_spool_id=3
  id=2  printer_id=1  ams_id=0    tray_id=3  spoolman_spool_id=2
  id=5  printer_id=1  ams_id=255  tray_id=0  spoolman_spool_id=9
  ```
- Legacy `spool_assignment` table: 0 rows
- 13 false-positive notifications over the past 7 days, every print start

## Symptom

```
Slots: A2
Expected profile:
- A2: PLA Basic
```

A2 (tray 1) is bound to spoolman_spool_id=10, yet flagged as missing every print.

## Root cause

`backend/app/services/spool_assignment_notifications.py:notify_missing_spool_assignments_on_print_start` queries only the legacy `SpoolAssignment` table:

```python
from backend.app.models.spool_assignment import SpoolAssignment
...
assignments_result = await db.execute(
    SpoolAssignment.__table__.select().where(SpoolAssignment.printer_id == printer_id)
)
assignments = assignments_result.fetchall()
assigned_global_trays = {
    _global_tray_from_assignment(assignment.ams_id, assignment.tray_id) for assignment in assignments
}
missing_global = sorted(used_global_trays - assigned_global_trays)
```

[...]

PR #2862, opened 2026-08-17
Feat/pause print unassigned spool

### LINKED ISSUE #950: [Feature]: Assign filament to archived print
### Problem or Use Case

I'm always frustated when I forgot to assign the current filament spool in bambuddy, I do not have an AMS so I have to assign them manually. 

### Proposed Solution

It would be great if you can assign a filament spool after printing. Similar to the project assignements in the archive page.

### Alternatives Considered

_No response_

### Feature Category

Print Archiving

### Priority

Would improve my workflow

### Mockups or Examples

_No response_

### Contribution

### Checklist

- [] I have searched existing issues to ensure this feature hasn't already been requested

firke:
## Description

Adds an opt-in setting that pauses a print at start when a tray the job uses has no spool assigned, so filament consumption isn't silently left untracked.

Today, reloading filament in the AMS makes `on_ams_change` delete the `spool_assignment` row. If a print then runs, every deduction site in `usage_tracker` skips that tray with a `continue` — no spool debited, no `spool_usage_history` row, inventory drifts silently. The condition was already *detected* at print start (a toast + push fired), but it was fire-and-forget and the print ran anyway. This escalates that existing check from "notify" to "pause" so you can assign the spool and resume. **Off by default.**

The existing service was extended rather than duplicated; its mapping resolution and both-tables assignment union (per #1473) are unchanged.

**Resume-and-count needs no changes to `usage_tracker.py`**, which is what makes pause the right action: deduction happens only at *completion* from the 3MF estimate, `_resolve_spool_id_for_tray()` falls through to the live assignment row when the print-start snapshot has no entry, and resume from PAUSE doesn't re-fire `on_print_start` (the `_was_running` guard) — so the session survives and it's one-shot-per-print for free. Resuming without assigning is treated as a deliberate override.

## Related Issue

Fixes #950

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#70

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition or update

## Changes Made

- **New setting `pause_print_on_unassigned_spool`** (default off) in Settings → Filament → "Filament checks". KV-stored, so no migration.
- **`check_spool_assignments_on_print_start`** (renamed) now calls `client.pause_print()` when enabled. Fails open — any error or unresolvable AMS mapping leaves the print untouched, and pause errors are swallowed separately so they can't suppress the warning notification.
- **`missing_spool_assignment` WS event gained a `paused` flag** — error-severity toast with dedicated copy, `paused` folded into the dedup signature.
- **New notification event `print_paused_unassigned_spool`** with its own template, provider toggle and DB column. Fires *instead of* the warn-only event when paused, never both. Defaults **`TRUE`** (unlike its sibling's `0`) so existing providers don't get a silently halted printer — same reasoning as `on_plate_not_empty`. The two methods now share `_notify_unassigned_spool()`.
- **Log viewer**: labels + colors for the new event and its pre-existing sibling, which rendered as raw snake_case.
- i18n across all 13 locales; tests for the migration (default backfill, idempotency, PG dialect guard), the event split, and a `usage_tracker` regression pinning the empty-snapshot → live-assignment fallback.

## Testing

- [x] I have tested with my printer model: BambuLab X2D

Automated (post-merge with upstream `main`): backend 10180 passed · frontend 3062 passed · `tsc` clean · i18n parity 13 locales · `ruff check`/`format` clean.

## Checklist


maziggy:

PR #2866, opened 2026-08-17
Serve a library file's geometry as STL so clients can show it in 3D

sadontsev:
## Description

Adds `GET /api/v1/library/files/{file_id}/mesh` — returns a library file's geometry as a binary STL.

Bambuddy can already turn a model into a *picture*: `stl_thumbnail` and `plate_thumbnail` both load a mesh with trimesh and render it server-side. What it cannot do is hand the geometry itself to a client, so any UI that wants to let a user orbit a model has to parse the container on its own. For a `.3mf` that means shipping a ZIP reader and a 3MF XML parser into every client, to reach a mesh the server can already read.

This serves that mesh, so a client needs no container handling at all.

## Related Issue

None — this adds an endpoint rather than fixing a reported issue.

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#72

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [x] Test addition or update

## Changes Made

- **`backend/app/services/mesh_export.py`** (new) — loads a model with trimesh, decimates it toward a vertex budget, exports binary STL. Raises `MeshExportError` carrying a sentence fit to show a user.
- **`backend/app/api/routes/library.py`** (+70) — the route. Auth and visibility copy the adjacent `download_file` exactly: `require_ownership_permission(LIBRARY_READ_ALL, LIBRARY_READ_OWN)` plus `_ensure_library_file_visible`.
- **`backend/tests/unit/services/test_mesh_export.py`** (new) — 15 unit tests.

### Why STL

Rather than a JSON vertex payload: roughly half the bytes, every 3D toolkit reads it, and it makes the response the same shape as the `.stl` files already in the library — so a caller has one code path instead of two.

### What it refuses, and why

**Sliced `.gcode.3mf` → 400.** Those carry toolpaths, which the G-code viewer shows properly. They are also unreadable here: trimesh's 3MF loader raises `KeyError: 'world'` on Bambu Studio's sliced output, whose scene graph has no root node the loader expects. Refusing on the filename means that arrives as a specific status rather than a confusing 422 after a wasted parse.

`400` rather than `415`: `415` conventionally describes the *request* body's content type, not the type of the resource being asked about, and it is absent from the wiki's documented status codes. Happy to change it.

**No geometry / unreadable / too small → 422** with a sentence fit to show a user. Parser internals are logged, not forwarded.

### The vertex budget is deliberately half the thumbnail service's

`stl_thumbnail.MAX_VERTICES` is 100k and bounds a render that happens once, server-side, and becomes a 256 px PNG. This bounds bytes that cross a network and get parsed by a client — a different problem.

STL is an unavoidably wasteful transport: 50 bytes per triangle, every vertex duplicated three times, so the export is *larger* than the container it came from. Measured on a real 3.9 MB project 3MF:

| vertex budget | STL | gzipped |
|---|---|---|
| 100k | 10.0 MB | 4.8 MB |
| 50k (this PR) | 5.0 MB | — |

There is no `GZipMiddleware` in `main.py`, so a client fetches the uncompressed size. 50k keeps that near five megabytes, and the detail lost at that density is not visible while orbiting.

### Off the event loop

Parsing and decimating runs in `asyncio.to_thread`, matching how `system.py` and `support.py` offload blocking work. Not theoretical: one real 3.1 MB model took **10.2 s** to decimate, which would otherwise have stalled every other request.

## Screenshots

Not applicable — no UI change.

## Testing

- [ ] I have tested with my printer model: not applicable — this endpoint reads a stored file and never touches a printer.

`./test_backend.sh` — ruff lint, ruff format, and the full suite: **10,084 passed**.

15 unit tests in `backend/tests/unit/services/test_mesh_export.py`, following the existing `test_stl_thumbnail.py` conventions (including its `trimesh`-available skip guard):

- the filename rules, including that `.gcode.3mf` is refused **even though it ends in `.3mf`**
- every refusal path — wrong type, missing, too small, unparseable, a ZIP with no model part
- a binary-STL header check (80-byte header, `uint32` count, 50 bytes per triangle)
- a **round-trip**: the export is reloaded with trimesh and must come back as a mesh, because an output that only *looks* like an STL would defeat the point of the endpoint
- all three decimation fallbacks — reduction that throws, that empties the mesh, and that succeeds

### Verified against real library files

Run inside a running container against that instance's own library:

| file | source | STL | faces |
|---|---|---|---|
| plain `.3mf` | 770 KB | 80 KB | 1,596 |
| plain `.3mf` | 3.1 MB | 7.3 MB | 146,914 |
| plain `.3mf` | 3.9 MB | 5.0 MB | 99,988 |
| sliced `.gcode.3mf` | 6.1 MB | *refused* | — |

The middle one decimates to 147k faces rather than the ~100k the budget asks for — quadric decimation cannot always reach its target. That is why a simplification which misses, throws, or empties the mesh falls back to delivering what it has rather than refusing: the client can decide what to do with a dense mesh, but it can do nothing with a 422.

## Checklist

## Additional Notes

- `Cache-Control: private, max-age=86400` — the mesh is a pure function of an immutable stored file, and `private` because the bytes sit behind a per-user permission check.
- Happy to change the vertex budget, the media type (`model/stl`), or the status codes if you'd prefer different conventions.

### DIFF
--- a/backend/app/api/routes/library.py
+++ b/backend/app/api/routes/library.py
@@ -1,5 +1,6 @@
 """API routes for File Manager (Library) functionality."""
 
+import asyncio
 import base64
 import binascii
 import contextlib
@@ -70,6 +71,7 @@ from backend.app.services.design_settings import (
     overrides_from_config,
 )
 from backend.app.services.filament_requirements import annotate_rack_groups
+from backend.app.services.mesh_export import MeshExportError, export_mesh_stl, is_exportable
 from backend.app.services.plate_thumbnail import inject_plate_thumbnails_if_missing
 from backend.app.services.process_overrides import apply_process_overrides
 from backend.app.services.slice_output_check import missing_start_gcode_message, start_gcode_is_missing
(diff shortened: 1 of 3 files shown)


maziggy:

Issue #2603, opened 2026-07-18
Multi-plate queue cancellation loses selected plate in Print History and displays Plate 1


Jostxxl:
## Summary

When a print selected from a multi-plate 3MF is cancelled from the queue, the Print History entry does not retain the selected plate. The UI falls back to **Plate 1**, even when a different plate was queued and cancelled.

This makes it impossible to tell which plate must be added back to the queue.

## Environment

- Current `dev` build: `059b733c`
- Native systemd deployment with PostgreSQL
- Source file: `heart 3.gcode.3mf`
- 22 plates in the 3MF
- Large farm with 93 configured printers

## Reproduction

1. Add multiple plates from a 22-plate 3MF to the queue.
2. Let one plate get assigned/started or remain associated with an archive.
3. Cancel/stop that queue print.
4. Open Print History.
5. The cancelled/aborted print is shown as **Plate 1**, regardless of the selected plate.

## Concrete database evidence

The most recent affected item is queue item 177:

```
print_queue.id              = 177
print_queue.status          = cancelled
print_queue.plate_id        = 22
print_queue.archive_id      = 436
print_queue.library_file_id = 55
error_message               = Stopped by user (printer was offline)
```

The queue therefore retains the correct selected plate: **22**.

Archive 436 contains:

```
filename   = heart 3.gcode.3mf
print_name = heart 3
status     = printing
subtask_id = 1980678774
```

However, `print_archives` has no `plate_id` column, and its `extra_data` contains aggregated metadata for all 22 plates rather than the selected queue plate. The history UI consequently has no reliable selected-plate value and falls back to Plate 1.

Other affected cancelled rows show the same data shape:

```
queue 161 -> plate_id 6  -> archive 377
queue 180 -> plate_id 7  -> archive 376
queue 182 -> plate_id 7  -> archive 374
```

The source queue rows retain the correct plate, while their archives do not.

## Expected behavior

Print History should identify the actual selected plate, for example:

```
heart 3 — Plate 22 — Cancelled
```

The plate number must remain available after cancellation, failure, completion, restart reconciliation, and queue cleanup so the operator can requeue the exact plate.

## Suggested fix

Persist the selected plate when creating the archive from a queue item. Possible approaches:

1. Add a nullable `plate_id` column to `print_archives` and copy `PrintQueueItem.plate_id` into it.
2. At minimum, store a dedicated selected plate field such as `extra_data.selected_plate_id`, rather than relying on aggregated 3MF metadata.
[...]

PR #2871, opened 2026-08-18
Report the selected plate on the archives API (#2796)

### LINKED ISSUE #2796: [Bug]: archive_to_response omits plate_id, so the archives API always returns plate_id: null
**What happens**
`GET /api/v1/archives/` and `GET /api/v1/archives/{id}` always return `plate_id: null`, even for archives whose `plate_id` column is populated. On our instance 181 of 257 rows carry a non-null `plate_id` in the DB; the API reports `null` for every one of them.

**Why**
`archive_to_response()` in `backend/app/api/routes/archives.py` builds the response dict field by field and never sets `plate_id`. `ArchiveResponse.plate_id` defaults to `None` (`schemas/archive.py:58`), so Pydantic fills the null in silently — no error surfaces. Both the list and the detail endpoint go through this helper, and `extra_data` doesn't carry the plate either, so there is currently no way to read the selected plate back out of the API.

The model column exists (`models/archive.py:81`) and is populated. This looks like the tail end of #2603, which added the column and the schema field — the serializer just never got the line.

**Fix**
One line in `archive_to_response()`, e.g. beside `print_name` to match the schema order:
```python
"plate_id": archive.plate_id,
```

**Version**
Observed on `ghcr.io/maziggy/bambuddy:latest`; checked against current `main` as of 2026-08-08 — still present.

Happy to send a PR instead if you prefer.

sgiffhorn:
## Description

`archive_to_response()` builds the archives response dict field by field and left `plate_id` out, so `GET /api/v1/archives/` and `GET /api/v1/archives/{id}` returned `plate_id: null` for every archive. `ArchiveResponse.plate_id` defaults to `None`, so nothing raised an error. On the reporting instance 181 of 257 rows have the column populated.

Two things worth knowing before reviewing:

* The endpoints that return the ORM row were unaffected. `toggle_favorite` returns `archive` and the upload path calls `ArchiveResponse.model_validate`; with `from_attributes = True` both read the attribute. Only the paths through the hand-built dict (list, search, detail, PATCH, and the project archive list in `projects.py`) dropped it.
* There is a visible effect without a frontend change. `ArchivesPage.tsx:1095` already renders the `printers.plateNumber` label when `plate_id` is set, so it starts appearing on archives of multi-plate 3MFs.

## Related Issue

Fixes #2796

## Documentation

**Pick one**:
- [x] No docs update required, reason: the archives response sample in `docs/reference/api.md` is illustrative and already omits about 40 real `ArchiveResponse` fields. No documented behaviour, config key, endpoint or install step changes.

## Type of Change

## Changes Made

- `backend/app/api/routes/archives.py`: add `"plate_id": archive.plate_id` to `archive_to_response()`, next to `print_name` to match the field order in `ArchiveResponse`.
- `backend/tests/integration/test_archives_api.py`: regression test over the list and the detail endpoint, plus an archive without a plate that must read `None`, so a fix substituting a fallback plate fails instead of passing.

## Screenshots

No frontend file is touched and I did not run the UI, so I have no before/after to attach. The visible delta is the plate label appearing in the archive card title via the existing render path above. Happy to add a screenshot if you want one.

## Testing

- With the line removed the new test fails on `assert None == 22`, the symptom from the issue. Restoring it turns it green.
- `ruff check` and `ruff format --check` clean.
- Full backend suite the way `test_backend.sh` runs it (`pytest backend/tests/ -n 16`, `test_bambu_ftp.py` ignored): 10338 passed, no failures.
- Also ran `test_archive_plate_2603.py`, `test_archive_purge_api.py` and `test_projects_api.py`, whose archive list goes through the same helper.

- [ ] I have tested with my printer model: n/a, the change is a response serializer and touches no printer path.

## Checklist

## Additional Notes

Targeted at `dev` per CONTRIBUTING. The Co-Authored-By trailer is off the commit.


maziggy:
I don't have a problem with using AI for assistance - but to be honest - I don't want to have Claude Code commits in the repo.

sgiffhorn:
> I don't have a problem with using AI for assistance - but to be honest - I don't want to have Claude Code commits in the repo.

Understood, I have removed Claude as author, or do you want me to resubmit the entire commit manually? I looked into the CONTRIBUTING.md and did not find anything related to AI usage, so assumed it would be ok. You might want to specify it there.

maziggy (approved):
Thanks, this is a clean report and a clean PR -- merging it.

I checked the fix past the one line: plate_id was the only ArchiveResponse
field the hand-built dict was missing, the key order now matches the schema
exactly, and with the line taken back out your new test is the only one that
fails, on the assert None == 22 you describe. Full backend suite green here
as well.

One correction to the description. The label does not only appear on
multi-plate archives. The print dialog auto-selects the only plate of a
single-plate file and sends plate_id: 1 with the queue item, the scheduler
copies it onto the archive, and the #2603 startup backfill did the same for
older rows -- which is where most of your 181 of 257 come from. Left as it
was, the card title would have read "Plate 1" on most of the grid. I have
gated it on plate_id > 1 on dev; the plate carousel already gates on
is_multi_plate, which is what PrintersPage does with current_plate_id.

Nothing to change on your side.

Issue #2641, opened 2026-07-23
[Bug Report] Title: Server hangs after hours/days: SQLite pool max_overflow=200 exhausts file


maziggy:
Title: Server hangs after hours/days: SQLite pool max_overflow=200 exhausts file descriptors (aiosqlite thread-per-connection)

Summary — After a day or two, Bambuddy becomes unresponsive while still holding port 8000 — it doesn't exit, so launchd's KeepAlive never restarts it. The process pegs a CPU core and stops answering HTTP. Root cause is file-descriptor exhaustion from the SQLite engine's oversized connection pool.

Environment

Bambuddy v0.2.4.9 (+6 commits, 6f5d68ec)
macOS 13.7.8 (Intel), Python 3.12.11, uvicorn under launchd (com.bambuddy.server)
SQLite (WAL); launchd plist sets no SoftResourceLimits → default 256 open-file limit
Symptoms — HTTP times out but port is open; process alive at ~76–84% CPU; KeepAlive doesn't help because it's hung, not exited.

Diagnosis / evidence (at hang time)

258 open FDs (at the 256 limit); 120 of them pointed at one bambuddy.db.
faulthandler dump: many threads parked in aiosqlite/core.py:59 _connection_worker_thread (one thread per open connection).
Logs: repeated sqlite3.OperationalError: database is locked on frequent writes (ams_sensor_history, smart-plug energy snapshots) + [TIMING] lines of 16–21 s per DB op.
Root cause — backend/app/core/database.py (~L29):

if is_sqlite():
    kwargs = {"pool_size": 20, "max_overflow": 200}   # ← up to 220 connections
With aiosqlite, each pooled connection = its own OS thread + FD(s). SQLite serializes writers, so under contention operations back up and SQLAlchemy opens overflow connections (up to 220) → 220+ threads/FDs → past the 256-FD ceiling. Descriptors exhausted, the loop can't even write its self-pipe (asyncio … _write_to_self) → hot hang instead of a clean error. Large overflow is pointless for SQLite (single writer) — it only burns threads/FDs.

Suggested fixes

Cap the SQLite pool (pool_size=5, max_overflow=10, or NullPool/StaticPool).
Reduce write contention — batch the high-frequency history/energy inserts; ensure a busy_timeout PRAGMA.
Defense-in-depth: raise the process FD limit (launchd SoftResourceLimits NumberOfFiles ~4096).
Make exhaustion fail loud (health check → restart) instead of a silent hot-hang.
[...]

Issue #2572, opened 2026-07-15
[Bug]: PostgreSQL connection pool exhaustion with large printer farms


Jostxxl:
### Product / area

Bambuddy — PostgreSQL, printer management, authentication middleware

### Bug description

On a large printer farm, Bambuddy exhausts its SQLAlchemy PostgreSQL pool. The engine currently uses a hard-coded pool_size=10 plus max_overflow=20. With 93 connected printers, all 30 connections were repeatedly observed as idle in transaction. Normal API requests then wait for the 30-second pool timeout or fail in the authentication middleware.

The SQL statements complete quickly. Connections remain checked out while printer, archive, settings and library workflows continue, including slow MQTT, FTP, camera and 3MF work.

### Environment

- Bambuddy: v0.2.5b2-daily.20260714-3-ga31a3a52
- Native/systemd installation
- PostgreSQL 16
- 93 printers currently; approximately 130 planned
- Local SSD, 95 GiB RAM

### Measurements

- Unauthenticated /api/v1/printers probe: about 25.5 seconds before returning 401
- PostgreSQL: 30/30 sessions idle in transaction
- Oldest observed transaction: over 1 minute 40 seconds
- Repeated error: SQLAlchemy QueuePool limit of size 10 overflow 20 reached, timeout 30 seconds
- Follow-on error: auth_middleware failing closed on auth-probe error
- Common last queries read settings/auth_enabled, printers, print_archives and library_files.

### Reproduction

1. Configure PostgreSQL and add about 80–100 active printers.
2. Allow printer callbacks, archive processing, snapshots, FTP and queue polling to run.
3. Open printer or queue pages and inspect pg_stat_activity.
4. Observe all 30 connections becoming idle in transaction while requests wait for pool checkout.

### Temporary mitigation tested

A local runtime override set pool_size=100, max_overflow=200 and PostgreSQL max_connections=400. The same API probe then returned in about 0.025 seconds, and pool/auth-probe errors stopped.

PostgreSQL idle_in_transaction_session_timeout=15s is not safe here: it terminated connections Bambuddy still expected to use and caused asyncpg/SQLAlchemy rollback errors.

### Proposed fix

1. Support environment variables for PostgreSQL pool_size, max_overflow and pool_timeout.
2. Do not hold AsyncSession transactions across MQTT, FTP, camera, 3MF parsing or other network/file I/O. Load and detach data, close the session, perform external work, then use a short new session for final writes.
[...]

Issue #2544, opened 2026-07-12
[Bug]: Backups Failing with read only file system


pwostran:
### Product

Bambuddy

### Area

Other / not sure

### Bug Description

Backups are failing with "read only file system" with latest update to a mounted NAS drive.  Worked fine until 6/30.

### Expected Behavior

Backups complete

### Steps to Reproduce

1. Click backup
2. Backup runs

### Troubleshooting steps already taken

Checked folder permissions.

### Printer Model

P1S

### Bambuddy Version

2.4.9

### Installation Method

Manual (git clone)

### Operating System

Linux (Ubuntu/Debian)

### Support Package (.zip) and / or relevant logs

[bambuddy-support-20260712-002025.zip](https://github.com/user-attachments/files/29934390/bambuddy-support-20260712-002025.zip)

### Screenshots

<img width="1000" height="1145" alt="Image" src="https://github.com/user-attachments/assets/b0467311-79b5-4ea1-88c9-0c921905fba2" />

<img width="819" height="459" alt="Image" src="https://github.com/user-attachments/assets/d32347dd-ea9b-4498-a063-8db340f1d5ae" />

<img width="500" height="1054" alt="Image" src="https://github.com/user-attachments/assets/0e5daead-af38-43be-92da-421f08e412ff" />

### Additional Context

here is some of the fstab.

<img width="720" height="70" alt="Image" src="https://github.com/user-attachments/assets/680e0b25-b107-43ba-ae4c-65c40faae7a7" />

### Final checks

- [x] I searched existing (open AND closed) issues and this bug hasn't already been reported or resolved
- [x] I checked the [Troubleshooting wiki](https://wiki.bambuddy.cool/reference/troubleshooting/) and the relevant feature page for my issue
- [x] I am using the latest version of Bambuddy (or the latest daily build)
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled
- [x] For any connection / printing / camera issue, I ran the in-app Connection Diagnostic and included the result above

maziggy:
Thanks for the detailed report -- the screenshots and the fstab line were exactly what was needed.

[...]

PR #2884, opened 2026-08-20
fix(db): cap SQLite pool fd bill + raise container nofile — prevents fd-exhaustion DB corruption

### LINKED ISSUE #2883: [Bug] SQLite/WAL: default pool (20+200) can exhaust the container's 1024-fd limit → 'disk I/O error' → database corruption
### Summary

On a single-printer install, Bambuddy 1.2.5.3 (Docker, SQLite backend) exhausted its container's default **1024 soft `nofile` limit ~17.5h after start**. From that moment every new DB connection failed with `sqlite3.OperationalError: disk I/O error` (raised from `PRAGMA journal_mode = WAL` in the connect handler). The instance kept serving in this degraded state for ~44h, after which the database went **`database disk image is malformed`** — login broke (login `POST` returned 200 but every subsequent token/api-key `SELECT` failed), the WAL froze, and the DB needed a `.recover` salvage.

### Root cause (reproduced deterministically)

SQLite in **WAL mode** never really closes an overflow connection's main-db file descriptor while any other connection to the same file stays open. The unix VFS parks the closing fd on its **per-inode unused-fd list** (to dodge the POSIX close-drops-advisory-locks trap) and only releases the list when the **last** connection to the file closes — effectively never in a running server, because the SQLAlchemy pool always keeps connections alive.

Parked fds are *reused* by later connections, so the steady-state fd bill is bounded by the **historical peak of concurrent connections** — which the current SQLite defaults in `backend/app/core/database.py` set at `pool_size=20 + max_overflow=200` = up to ~220 parked db fds, plus wal/shm fds of live connections, plus normal socket load. Under Docker's default 1024 soft nofile that ceiling is reachable on an ordinary busy instance (active print + phone app + web UI was enough here).

Minimal repro (SQLAlchemy 2.0.52 + aiosqlite 0.22.1; identical on 2.0.51 / 0.21.0 — this is not a driver regression):

```python
# engine: sqlite+aiosqlite, pool_size=1, max_overflow=30,
# with Bambuddy's connect listener (PRAGMA journal_mode = WAL)
# fire 25 concurrent sessions, then count fds on the db file:
baseline fds: 1
after 25-concurrent burst: 26   # +25, one per overflow connection
after gc.collect():        26   # not a Python-level leak
after same-peak re-burst:  26   # plateau — parked fds are reused
after engine.dispose():    1    # released only when last connection closes
```

Bisect: the effect appears exactly when `PRAGMA journal_mode = WAL` is active; `busy_timeout`/`synchronous` alone are clean. `strace` on the production instance during a 60-parallel authenticated burst shows the overflow connections' `openat(".../bambuddy.db")` calls with **zero matching `close()`**, while their `-wal`/`-shm` fds close normally.

### Production evidence

- Fresh container: 6 db fds. After 90 min of ordinary use (active print, phone app, web UI): **92 db fds vs 20 wal fds** — 72 parked.
- `nofile` ceiling reached 17.5h after the Sunday image update restart; ~5,500 `disk I/O error` events over 44h; then `database disk image is malformed` on live pages (auth tables among the casualties → login outage).
- Related: #2641 reported the fd-exhaustion hang shape; this issue adds the parked-fd mechanism, a deterministic repro, and the corruption end-state.

### Suggested fix

PR: https://github.com/maziggy/bambuddy/pull/2884

1. Cap the SQLite pool defaults (`pool_size 20 → 10`, `max_overflow 200 → 20`; both stay env-overridable via `DB_POOL_SIZE` / `DB_MAX_OVERFLOW`). This bounds the parked-fd bill at ~30. Installs needing more DB concurrency than that are better served by the PostgreSQL backend anyway (per #2641).
2. Ship `ulimits: nofile: 65536` in the provided `docker-compose.yml` as defense in depth.

Happy to adjust numbers/approach — the important part is that the *default* config should not be able to fd-exhaust itself into database corruption on a stock Docker install.

### Environment

Bambuddy 1.2.5.3 (ghcr.io image) · Docker on Debian LXC · SQLite backend (WAL) · 1 printer (H2D) + phone app + PrintPulse-style API polling

M2ABRAMSTANK:
## Description

SQLite backends only: cap the connection-pool defaults (`pool_size 20 → 10`, `max_overflow 200 → 20`, both still env-overridable via `DB_POOL_SIZE` / `DB_MAX_OVERFLOW`), and ship `ulimits: nofile: 65536` in the provided `docker-compose.yml`.

With WAL active, SQLite's unix VFS **parks** the main-db fd of every closing connection on a per-inode unused-fd list (POSIX advisory-lock safety) for as long as *any* connection to the file stays open — and the pool always keeps one open. Parked fds are reused, so the steady-state fd bill equals the historical **peak** of concurrent connections: up to ~220 with the previous 20+200 defaults, plus live wal/shm fds and sockets. Under Docker's default 1024 soft nofile a stock install can exhaust its fd table from ordinary traffic; a production instance did exactly that ~17.5h after a restart, spent 44h with every new DB connection failing (`disk I/O error` from the connect-time `PRAGMA journal_mode = WAL`), and ended with `database disk image is malformed` and a broken login.

Full mechanism, deterministic repro (fd plateau at peak concurrency, gc-immune, fully released only on `engine.dispose()`), strace evidence, and production timeline in #2883.

Notes:
- Not a driver issue: identical behavior on SQLAlchemy 2.0.51/2.0.52 × aiosqlite 0.21.0/0.22.1.
- 10+20 still allows 30 concurrent SQLite connections — generous for a single-file database serialized by one writer lock; installs that genuinely need more concurrency are better served by the PostgreSQL backend (#2641).
- The compose `ulimits` entry is defense in depth so a future concurrency bump can't silently reintroduce the ceiling.
- Rebased onto `dev` per CONTRIBUTING (originally opened against `main`); the `dev` suite's pool-default test (`test_sqlite_defaults_when_unset`, #2572) is updated in the same commit to match the new SQLite defaults, and the pool docstring no longer claims the old values.

## Related Issue

Fixes #2883

## Documentation

**Pick one**:
- [x] No docs update required — reason: internal pool-default tuning; `DB_POOL_SIZE`/`DB_MAX_OVERFLOW` stay env-overridable and their documented usage is unchanged.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- `backend/app/core/database.py`: SQLite pool defaults 20+200 → 10+20 (env-overridable), docstring updated to match
- `docker-compose.yml`: `ulimits: nofile: 65536` (defense in depth)
- `backend/tests/unit/test_db_pool_and_auth_cache.py`: `test_sqlite_defaults_when_unset` updated to the new defaults

## Screenshots

Not applicable — no UI change.

## Testing

- [x] I have tested with my printer model: H2D

`pytest backend/tests/unit/test_db_pool_and_auth_cache.py` — 9 passed. Full backend suite green locally on `dev` + this change. Deployed on the production instance from #2883 since 2026-08-19 (nofile) / 2026-08-20 (pool caps): fd count stable, no recurrence.

## Additional Notes

The fd-parking behavior is easiest to observe with `ls /proc/<pid>/fd | wc -l` while cycling connections; the count plateaus at peak concurrency and only `engine.dispose()` releases it.


maziggy (changes requested):
The diagnosis is correct and I verified it independently rather than take the
repro on trust. Bisecting WAL against DELETE with our own connect listener
reproduces it exactly: the fd count plateaus at pool_size + max_overflow, is
immune to gc, and drops only on engine.dispose(). Measured cost, peak fds during
a burst versus what stays parked afterwards:

  20+200 (today)  441 peak / 220 parked
  10+90           201 peak / 100 parked
  10+20 (this PR)  61 peak /  30 parked

CI is green on the branch here: ruff clean, 10499 backend tests pass.

One thing to settle before it lands. 30 is the exact ceiling that #2572 showed
saturating, a 93-printer farm logging "QueuePool limit of size 10 overflow 20
reached, connection timed out". That was PostgreSQL and it predates the session
hygiene fixes that halved per-request checkouts, so it does not read across
directly, but I would rather not adopt that number by coincidence. 10+90 keeps
the fd bill at a fifth of a 1024 budget while leaving roughly 3x margin over the
number we know saturates. Are you happy to change it to that?

Two related gaps I plan to cover, either here or in a follow-up. systemd installs
have the same 1024 soft default and deploy/bambuddy.service sets no LimitNOFILE,
so bare metal has the identical ceiling the compose change fixes for Docker. And
check_pool_fits_server() early-returns on SQLite, so the ceiling-versus-RLIMIT
warning that PostgreSQL already gets is missing for the backend that actually
needs it.

Thanks for the write-up on #2883, the parked-fd mechanism is the part #2641
missed and it is what makes the fd bill a ratchet rather than something a quiet
spell recovers.

M2ABRAMSTANK:
Happy to — 10+90 it is: `a428de1c`. I updated the in-code comment and the
defaults test to carry the same reasoning (parked-fd bill ≈ a tenth of the 1024
budget, ~3x margin over the #2572 saturation point) so the number doesn't read
as another coincidence later.

On the two gaps — I took a look while I was in there (no code yet; your call
whether either belongs here or in your follow-up):

**systemd:** it's actually *two* independently-maintained units. There's
`deploy/bambuddy.service` (the copyable template) and the heredoc inside
`install/install.sh`'s `create_systemd_service()` — which is what the
documented install path actually writes, and it does not read the template.
Neither sets `LimitNOFILE`, so both land on systemd's soft-1024 default. A fix
needs the same line in both. Two wrinkles: `create_systemd_service()`
overwrites the unit on every reinstall and only carries `ReadWritePaths`
forward (#2544), so a user's hand-added limit dies on reinstall unless it lives
in a `systemctl edit` drop-in. `install/update.sh` doesn't retrofit unit settings
on upgrade today, but it already stops and starts the live unit each run, so
teaching it to reapply the limit would be a small addition if you want upgrades
to heal this. And the macOS launchd plist the installer generates has the same gap via
its missing `SoftResourceLimits` key. While grepping I also noticed
`docker-compose.test.yml` never got the `ulimits` block the main compose file
gained in this PR — one more copy of the same ceiling.

**`check_pool_fits_server()`:** the SQLite early-return is the first statement
after the docstring, and `test_sqlite_is_skipped_entirely` pins it ("No such
concept, and the probe SQL is PostgreSQL-only"), so an SQLite branch means
narrowing that test as part of the change. The comparison would be
`pool_size + max_overflow` (plus wal/shm fds of live connections and some
non-DB headroom) against `RLIMIT_NOFILE` soft. Nothing in the codebase reads
rlimits today — `resource` is never imported anywhere — but psutil is already a
dependency and `psutil.Process().rlimit(psutil.RLIMIT_NOFILE)` would do it.
POSIX-only, so the Windows target would need the probe skipped, same shape as
the existing early-return.

If you'd like either of them in this PR rather than the follow-up, say the word
and I'll add it. Also happy to add a cheap ratchet test asserting the SQLite
ceiling stays ≤ 100 so the defaults assertion isn't the only guard — tell me if
that's wanted or noise.

maziggy (changes requested):
Second pass: a428de1c is exactly what I asked for, and the reasoning now sits in three places that agree - the docstring, the inline comment, and the test's own name. Verified on f35189c1: ruff clean, ruff format clean, 12160 backend tests pass. The two dev merges pulled in thirteen commits touching database.py and clobbered none of them; the branch's only delta against dev is the pool block, the defaults test and the compose ulimits.

One thing to fix before it lands. backend/app/core/config.py:78 still carries the old numbers:

# ``None`` = use the built-in, dialect-aware
# default (PostgreSQL: pool_size 20 + max_overflow 80; SQLite: 20 + 200).
That is the comment directly above db_pool_size / db_max_overflow, so it is what anyone reading the knobs sees first, and it now contradicts the code one file away. SQLite: 10 + 90 and it is done. Nothing tests a comment, which is exactly why this one needs eyes.

On the two gaps, your audit checks out on every point - I verified each rather than take it on trust. Two systemd unit sources, neither carrying LimitNOFILE; create_systemd_service() overwrites the unit and carries only ReadWritePaths forward; update.sh already stops and starts the unit; the generated launchd plist has no SoftResourceLimits; docker-compose.test.yml has no ulimits; resource is imported nowhere in the codebase; psutil is a hard dependency at >=6.0.0.

Both stay out of this PR, and the reason is the number you just changed. At 10+90 the parked-fd bound is about 100 against a 1024 default, so the DB pool can no longer reach that ceiling by itself, on Docker or on bare metal. LimitNOFILE stops being the fix and becomes defence in depth against everything else holding fds - per-printer MQTT sockets, camera streams, FTP, ffmpeg pipes - which is a different conversation that deserves its own numbers rather than being smuggled in here. docker-compose.test.yml is a CI-only stack nobody deploys, so the missing block there is consistency and nothing more. I will carry both into the follow-up.

One correction on the probe, because it would have bitten you: psutil.Process().rlimit() is not POSIX-only, it is Linux and FreeBSD only - psutil gates the method on the platform module providing it, with the comment "# Linux / FreeBSD only" right above it. macOS is a supported install target here, with an installer, a launchd plist and an updater, and it is the platform with the meanest default nofile of the three. A psutil probe would raise AttributeError or silently skip exactly where the check is worth the most. Stdlib resource.getrlimit(resource.RLIMIT_NOFILE) covers Linux and macOS and is absent on Windows, which is the same skip shape you would have had to write anyway.

On the ratchet test: not needed. test_sqlite_defaults_when_unset already asserts 10 and 90 exactly, so a "<= 100" assertion is strictly weaker than what is there and would only be a second thing to update. Good instinct to ask rather than add it.

PR #2892, opened 2026-08-20
feat: support inventory-only spool assignments

MikeFox303:
## Summary\n- add backwards-compatible apply_to_printer assignment flag (default true)\n- persist the flag and guard manual, AMS-change, and RFID K-profile writes\n- keep inventory events and usage-tracker snapshots unchanged\n\n## Validation\n- 214 targeted inventory/RFID/usage tests passed\n- ruff check passed\n- full suite on Windows is blocked by pre-existing os.geteuid collection issue in test_slice_external_folder_output.py

### DIFF
--- a/.github/workflows/test-image-2892.yml
+++ b/.github/workflows/test-image-2892.yml
@@ -1,0 +1,33 @@
+name: Test Bambuddy PR 2892 image
+
+on:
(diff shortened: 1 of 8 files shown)


maziggy:

Issue #2849, opened 2026-08-16
[Bug]: Drag and drop onto busy printer should queue instead of rejecting


abraha2d:
### Problem or Use Case

When a printer is already busy printing, it would be nice to queue additional prints by dragging and dropping the files onto the printer on the Printers page.

Currently, attempting that just rejects the drop with a "Printer is busy" error. One has to manually upload the file to Archives and queue the print from there.

### Proposed Solution

Dropping a file onto a busy printer should pop up the same dialog as dropping it onto an idle printer.

### Feature Category

Print Queue & Scheduling

### Priority

Would improve my workflow

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Good catch! Forgot the printer card when refactoring the print dispatch.

Available/Fixed in branch dev and available with the next release or daily build. Please let me know if it works for you.

If you find Bambuddy useful, please consider giving it a ⭐ on [GitHub](https://github.com/maziggy/bambuddy) — it helps others discover the project!

Issue #1214, opened 2026-05-05
[Bug]: False-Positive Being Triggered for plate check


Forsyth-Creations:
### Component

Bambuddy

### Bug Description

Hey there! I love Bambuddy, and have recently run into an issue:

<img width="1087" height="1180" alt="Image" src="https://github.com/user-attachments/assets/339c951e-f33c-4e4b-be8c-69273b62c3ac" />

The automated plate check seems to be throwing a false positive based on the grey portion of the bed on the right side (For a P1S). Anyway, my calibration could be to blame, but if this is a common issue for others let me know!

### Expected Behavior

I would expect that it ignore this region, or not assume that it's a part. 

### Steps to Reproduce

1. Enabled plate detection, and took several photos of an empty bed at discrete height intervals (30 mm Z change)
2. Went to start a print 
3. Print was automatically paused due to false positive 

### Printer Model

None

### Bambuddy Version

v0.2.3.2

### SpoolBuddy Version

NA

### Printer Firmware Version

01.10.00.00

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

### Relevant Logs / Support Package

NA

### Screenshots

<img width="1087" height="1180" alt="Image" src="https://github.com/user-attachments/assets/8acede77-497e-457d-91b7-95c6c286e70e" />

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I am using the latest version of Bambuddy
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled

maziggy:
Yes, unfortunately the detection is not working reliable. Problem is, that the camera is in a fixed position, but the bed can be lowered. Until now we haven't found any reliable solution for this. I'm sorry.

-----
If you find Bambuddy useful, please consider giving it a ⭐ on [GitHub](https://github.com/maziggy/bambuddy) — it helps others discover the project!

Forsyth-Creations:
@maziggy I have some ideas on how to fix this, actually. But it would require an automated calibration process, instead of discrete photos. I'll make a fork and make some updates. Although, I suspect this could be solved temporarily with some dark electrical tape 😂

maziggy:
If you want to contribute, please note our rules in CONTRIBUTING.md.

Forsyth-Creations:
[...]

PR #2894, opened 2026-08-20
feat: selectable AI vision backend for the Build Plate Check (opt-in, per-printer)

### LINKED ISSUE #2893: [Feature]: Selectable AI vision backend for the build-plate empty check
### Problem or Use Case

The built-in Build Plate Check (`plate_detection.py`) uses calibration-based pixel differencing against a reference photo of the empty plate. It's sensitive to lighting changes, Z-height differences, and any other drift from the calibration shot — issue #1214 tracks exactly this ("False-Positive Being Triggered for plate check"). maziggy's own comment there:

> Yes, unfortunately the detection is not working reliable. Problem is, that the camera is in a fixed position, but the bed can be lowered. Until now we haven't found any reliable solution for this. I'm sorry.

([source](https://github.com/maziggy/bambuddy/issues/1214#issuecomment-4378987346))

Early bench numbers already reproduce this failure mode directly: under combined calibration drift (chamber light off AND a different plate type vs. the calibration shot — two variables drifted together), the existing check false-positives 10/10 on a genuinely empty plate, with a diff reading indistinguishable from the one it reports on an actually-occupied plate.

I'd like to offer a second, opt-in detection backend that doesn't depend on pixel-perfect calibration: send the snapshot to a vision-language model (any OpenAI-compatible endpoint — local Ollama, OpenRouter, hosted OpenAI, etc.) and use its occupied/empty judgment instead of the pixel diff.

### Proposed Solution

A `bedcheck_backend` setting (`opencv` default / `ai`) behind the existing `check_plate_empty()` service function — same two call sites (auto-pause on print start, manual "Check Plate" button), same result shape, zero call-site changes. `ai` mode posts one downscaled snapshot to a configurable `base_url`/`model`/`api_key` (OpenAI-compatible `chat/completions`, structured JSON output) and maps the verdict back onto the existing result type. Off by default; fails open (assumes empty) on any error, same as today's behavior on a capture failure. No schema migration for the settings — key/value rows like the existing Obico integration. Architecturally this is the same pattern Bambuddy already ships for AI failure detection (`obico_detection.py`) — following that precedent, not introducing a new one.

Facts worth calling out up front, since they're easy to assume wrong from the outside:

- The existing per-printer "Build Plate Check" toggle (`plate_detection_enabled`) is unchanged and still gates whether the auto-check runs at print start, per printer. The backend choice selects *which* engine runs that check — global default plus an optional per-printer override for mixed fleets (nullable column, follows the global setting unless set).
- The AI backend needs **no calibration step**. Unlike the OpenCV path, there's no reference photo to capture and no "not yet calibrated" state — `ai` mode works as soon as a base URL and model are configured. That's the actual point of offering it as an alternative to the calibration-sensitive pixel-diff approach.
- **Call cadence is unchanged and minimal**: once at print start (for printers with the check enabled), manual checks, and a settings test button. Never during a print, never on a timer — this is not a failure-detection replacement and doesn't touch the Obico integration.

I'd frame this as a **selectable alternative backend**, not a replacement for the OpenCV path (which stays the default) or a fix for #1214 specifically — happy to have you weigh in on scope/overlap with #1214, since Forsyth-Creations is assigned there working on auto-calibration for the existing detector. These feel complementary rather than competing (auto-calibration makes the existing detector more robust; this adds a second, differently-shaped detector for people who'd rather point at a model than maintain reference photos) but I'd rather you make that call than assume it.

Implementation note: I have this working end-to-end on my own instance (local Ollama `qwen2.5vl:7b`, ~1.1s median per check, structured-output verdicts, fail-open verified) with tests, i18n for all locales, and docs ready — I can open the PR immediately on your go-ahead.

### Alternatives Considered

Mimicking Obico's own `GET /p/?img=<url>` protocol instead of a generic OpenAI-compatible chat API — rejected because it's a bespoke, single-purpose protocol versus the OpenAI chat-completions shape, the de facto standard already implemented by Ollama, vLLM, OpenRouter, and every major hosted provider.

### Mockups or Examples

The UI is built: a dedicated "Build Plate Check" settings tab (backend/connection card, per-printer monitored-printers card, status card — same two-column layout as the Failure Detection tab), and a decision panel in the printer card's plate-check modal showing which backend produced the verdict and why. Screenshots ready for the PR.

### Feature Category

Monitoring & Stats

### Priority

Would improve my workflow

### Contribution

### Checklist

M2ABRAMSTANK:
## Description

Adds a second, opt-in detection backend for the Build Plate Check: instead of the existing
calibration-based OpenCV pixel-diff, an `ai` backend posts one downscaled camera snapshot to a
configurable OpenAI-compatible endpoint (verified end-to-end against local Ollama's `/v1`;
hosted OpenAI-style endpoints follow the same documented `/v1/chat/completions` contract and
are covered by unit tests, but have not yet been exercised against a live hosted service) and
uses the model's occupied/empty judgment instead. Off by default (`bedcheck_backend` defaults
to `opencv`); same two call sites, same result shape as today (new response keys are additive);
fails open (assumes empty) on every error class, matching the existing capture-failure
behavior. See the linked issue for motivation and framing against #1214.

Scope notes:

- **Call cadence is unchanged and minimal**: the check (either backend) runs once at print
  start (only for printers with the plate check enabled), on a manual check from the printer
  card, and on the settings Test-connection button (synthetic frame). Never during a print,
  never on a timer — during-print failure detection remains the existing Obico integration,
  untouched.
- **Selectable per printer**: a new nullable `printers.bedcheck_backend_override`
  (`'opencv' | 'ai'`, `NULL` = follow the global setting) lets mixed fleets pin individual
  printers to either backend. Settable from the printer card's plate-check modal and from the
  new settings tab; value validated at the schema layer (422 on anything else) and the
  dispatcher ignores stale/invalid stored values by falling back to the global setting.
- **Decision transparency**: `PlateDetectionResult` now carries `backend` and `ai_reason`
  (additive keys, `backend` defaults to `'opencv'` so pre-existing construction sites are
  unchanged), and the printer-card plate-check modal renders a decision panel — which backend
  produced the verdict, confidence, and either the OpenCV pixel-difference or the model's
  stated reasoning. Every fail-open result self-identifies its backend.
- **Own settings tab**: the feature gets a dedicated "Build Plate Check" settings tab (it is
  not failure detection, and the Failure Detection tab is Obico's) with the backend/connection
  card, a Monitored-printers card (per-printer enable + backend), and a Status card showing the
  effective per-printer configuration — laid out in the same two-column responsive pattern as
  the Failure Detection tab.
- **`both` mode deliberately not offered** (a fail-fast chain would skip the AI on exactly the
  #1214 case) and **few-shot deliberately cut** after a real A/B (details below).

### Bench results — PARTIAL (Phase 4)

**Status: PARTIAL.** Only 2 of 10 planned cells (`b`, `c`) have a real, non-self-referential
OpenCV measurement so far; the rest are pending — protocol and current status below. On the
cells measured, the headline result is exactly the failure mode issue #1214 describes: under a
combined lighting + plate-type drift from the calibration shot (see the confound note below the
table), the existing OpenCV check **false-positives 10/10** on one genuinely-empty scene, while
the AI backend scores **0/10 FP** on that same scene, **0/1 FN** on the occupied frame,
**0 verdict flips across k=3** repeats on any of the 13 labeled frames, and a **median 1.11s**
(mean 1.27s) per check on a local `qwen2.5vl:7b` (Ollama, LAN).

| Cell | What it tests | n | OpenCV FP/n | OpenCV FN/n | AI (qwen2.5vl:7b) FP/n | AI FN/n |
|---|---|---|---|---|---|---|
| (a) sanity frames | The 2 calibration-reference photos themselves — AI-only sanity check, not an independent baseline (see note below) | 2 | n/a (self-referential, excluded) | — | 0/2 | — |
| (b) baseline occupied | Mid-print on plate, matching calibration lighting | 1 occupied | — | **0/1** | — | **0/1** |
| (c) lighting change (1 scene, 10 frames) | Chamber light off vs. the lights-on calibration reference — confounded with a plate-type change, see note below | 10 empty (1 distinct scene) | **10/10** | — | **0/10** | — |
| (d) Z-height / gantry position | Empty+occupied at Z/gantry states differing from calibration | 0 | unmeasured | unmeasured | unmeasured | unmeasured |
| (e) plate swap | Empty on a plate type differing from the calibrated one | 0 | unmeasured | — | unmeasured | — |
| (f) hard-occupied (purge blob, skirt remnant, white-on-light, dark-on-dark — 4 sub-cases) | Small/low-contrast occupancy | 0 | unmeasured | unmeasured | unmeasured | unmeasured |
| (g) stale reference | Empty+occupied against an intentionally-aged calibration reference | 0 | unmeasured | unmeasured | unmeasured | unmeasured |

`unmeasured` = n=0 for that arm — not a passing score, no data yet. **Coverage:** cell (a)'s
only 2 candidate frames *are* the OpenCV calibration references themselves — scoring either as
the "live" snapshot while it's also a reference is self-referential, so it's excluded from the
OpenCV arm and reported AI-only as a sanity check. **Cell (c) confound, stated plainly:**
OpenCV's calibration references show a textured-PEI plate, lights on; the lights-off test
frames show a black-PEI plate, lights off — two variables drifted together (lighting *and*
plate type), not lighting alone. It's still the closest thing this corpus has to the #1214
failure mode, but it measures combined calibration drift, not an isolated lights-off test.
**Cell (c) is also one static scene** (8 loop-capture frames + 2 deliberate captures ~4s
apart), reported as `n=10 frames / 1 distinct scene` for that reason.

Determinism: k=3, 0/13 labeled frames flipped verdict (flip rate 0.0).
Latency: 39 mixed calls, min 0.90s / median 1.11s / mean 1.27s / max 1.78s. Warm/cold split (measured separately, k=10 warm after a forced model unload): **warm p50 1.17s / p95 1.32s**; **cold first call 8.4s** (model load included — the verdict on that call was still correct). Idle VRAM residency on this rig measures ~0W, so keeping the model loaded avoids the cold hit at no meaningful power cost.
Tokens/check at 768px (`DOWNSCALE_MAX_EDGE`): ~1228 prompt / ~58 completion.

**Unstaged production datapoint** (my own instance, running this branch): on a
genuinely-empty bed (verified by snapshot), OpenCV reported "objects detected" at 4.8–17.7%
difference across the day while the AI backend answered `is_empty` at 0.95 confidence 3/3 with
zero flips — and later the same day, after a real print finished and the part was left on the
plate, the AI correctly flipped to occupied ("remnants of a previous print"). One rig, not a
controlled study — but it is #1214 observed live, and the correct catch of the state change it
exists to catch.

**D1 gate (zero-shot vs. few-shot) — resolved, not pending.** Run on the 11 non-reference
labeled frames: zero-shot was already at ceiling (0/11 errors), so few-shot (2 reference images
attached) had no floor to improve on — it scored 1/11 errors instead, at +2147 mean prompt
tokens and +0.36s mean latency. Per D1's rule ("ship few-shot only if the bench shows a
measurable gain"), the few-shot code path was **cut from this branch** rather than shipped
inert — this diff is zero-shot-only.

**Never-calibrated printer (measured, both backends, real corpus frames):** for a printer with
zero calibration references, OpenCV returns `is_empty=True, confidence=0.0,
needs_calibration=True` on **both** the verified-empty frame and the genuinely-occupied frame —
the documented fail-open waves an occupied plate through until someone calibrates. The AI
backend needs no calibration and read the same two frames correctly (`occupied → is_empty=false`
@0.95, `empty → is_empty=true` @0.95). Not a defect in either backend — it quantifies the
out-of-box behavior gap: the AI backend is useful before any calibration exists.

**Cost:** at ~1,228 prompt + ~58 completion tokens per check — local Ollama on my RTX 3090:
~0.12 Wh/check (~1.3s at ~350W), i.e. **~$0.02 per 1,000 checks** at $0.15/kWh. A hosted
mini-tier vision endpoint at $0.15/$0.60 per 1M tokens prices out to **~$0.22 per 1,000
checks** — order-of-magnitude only, since image-token accounting differs per provider. With the
check running only at print start + manual + test-connection, even a busy 20-print/day printer
is ~600 checks/month.

**`qwen2.5vl:32b` second-model run — dropped, by the same rule that cut few-shot:** 7b is at
ceiling on this corpus (0/11 errors on the non-reference frames), so a 3× larger model has no
floor to improve on and nothing to measure against here. If review wants a second model anyway,
I'll run it on request.

**Remaining before this can be called a complete bench:** cells (d), (e), all four (f)
sub-cases, and (g) — every one of them needs a physical setup change at the printer (moving the
gantry/bed, swapping plates, placing purge blobs and small parts, aging a reference), so I'm
adding them as those states occur in normal use rather than staging them artificially. The
software-side items from the earlier version of this list (warm/cold p50/p95, cost line, the
never-calibrated frame, the second-model decision) are all done above. Happy to prioritize the
physical cells before merge if useful for review — your call.

**Limitations:** one printer (H2D), one camera (built-in), N = 13 labeled frames, one labeler.
Treat FP/FN rates as indicative for this rig, not a universal accuracy claim.

## Related Issue

Fixes #2893

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#73

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- New `services/bedcheck_ai.py`: OpenAI-compatible chat-completions call (structured JSON
  response format), one parse-failure retry, confidence clamp, and a generic fail-open reason
  mapper — never the raw exception string, which for `httpx.HTTPStatusError` embeds the
  configured base URL and would leak it to CAMERA_VIEW-gated callers.
- `plate_detection.py`: `check_plate_empty()` becomes a selector (`opencv` default / `ai`,
  fail-closed to `opencv` on any settings-read error); the OpenCV implementation is renamed
  `_check_plate_empty_opencv` with its body unchanged. `PlateDetectionResult` gains additive
  `backend` + `ai_reason` fields (default `'opencv'`/`None` — every pre-existing construction
  site keeps its meaning unmodified); all AI paths, including every fail-open branch, tag
  `backend="ai"`.
- **Per-printer override**: nullable `printers.bedcheck_backend_override` column (guarded
  `ALTER TABLE` in the existing startup-migration path — same pattern as
  `plate_detection_enabled`), `PrinterUpdate` validator (only `'opencv'`/`'ai'`/`null`;
  omitted = unchanged, `null` = clear), exposed on `PrinterResponse`, and honored by the
  dispatcher at both call sites (print-start auto-check and the manual camera route) with
  precedence: valid override > global setting > `'opencv'`.
- 4 new settings keys, key/value rows, no schema migration: `bedcheck_backend`,
  `bedcheck_ai_base_url` (LAN-guarded via `assert_safe_lan_service_url`), `bedcheck_ai_model`,
  `bedcheck_ai_api_key` (sensitive, redacted on read like other API keys).
- New `routes/bedcheck_ai.py`: one `/bedcheck-ai/test-connection` endpoint
  (`RequirePermissionIfAuthEnabled(SETTINGS_UPDATE)`), mirroring Obico's test-connection
  pattern — sends a synthetic 64×64 gray frame, never a real printer's camera, and re-asserts
  the LAN guard at request time.
- **Frontend, printer card**: the plate-check modal gains a decision panel (backend badge,
  verdict, confidence, OpenCV pixel-difference or the model's verbatim reasoning, plus an
  "applies to OpenCV only" hint over the references/ROI sections when AI is active) and a
  per-printer backend selector (Global default / OpenCV / AI).
- **Frontend, settings**: new dedicated "Build Plate Check" tab (nav entry, settings-search
  registration, `SettingsSearchTab` union) hosting the backend/connection card, a
  Monitored-printers card (per-printer plate-check enable + backend select), and a Status card
  (global backend, endpoint · model, per-printer effective mode) — two-column responsive
  layout matching the Failure Detection tab. The AI card is no longer inside Failure
  Detection.
- i18n: all new user-facing strings in **all 13 locales** with real translations (parity check
  green, 6,182 leaves per locale).
- Tests: `backend/tests/unit/test_ai_bed_check.py` (settings validation, confidence clamp,
  verdict parsing, one test per fail-open error class — each also asserting backend
  attribution — and verdict-to-result mapping), extended `TestSelectorDispatch` (per-printer
  override precedence, invalid-override fallback) + `TestBedcheckBackendOverrideSchema` in
  `test_plate_detection.py`, SSRF-guard registration tests, API-key scrubbing coverage, and
  two SettingsPage tab-navigation tests.

```
 backend/app/api/routes/bedcheck_ai.py              |  39 ++
 backend/app/api/routes/camera.py                   |   1 +
 backend/app/api/routes/settings.py                 |   1 +
 backend/app/core/database.py                       |   4 +
 backend/app/main.py                                |   3 +
 backend/app/models/printer.py                      |   3 +
 backend/app/schemas/printer.py                     |  13 +
 backend/app/schemas/settings.py                    |  49 +-
 backend/app/services/bedcheck_ai.py                | 408 +++++++++++++++
 backend/app/services/plate_detection.py            | 130 ++++-
 backend/tests/conftest.py                          |   6 +
 backend/tests/integration/test_settings_api_key_scrubbing.py |   6 +-
 backend/tests/unit/services/test_plate_detection.py | 318 +++++++++++-
 backend/tests/unit/test_ai_bed_check.py            | 551 +++++++++++++++++++++
 backend/tests/unit/test_outbound_url_ssrf_guards.py |  23 +-
 frontend/src/__tests__/pages/SettingsPage.test.tsx |  33 ++
 frontend/src/api/client.ts                         |  31 ++
 frontend/src/components/BedCheckAiSettings.tsx     | 323 ++++++++++++
 frontend/src/lib/settingsSearch.ts                 |   3 +-
 frontend/src/pages/PrintersPage.tsx                |  75 ++-
 frontend/src/pages/SettingsPage.tsx                |  31 +-
 frontend/src/i18n/locales/* (13 files)             | 512 +
 static/ (4 files, rebuilt bundle: chunk hash renames + index.html ref) | ±136
 38 files changed, 2685 insertions(+), 150 deletions(-)
```

(Literal `git diff --stat upstream/dev...HEAD` at the current tip, with two rollups per this
repo's convention of not pasting generated output line-by-line: the 13 i18n locale files
(+512 total, real translations) and the rebuilt `static/` bundle (4 files, ±136). Every other
line is unedited from the real diffstat.)

## Screenshots

Attaching as a comment right after opening (three shots: printer-card plate-check modal with the decision panel + per-printer selector; the Build Plate Check settings tab in a wide window; Test-connection success state).

## Testing

- [x] I have tested with my printer model: Bambu Lab H2D

`ci.yml` does not run on this PR — it triggers only on `pull_request: branches:[main]` and this
PR targets `dev`, matching every other `dev`-based PR. Same for `codeql.yml`. GitGuardian and
`security.yml` (Bandit/Trivy/pip-audit/npm-audit) do run.

Local verification, run in full at the current tip in place of the CI that won't fire:

- **Backend suite**: `10,496 passed` (`pytest backend/tests/ -n 30`, excluding one unrelated
  FTP test file which passes separately: `99 passed`). Two flaky failures can appear per
  parallel run in pre-existing tests this PR doesn't touch; each passes in isolation
  (confirmed `-n 30` contention, not regressions).
- **Backend lint**: `ruff check` clean; `ruff format --check` clean. All pre-commit hooks pass
  on every changed file (ruff, ruff-format, whitespace, EOF, merge-conflict, private-key,
  import-shadowing, tsc).
- **Frontend**: `npx tsc -b` clean (note: project-mode `tsc -b`, which checks more than
  `tsc --noEmit`), `npm run lint` clean, `npm run test:run` `3,186 passed` — the only 3
  failures are `PrintersPageDropOnBusy.test.tsx` (upstream #2849), reproduced identically on a
  clean `upstream/dev`@`de74914d` worktree and passing inside Docker's pinned Node, i.e.
  pre-existing host-environment breakage, not this PR.
- **i18n parity** (`npm run check:i18n`): 13/13 locales, 6,182 leaves each, in parity with
  `en`.
- **Docker suite** (`./test_docker.sh`): 11/11 — test + production image builds, backend and
  frontend suites inside Docker, health/API/static endpoint checks, and the integration
  test-runner, all at the current tip.
- **Security**: bandit (CI invocation: `--severity-level medium -x backend/tests`) — no
  issues; `pip-audit` — no known vulnerabilities; `npm audit --audit-level=high` — 0
  vulnerabilities.
- **Static build**: single `npm run build` on the final integrated commit; churn proportional
  to this feature's own frontend surface.
- **Live**: this exact branch is deployed on my production instance (Docker,
  amd64) against a real H2D + local Ollama — settings persistence across restarts, the
  per-printer override end-to-end (including 422 on invalid values and the guarded startup
  migration on an existing SQLite DB), test-connection, fail-open against a dead endpoint
  (generic message, no URL leak), and both backends' live verdicts have all been exercised
  there.

## Checklist

## Additional Notes

Frontend `static/` bundle rebuilt and committed per project convention (`npm run build`, run on
the final frontend state of the branch). The AI backend never receives calibration references
or the detection ROI — it evaluates the full frame zero-shot; the references/ROI machinery
remains OpenCV-only and untouched. The `bedcheck_ai_api_key` is registered in
`_SENSITIVE_FIELDS_FOR_API_KEY` and scrub-tested.


M2ABRAMSTANK:
image
image

M2ABRAMSTANK:
Screenshots, as promised in the description — from my production instance running this branch (v1.2.6b1 base, real H2D + local Ollama):

**Build Plate Check settings tab** — backend/connection card, Monitored printers, and the Status card showing the effective per-printer configuration:

settings tab

**Test connection success** — live round-trip against the configured endpoint (`qwen2.5vl:7b` on LAN Ollama), schema-valid verdict in 1.5s:

test connection

The third shot (printer-card plate-check modal with the decision panel, objects detected):

image

Printer-card plate-check modal with the decision panel, no objects detected
image

maziggy (changes requested):
Thanks for the effort here. The implementation quality is high, and the security work
in particular (the SSRF guard registered and then re-asserted at request time, the
generic error strings with a test that a URL leak would fail) is better than most PRs
I get.

Four things I would need changed regardless of that decision:

1. The AI path fails open on every error class with nothing but a log line - wrong
model name, dead endpoint, 401, unsupported response_format all end as "plate is
empty". On a manual check the new panel renders a green "Plate appears empty" at 0%
confidence with "AI bed-check unavailable" printed underneath it. A safety check that
can be silently dead needs a visible health state.

2. bedcheck_ai.py:297 puts the model's confidence into difference_percent. main.py:3119 broadcasts it as "Diff: 95.0%" and main.py:3129 feeds it to the {difference_percent} notification variable, which is documented as a pixel difference. The modal hides that row; the notifications do not.

3. DEFAULT_TIMEOUT is 60s and sits inline in on_print_start, ahead of archive creation,
with the chamber light forced on and the print running unpaused - up to 120s if the
parse retry fires. OpenCV returns in about a second. The print-start path needs a much
shorter budget; the test button can keep 60.

4. The settings description and all 13 locales advertise https://api.openai.com/v1, but
the request sends max_tokens (the newer OpenAI models reject it and want
max_completion_tokens) and response_format json_schema strict, which not every vision model accepts. Both failure modes land in point 1. Either verify hosted or take it out of the copy.

Smaller: the auto-save in BedCheckAiSettings has no onError, so a 422 from the URL guard on a half-typed value is silent; the Monitored-printers card is not gated on
printers:update the way the printer-card selector is; the route reaches into the private
_load_ai_settings(); and the cold-start figure is 27.84s in the code comment but 8.4s in
the description.

One last thing on the bench: it reads as more conclusive than it is. Two of ten cells,
thirteen frames, one printer, one labeler, and the 10/10 OpenCV result is a single static
scene where lighting and plate type drifted together. You say all of this yourself, which
I appreciate - but it shows OpenCV breaking under compound drift, not that the AI backend is better.

Issue #2801, opened 2026-08-09
[Bug]: Plate-clear gate kills AMS drying in a loop — drying stopped for a print that can never dispatch


superflyer11:
### Product

Bambuddy

### Area

Filament / AMS / Spoolman

### Bug Description

When a printer is in `FINISH` with `awaiting_plate_clear=True` and there is a pending queue item, the scheduler stops and restarts AMS drying roughly once a minute, indefinitely. Drying never runs for more than ~30 s, and manual drying cycles on *other* AMS units on the same printer are killed as collateral.

This is distinct from #2770 and #1892 (the 700 s / 30-minute early-stop). Here the stop reason is explicitly `print takes priority`, driven by a queue item that cannot dispatch.

There appear to be two separate defects:

**1. `_stop_drying()` is called before verifying the stop would actually unblock dispatch.**

In `backend/app/services/print_scheduler.py` (~L599-618):

```python
if not printer_idle:
    if self._drying_in_progress.get(item.printer_id):
        if block_for_drying:
            busy_printers.add(item.printer_id); continue
        else:
            await self._stop_drying(item.printer_id)          # unconditional
            printer_idle = self._is_printer_idle(item.printer_id, require_plate_clear)
            if not printer_idle:                              # still False — plate not clear
                busy_printers.add(item.printer_id); continue  # drying killed for nothing
```

`_is_printer_idle()` returns False whenever `require_plate_clear and is_awaiting_plate_clear(printer_id)` (~L2132). So when the *only* blocker is the plate-clear acknowledgment, drying is sacrificed for a print that was never going to start. `_check_auto_drying` then re-arms on the next tick because humidity is still over threshold, and the cycle repeats forever.

`_stop_drying()` also sweeps **every** AMS with `dry_time > 0` (~L2532), not just the unit auto-drying started — so a manually started cycle on another AMS is stopped too, even though manual drying is not tracked in `_drying_in_progress`.

**2. `mid_print` is inferred from `busy_printers`, which does not mean "printing".**

~L2352:

```python
mid_print = (pid in busy_printers and print_drying_enabled and supports_drying_while_printing(model, firmware))
```

`busy_printers` means "the queue could not dispatch to this printer this tick". The printer is added to it at ~L614 *because* the plate-clear check failed. So a printer that is idle-but-plate-not-clear is classified as mid-print, which:

[...]

Issue #1892, opened 2026-07-03
[Bug]: Auto-drying stops manually started AMS drying after exactly 30 minutes (humidity re-check during heating always reads low)


Spionkiller01:
### Summary

When AMS drying is started **manually** (on the printer) while Bambuddy's auto-drying feature is enabled, Bambuddy sends a **stop-drying command after exactly 30 minutes**, cutting e.g. a 8-12 h PETG/PA drying cycle short every single time.

I first suspected the printer and contacted Bambu Lab support — they analysed the printer logs and confirmed that **an external tool issued the stop-drying MQTT command**. That external tool is Bambuddy.

### Environment

- Bambuddy: current `latest` Docker image
- AMS 2 Pro (`module_type` n3f/n3s path)
- Auto-drying (ambient/queue) enabled, humidity threshold configured to 40%

### Log evidence (Bambuddy backend)

```
2026-07-03 11:30:30,007 INFO [backend.app.services.print_scheduler] [-] Auto-drying: printer 1 AMS 0 — humidity 18% <= threshold 40% after 30m, stopping drying
```

Exactly 30 minutes after the manual start, every time.

### Root cause

`backend/app/services/print_scheduler.py`, `_check_auto_drying()`:

```python
# Already drying — check if humidity dropped below threshold (with minimum drying time)
if dry_time > 0:
    if pid not in self._drying_in_progress:
        # Drying we didn't start (manual or from before restart) — track but don't stop
        self._drying_in_progress[pid] = time.monotonic()
    started_at = self._drying_in_progress[pid]
    elapsed = time.monotonic() - started_at
    if humidity is not None and humidity <= humidity_threshold and elapsed >= self._min_drying_seconds:
        ...
        printer_manager.send_drying_command(pid, ams_id, temp=0, duration=0, mode=0)
```

Two problems combine here:

1. **The code contradicts its own comment.** The comment says manual / pre-restart drying should be *"track but don't stop"* — but the very next lines apply the humidity auto-stop to it anyway. Manually started drying is treated exactly like Bambuddy-initiated drying.

[...]

PR #2895, opened 2026-08-21
feat: sustained-humidity wait before ambient auto-drying starts (opt-in)

### LINKED ISSUE #2518: [Feature]: Delay automatic drying
### Problem or Use Case

When I go to swap in a new roll of filament the humidity increase will trigger a dryer run, which is useful but while drying the ams2 cannot run a filament detection.

### Proposed Solution

If I could delay the auto drying 5 minutes or so the ams would have time to finish its filament detection before the drying kicked in. 

### Alternatives Considered

Manually canceling the drying works though I have to be quick to get the filament re-read before the drying kicks in again

### Feature Category

Monitoring & Stats

### Priority

Nice to have

### Mockups or Examples

_No response_

### Contribution

### Checklist

M2ABRAMSTANK:
## Description

Ambient auto-drying starts the instant an AMS humidity reading exceeds the Fair threshold. Opening the AMS lid puts a gulp of room air on the sensor, so a transient the desiccant would absorb on its own in minutes buys a multi-hour dry cycle — on the reporter's P2S without the AMS 2 power adapter, reportedly one that also locks out filament detection for the duration.

This adds **`ambient_drying_sustained_minutes`** (int, default **0 = off**, UI-clamped 5–240): when set, a pure-ambient start waits until the unit's humidity has stayed above the threshold **continuously** for that many minutes.

*Filed per the implementation proposal in #2518, where I've also requested assignment — @ryansouza is the current assignee, so happy to defer or hand this off if he's actively on it.*

Semantics (as proposed in the #2518 comment):

- **Reset on any dip**: a reading back at/below the threshold clears the wait; the clock restarts at the next crossing.
- **Absent reading = no information**, not a dip — a sensor skipping a beat doesn't reset the count.
- **Observation gap > 120s restarts the wait** (`AUTO_DRY_SUSTAINED_GAP_SECONDS`): a stretch nobody measured (restart, disconnect, print ran) is not evidence the humidity stayed high.
- **Only a pending schedule stays instant**: a printer with a scheduled queue item keeps today's instant behavior, whether idle or printing — minutes burned ahead of a scheduled print is what pre-print drying exists to prevent (the exemption keys on the pending schedule itself, not on the queue-drying toggle). An ambient start that merely happens mid-print — Continue drying while printing is a permission overlay, not a trigger — serves the same wait as any other ambient start. (This tightened during live testing: the first cut exempted mid-print starts, and a 2-point threshold crossing mid-print promptly bought an instant 12h command with no deadline behind it.)
- **Overlaps, never stacks**: the wait is armed *before* the cooldown/suspension gates and evaluated *after* them, so a post-cycle re-dry waits max(30-min cooldown, N), not the sum.
- **Delays starts only**: running and manual cycles untouched, per the #1892/#2801 behavior. Default 0 is verified byte-identical to today's behavior.

Streak state is a separate in-memory dict keyed `(printer_id, ams_id)` — deliberately not `_auto_dry_units`, whose membership means "Bambuddy armed a cycle" (conflating them is the #2801/#1892 regression class). State is popped at both cycle-end paths (`pop("running")` in-loop and `forget_auto_dry_cycle`) and pruned for disconnected printers.

**Open question for you**: is 120s the right observation-gap ceiling? ~4 consecutive missed scheduler passes silently restart every streak. It's deliberately conservative (never counts unmeasured time toward the wait) but a slow/loaded instance will wait longer than configured. Happy to raise it or make it derive from the scheduler interval — your call.

## Related Issue

Fixes #2518

## Documentation

- Wiki: maziggy/bambuddy-wiki#74

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- `backend/app/services/print_scheduler.py`: sustained-wait streak tracking + gate in the ambient arm path
- `backend/app/schemas/settings.py`: `ambient_drying_sustained_minutes` on AppSettings (default 0, ge=0 le=240) + AppSettingsUpdate
- `backend/app/api/routes/settings.py`: key added to the int-coercion list
- `frontend/src/pages/SettingsPage.tsx`: toggle + minutes input under Settings → Print Queue → Auto-Drying (visible only with ambient drying on). The input holds a transient draft while focused and commits clamped 5–240 on blur/Enter — the same treatment the per-filament humidity thresholds got in #1605 — so mid-typing values aren't rewritten and a stray value can't 422 the debounced whole-page save; empty/non-numeric reverts to the saved value. Covered by 4 SettingsPage tests (3 fail against a per-keystroke clamp).
- `frontend/src/i18n/locales/*`: 3 keys × 13 locales (`check:i18n` parity green, 6,150 leaves)
- `backend/tests/unit/test_scheduler_auto_drying.py`: 66 → 81 tests
- `CHANGELOG.md` entry under 1.2.6b1 Added; `static/` rebuilt on the branch

## Screenshots

Before (Queue Auto-Drying card, ambient drying on):

before

After — new **Require sustained humidity** toggle and **Sustained time (minutes)** input, shown only while ambient drying is enabled:

after

## Testing

- [ ] I have tested with my printer model: H2D — deployed on my own instance (H2D + AMS 2 Pro) since 2026-08-20 with the settings roundtrip verified live; the physical lid-open verification of the streak arm/reset will follow as a PR comment, and I'll tick this box then

`test_scheduler_auto_drying.py`: **82 passed** (16 new: default-off unchanged, arm/mature/reset, dip reset, absent-reading tolerance, observation gap, manual-cycle immunity, scheduled-item exemption (idle and mid-print), mid-print ambient waits, cooldown overlap, disconnect prune). Full backend suite green (10,415 passed); `test_docker.sh` 11/11. Frontend green except 3 pre-existing failures in `PrintersPageDropOnBusy.test.tsx` (#2849) that fail identically on pristine `dev` and pass inside the Docker suite (host Node version). `npm run check:i18n` run standalone (a vitest failure short-circuits the chained run) — parity green, 6,150 leaves per locale.

Beyond line coverage, the new logic was **mutation-tested**: targeted mutants (in-loop cycle-end pop removed, arm-during-cooldown overlap inverted, mid-print exemption dropped, sync-prune removed, and others) were each applied to the scheduler and verified to make at least one test fail; five initially-surviving mutants got five new tests, then every mutant was re-applied and confirmed killed.

## Checklist

## Additional Notes

In-memory state, per the proposal in the #2518 comment: the wait restarts if Bambuddy restarts — worst case is one delayed dry, and the >120s-gap rule makes a restart conservative (waits longer, never shorter).


maziggy (changes requested):
This is good work. The state design in particular: keeping _auto_dry_above separate from _auto_dry_units, with the comment naming the exact #2801/#1892 regression class that conflating them would reproduce, is the right call and the right reason. Cleanup covers all four paths, arm-before-gates gives real overlap with the cooldown rather than stacking, and the mid-print tightening you found in live testing is the correct
semantics. The mutation testing is appreciated.

One thing I want before this merges. The hardware box is still unticked, and the physical
lid-open verification is the one measurement the whole feature rests on. The design
assumes the spike decays back under the threshold within the wait, and the toggle picks 10 minutes as its default with nothing behind that number. If a real lid-open transient on a P2S takes 40 minutes to come back under Fair, a 10-minute default rides out nothing and the user still buys the 12h cycle. Please post the humidity trace from lid-open to back-under-threshold, and set the default from what it shows.

To answer your open question: keep the conservative semantics, but derive the ceiling
instead of fixing it - max(4 * self._check_interval, 120) - and log the restart at INFO
with the measured gap. Right now the dip reset logs at INFO and has a test asserting it,
while the gap restart writes a fresh entry with no log at all. That is the asymmetry that
matters: a slow pass silently voids every streak, and someone who set 60 minutes and
never gets a dry has no way to see why.

Two smaller things:

- The streak update has no sustained_minutes > 0 guard, so every install writes entries on every above-threshold pass whether or not anyone turned this on. Bounded and pruned, so not a leak, but the off path should be inert and it is a one-line guard.

- printers_with_scheduled exempts all AMS units on a printer with any pending scheduled item, including units the job never touches. That granularity predates you, but this setting makes it visible: anyone running a standing schedule effectively never gets the wait. Worth a line in the wiki PR at least.

Nit: the wait gate sits ahead of the dry_sf_reason check, so a unit the firmware refuses
to dry logs "waiting" for N minutes before hitting the skip it was always going to hit.

Finally, please rebase. The only conflicts are CHANGELOG.md, static/index.html and the
bundle rename - nothing in print_scheduler.py or the locales. Drop the CHANGELOG entry, I write those, and rebuild static/ as the last commit so it stops colliding on every dev merge.

PR #2918, opened 2026-08-22
Allow Avery label sheets to start at an unused position (#2879)

### LINKED ISSUE #2879: [Feature]: Allow starting position offset when printing Avery label sheets
### Problem or Use Case

Bambuddy currently supports Avery 5160 and Avery L7160 label sheets, but generated PDFs always begin with the first label in the top-left position.

This makes it difficult to reuse a partially used sheet. For example, if I previously printed seven labels from an Avery 5160 sheet, the next available label is position 8. Bambuddy currently places the next label at position 1, which would print over a used or removed label. The only reliable option is to use a new sheet each time.

Partial-sheet reuse is a common part of printing address labels. Avery 5160 sheets contain 30 labels, so requiring a new sheet for every small batch creates a significant amount of unnecessary waste. Spool labels are also commonly printed in small quantities as new filament is purchased, rather than in batches large enough to fill an entire sheet.

The current sheet renderer calculates each label position beginning from index zero:

https://github.com/maziggy/bambuddy/blob/dev/backend/app/services/label_renderer.py#L460-L469

Specifically, enumerate(chunk) always begins at zero, and the row and column are calculated directly from that index. Because there is no starting-position value in the label request, every generated sheet begins at the first slot:

https://github.com/maziggy/bambuddy/blob/dev/backend/app/api/routes/labels.py#L53-L65

The original spool-label feature in #809 added support for Avery sheet templates. This request would make those templates practical for repeatedly printing small batches without discarding the unused portion of each sheet.

### Proposed Solution

Add an option to choose the first available label position when printing a sheet template.

For example:

Avery 5160 would allow positions 1 through 30.
Avery L7160 would allow positions 1 through 21.
Selecting position 8 would leave positions 1 through 7 blank and place the first selected spool label in position 8.
Labels would continue in normal reading order from left to right and top to bottom.
If the selected labels exceed the remaining positions on the first page, additional pages would begin at position 1.
The default would remain position 1, preserving the current behavior.
The option would only apply to sheet templates. Single-label templates would remain unchanged.

The UI could present this as either:

Starting label position: 8
Skip used labels: 7

I think “Starting label position” is slightly clearer because it corresponds directly to the physical location on the sheet. A short explanation such as “Positions 1 through 7 will be left blank” would make the result unambiguous.

The existing print modal passes the selected template and options directly to the label endpoints:

https://github.com/maziggy/bambuddy/blob/dev/frontend/src/components/LabelTemplatePickerModal.tsx#L269-L280

A starting-position value could follow the same request path and be used by the sheet renderer when calculating the first page’s row and column. Backend validation should restrict the value according to the selected sheet template.

It may also be useful to remember the last selected starting position while the modal remains open, but I do not think it needs to be saved permanently. Starting at position 1 by default would remain the safest behavior.

### Alternatives Considered

I considered editing the generated PDF manually, but repositioning labels accurately while preserving the Avery layout is inconvenient and easy to get wrong.

Printer settings such as margins, scaling, or page offset do not solve the problem because they shift the entire page rather than leaving a specific number of label positions blank.

Another workaround is to use a new sheet for every print job, but that wastes most of the sheet when printing only one or two spool labels at a time.

Users could also attempt to rotate or reverse-feed a partial sheet, but that depends on the printer, can cause alignment problems, and does not provide a reliable way to select an arbitrary unused position.

### Feature Category

Spool Inventory

### Priority

Would improve my workflow

### Mockups or Examples

A simple control in the spool label dialog would be sufficient:

Sheet options

Starting label position: [8]

Positions 1 through 7 will be left blank.

For Avery 5160, the positions would follow this order:

 1   2   3
 4   5   6
 7   8   9
10  11  12
...
28  29  30

A visual 3 by 10 position picker could be helpful in the future, but a numeric field or dropdown would fully address the use case without requiring a more complicated interface.

### Contribution

### Checklist

whitigol:
## Description

Avery sheet PDFs always started in the upper-left slot, so reusing a partially used sheet meant printing over an existing label or discarding the unused portion. This adds a one-based starting label position to the existing print-label dialog and both inventory label endpoints.

The selected offset applies only to the first Avery page. Labels continue left-to-right and top-to-bottom, while later pages restart at position 1. Position 1 remains the default, preserving existing callers and single-label templates.

This PR also fixes two label-printing bugs found while validating the feature: selecting a spool near the bottom of the dialog could scroll the entire modal out of view, and a transparent filament swatch could leak its PDF alpha state into the QR code drawn after it.

## Related Issue

Fixes #2879

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#75

Website: not required because this is a supporting enhancement to the existing Printable Spool Labels feature and does not add or change a broader marketing claim.

**Pick one**:
- [x] Docs PR(s) linked above
- [ ] No docs update required

## Type of Change

- [ ] Breaking change
- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition or update

## Changes Made

- Add Starting label position to the print-label dialog, with localized guidance and template-specific validation: 1–21 for Avery L7160 and 1–30 for Avery 5160.
- Send starting_position through both built-in inventory and Spoolman label requests; keep 1 as the backward-compatible default.
- Offset only the first sheet page and restart overflow pages at the first slot.
- Reject positions outside the selected sheet capacity and reject non-default offsets for single-label templates.
- Keep modal scrolling confined to the spool list by making the outer panel clip overflow instead of becoming a focus-scroll container.
- Isolate ReportLab swatch drawing in a saved graphics state so RGBA transparency cannot fade the QR code or later label content.
- Add backend, frontend, renderer, and integration coverage; update all 13 locales, README, and CHANGELOG.

## Screenshots

| Before | After |
| --- | --- |
| The picker has no way to skip used sheet positions.<br>Before: label picker without a starting label position field | Avery sheets can begin at the first unused position.<br>After: label picker with the starting label position field and sheet ranges |

## Testing

- [x] I have tested with a Brother monochrome printer using Avery 5160 US Letter label sheets; the physical print aligned correctly
- [x] The PDF print dialog/preview showed the selected starting-position offsets correctly before printing
- [x] Docker backend suite: 10,692 passed, 22 skipped
- [x] Label renderer unit suite: 40 passed
- [x] Label picker component suite: 21 passed
- [x] ESLint, TypeScript type checking, i18n parity, and production frontend build passed
- [x] Pre-commit hooks passed for every commit
- [x] Browser reproduction: toggling spool #48 no longer scrolls the modal panel out of view
- [x] Transparent-spool PDF request returned 200 application/pdf; regression coverage verifies the QR draw occurs after alpha state restoration
- [x] Docker production image built and reached healthy status
- [x] Wiki strict MkDocs build and desktop/mobile preview passed

The full Docker frontend run completed 3,230 tests and repeatedly hit one unrelated existing one-second timing failure in SliceModal.test.tsx under full parallel load. The exact failing test passes when rerun alone, and all 21 changed label-picker tests pass in both local and Docker runs. No unrelated slicing code was changed.

## Checklist

## Additional Notes

- The branch targets and is based on upstream dev as required by CONTRIBUTING.md.
- The starting position resets to 1 whenever the dialog opens; it is intentionally not persisted.
- The two bug fixes are separate commits from the issue implementation and documentation changes.
- No database migration or configuration change is required.


maziggy:

whitigol:

Done 🙂

maziggy:
There are still conficts. Changelog is not required.

whitigol:
Try now?

maziggy (changes requested):
Thanks @whitigol - this is a well-built PR, and the two drive-by fixes are welcome.

I verified the transparency bug independently: on dev the swatch sets the alpha graphics
state and never restores it before the QR image is drawn, so a fully transparent spool was
rendering its QR at alpha 0. Good catch. Pagination is correct at every boundary I tried
(position 30 with 3 labels, position 8 with 23 and with 24, position 1 with 30 and with 31),
and all 13 locales are properly translated rather than left in English.

Two things I would like changed before merge:

1. The starting-position bound in the modal is hardcoded to 30 for every template. With
   L7160 selected and 25 entered, the helper text says positions 1 through 24 will be left
   blank and the invalid hint says 1 to 30, while the L7160 button is disabled - the
   guidance contradicts the disabled button exactly when the user needs to know why.
   Please drive that feedback from the selected sheet's capacity.

2. "count" is a reserved pluralization key in i18next, so starting at position 2 renders as
   "Positions 1 through 1 will be left blank on the first sheet." Either add the _one form
   per locale or switch to a non-reserved variable name.

Non-blocking, for a later pass: capacity now lives in three places (the renderer, the modal,
and the range sentence in all 13 locale files), so adding a third sheet template means
touching all of them. get_sheet_capacity() already exists server-side if you want to collapse
that. Same for the alpha regression test - it keys off the ReportLab-generated ExtGState name
/gRLs0, which works today but will not age well.

No CHANGELOG needed on your side, I will add the entry at merge.

whitigol:
Good catches, thanks! Both issues are fixed now and I added coverage for the cases you listed.

maziggy (approved):

Issue #2908, opened 2026-08-22
[Bug]: Per-spool core weight is silently discarded in Spoolman mode — the edit field renders, accepts input, and never persists


ojimpo:
### Product

Bambuddy

### Area

Filament / AMS / Spoolman

### Bug Description

In Spoolman mode the Inventory edit form shows a **core weight** field, pre-filled with the
current value, and accepts edits — but the value is never written anywhere. Both the create and
update schemas declare it and drop it:

`api/routes/spoolman_inventory.py:309` (create) and `:350` (update):

```python
core_weight: int = Field(
    250, ge=0, le=10_000
)  # Accepted for schema parity but not persisted to Spoolman (stored on filament type, not spool)
```

The read path still surfaces a value (`_spoolman_helpers.py:270` derives it from
`spool.spool_weight ?? filament.spool_weight ?? 250`), so there is nothing in the UI to suggest
the edit was ignored — the field simply snaps back to the inherited value.

This is not cosmetic. The measured-weight endpoint uses that same tare:

`api/routes/spoolman_inventory.py:1109`:

```python
core_weight = _safe_float(raw_tare, 250.0)
remaining = max(0.0, data.weight_grams - core_weight)
```

so entering a gross weight for a spool whose real empty weight differs from the filament-level
value silently produces a wrong remaining weight. In my inventory the two tares in use are 250 g
(Bambu reusable spool + cardboard core) and 180 g (a third-party spool) — a 70 g error on every
weigh-in for the mismatched spools.

Spoolman does have a per-spool field for this: `SpoolParameters.spool_weight`, documented as
"The weight of an empty spool, in grams. (tare weight)", and it already takes precedence over
`filament.spool_weight` in Bambuddy's own tare resolution above. The write is simply missing.

### Expected Behavior

Either persist `core_weight` to the Spoolman spool's `spool_weight` (the field exists and is
already read with priority), or don't render an editable field in this mode.

### Steps to Reproduce

1. Spoolman mode enabled.
2. Inventory → any spool → Edit. Note the core weight field shows a value (e.g. 250).
3. Change it to a different value (e.g. 180) and save. No error is shown.
4. Reopen the editor: the field is back to the inherited value.
5. `curl -s http://<spoolman>/api/v1/spool/<id>` — `spool_weight` is unchanged (`null` if it was
   inheriting).
6. Now use the weigh action with a known gross weight: the remaining weight is computed from the
   old tare.

### Troubleshooting steps already taken

[...]

PR #2923, opened 2026-08-23
Take an RFID spool's core weight from the row that names it (issue #2909)

### LINKED ISSUE #2909: [Bug]: Spool-catalog core weight lookup has no matching logic — every Bambu RFID spool gets the "Plastic High Temp" weight
### Product

Bambuddy

### Area

Filament / AMS / Spoolman

### Bug Description

`services/spool_tag_matcher.py:139-145` looks up an empty-spool weight for a newly detected
Bambu RFID spool:

```python
core_weight = 250  # Default for Bambu Lab plastic spools
cat_result = await db.execute(
    select(SpoolCatalogEntry).where(SpoolCatalogEntry.name.ilike("Bambu Lab%")).limit(10)
)
for entry in cat_result.scalars().all():
    # Pick the best match (prefer exact, fallback to first Bambu Lab entry)
    core_weight = entry.weight
    break
```

The comment describes a "prefer exact, fallback to first" strategy, but **there is no matching
step** — the loop takes the first row and breaks. There is no `ORDER BY` either, so in practice
this is insertion order, and the seeded catalog puts these three Bambu Lab rows in this order:

```
Bambu Lab - Plastic High Temp   216
Bambu Lab - Plastic Low Temp    250
Bambu Lab - Plastic White       253
```

Every Bambu spool therefore gets **216 g**, the high-temperature spool's weight, regardless of
material. On my printer all three internally-created spools (PLA Basic, PLA Matte, PETG Basic)
came out with `core_weight=216`.

The `limit(10)` suggests the intent was to choose among candidates.

### Expected Behavior

Either pick the catalog entry that corresponds to the spool actually detected (material /
sub-brand aware), or fall back to the documented default of 250 rather than silently using the
first Bambu Lab row.

### Steps to Reproduce

1. Internal inventory mode (Spoolman disabled), default spool catalog.
2. Insert any Bambu RFID spool that is *not* a high-temperature spool — e.g. PLA Basic — into an
   AMS tray so it is auto-created.
3. Inspect the created spool: `core_weight = 216` (the `Bambu Lab - Plastic High Temp` value),
   not 250.
4. Repeat with a PETG spool — also 216.

- Read the running container's source, so the quoted code is from 1.2.6b1.
- Verified the resulting rows in the database: all three RFID-created spools carry
  `core_weight=216`.
- Confirmed the seeded catalog ordering by querying `spool_catalog` directly.
- Confirmed the blast radius: in Spoolman mode this value is not used for tare resolution
  (`spool.spool_weight ?? filament.spool_weight ?? 250` is), so the impact is limited to internal
  inventory mode.
- Searched open AND closed issues for spool catalog / core weight.

### Printer Model

X2D

### Bambuddy Version

### Printer Firmware Version

01.02.00.00

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

No support package. Debug logging is off on this host by policy and this leaves no log trace —
it is a silent wrong-value selection. Evidence is the source above plus the created rows.

### Additional Context

Low priority, and I'd understand closing it as "won't fix" if the internal catalog is on its way
out — but the mismatch between the comment and the code is the kind of thing that reads as
intentional to the next person touching it, so it seemed worth writing down.

Separately: for what it's worth, I measured my own Bambu spools while setting up an inventory
this week. A current-generation reusable spool is **207 g** bare, and the refill cardboard core
adds **43 g**, so the as-shipped tare is **250 g** — which matches both the
`Bambu Lab - Plastic Low Temp` catalog row and the `spool_weight: 250.0` that Spoolman's external
Bambu Lab entries all carry. Happy to contribute catalog corrections if useful.

ojimpo:
## Description

`create_spool_from_tray` looked the core weight up with a `Bambu Lab%` prefix query and took the first row it got back. There is no matching step and no `ORDER BY`, so the tare is whichever of the three Bambu Lab rows the database returns first — `Bambu Lab - Plastic High Temp` (216 g) on SQLite, nothing guaranteed on Postgres once the table has seen updates. The comment above the loop promised "prefer exact, fallback to first Bambu Lab entry"; only the fallback was ever written.

This selects `Bambu Lab - Plastic Low Temp` (250 g) by name, falls back to a 250 g constant when that row is missing, and records the row it used in `core_weight_catalog_id`.

### Why this is not just a display defect

`core_weight` is the tare in SpoolBuddy's weigh flow — `backend/app/api/routes/spoolbuddy.py:879`:

```python
net_filament = max(0, req.weight_grams - spool.core_weight)
spool.weight_used = max(0, spool.label_weight - net_filament)
```

A 34 g low tare makes every scale weighing of an RFID-created spool read 34 g more filament than is on it, and write `weight_used` 34 g low — inside the one feature whose whole purpose is an accurate weight. Remaining weight, low-stock alerts and the pre-print filament check all read that number afterwards.

The wiki already states the intended value: `docs/spoolbuddy/usage.md` — "Core weight (empty spool weight, default 250 g)". This makes the RFID path agree with it.

### `core_weight_catalog_id`

The `Spool(...)` constructor passed `core_weight` alone, so the column that exists to record which catalogue row supplied the weight stayed NULL. From my instance (Bambuddy `daily`, `APP_VERSION` 1.2.6b1, SQLite):

```
spool_catalog, Bambu Lab rows, by id
(23, 'Bambu Lab - Plastic High Temp', 216)
(24, 'Bambu Lab - Plastic Low Temp',  250)
(25, 'Bambu Lab - Plastic White',     253)

spool: id, material, subtype, colour, core_weight, core_weight_catalog_id, origin
(1, 'PLA',  'Basic', 'Black',       216, None, 'rfid_auto')
(2, 'PLA',  'Matte', 'Ivory White', 216, None, 'rfid_auto')
(3, 'PETG', 'Basic', None,          216, None, 'rfid_auto')
```

Three spools, three different filaments, one weight, no audit trail.

**Correction to what I first wrote here, from @maziggy's review:** I said the form's catalogue dropdown rendered empty. It does not, and that is worse. `AdditionalSection.tsx` falls back to `matchingEntries[0].name` for display and auto-selects when exactly one row matches the weight — and `DEFAULT_SPOOL_CATALOG` has exactly one 216 g row (I checked: one row at 216, one at 250, out of 91). So the dropdown reads "Bambu Lab - Plastic High Temp", and opening the form and saving any unrelated field writes that id to the database. The arbitrary pick gets laundered into what looks like a deliberate user selection while `core_weight` stays wrong. After this change the same mechanism selects Low Temp and becomes harmless.

### Scope

This is the internal-inventory path only. `create_spool_from_tray` is called from the AMS handler behind `if not _spoolman_on` (`backend/app/main.py:2228`), so a Spoolman-mode install never reaches this lookup — `_map_spoolman_spool` takes `core_weight` from Spoolman's own `spool_weight`, defaulting to 250 (`_spoolman_helpers.py:270`), and SpoolBuddy's weigh flow only falls through to that after missing the local DB. The three spools above predate this instance switching to Spoolman mode.

Which also means this does not touch #2908. That one is per-spool core weight not persisting through the Spoolman write path; this one is which catalogue row the local creation path reads. Different code, no shared fix.

### Notes on the shape

- **Falls back to the constant, not to another catalogue row.** Falling back to a row is how the arbitrary pick started. A renamed or deleted entry now gives 250 g and a NULL catalogue id rather than silently selecting High Temp again.
- **Matches on name, so an edited row wins.** The catalogue is user-editable; someone who has weighed their own empty spool gets their number instead of the shipped default.
- **Keeps `order_by(...).limit(1)`** — the same deterministic tiebreak the colour catalogue lookup fifteen lines above already carries. Not cosmetic: without `limit(1)` a duplicate-name row makes `scalar_one_or_none()` raise `MultipleResultsFound` and the AMS read fails outright. Now covered by a test.
- **Case-insensitive comparison**, matching how the colour lookup directly above compares (`func.upper(...)`), so a row edited to a different capitalisation still matches. Now covered by a test.
- **No material awareness**, as agreed on the issue. It is buildable here — `material` and `tray_info_idx` are both in scope — but the only row it could select is High Temp, and 216 g is not a number I measured.

### Backfill — follow-up PR

**Not in this PR.** Existing `data_origin='rfid_auto'` rows keep whatever they have; the three above stay at 216 until corrected.

I originally offered to guard a backfill on `core_weight_catalog_id IS NULL`, reasoning that a migration cannot tell a wrong 216 from a value a user has fixed by hand. @maziggy's review shows that guard misses in both directions, and the laundering above is why: it skips any spool whose form was ever opened and saved — exactly the ones that have already put a wrong tare through a weighing — while a user who hand-typed a corrected weight matching no catalogue row has their id cleared to NULL by the same effect, so the guard would overwrite the one case it was written to protect.

Keyed on the value instead — `data_origin = 'rfid_auto' AND core_weight = 216` — it selects the wrong spools and only the wrong spools. Coming as a separate PR.

## Related Issue

Fixes #2909

## Documentation

**Pick one**:
- [x] No docs update required — reason: bug fix with no new setting, endpoint or UI element. The corrected tare is the 250 g default the wiki already documents (`docs/spoolbuddy/usage.md`), so the docs describe the fixed behaviour as they stand.

## Type of Change

## Changes Made

- `backend/app/services/spool_tag_matcher.py` — replace the `Bambu Lab%` prefix query with an exact-name lookup of `Bambu Lab - Plastic Low Temp`, ordered and limited; fall back to a named 250 g constant when the row is absent
- Same file — pass `core_weight_catalog_id` to the `Spool(...)` constructor so the audit column and the spool form's dropdown reflect the row that was used
- `backend/tests/unit/services/test_spool_tag_matcher.py` — new "core weight catalog lookup (#2909)" section, five tests

## Screenshots

Not applicable — no frontend change. The spool form's catalogue dropdown starts showing the row it was always meant to show, but nothing in the frontend was touched.

## Testing

- [x] I have tested with my printer model: **X2D** (the three spools above were auto-created by it from AMS RFID reads)

The fixture seeds the three Bambu Lab rows in `DEFAULT_SPOOL_CATALOG` order — High Temp first — so the tests reproduce the arbitrary pick rather than asserting against a constructed one. Four of the five fail on `dev` before the change:

```
assert 216 == 250    test_core_weight_matches_low_temp_row_not_first_bambu_row
assert None == 2     test_core_weight_records_the_catalog_row_it_used
assert 216 == 244    test_core_weight_honours_a_user_edited_catalog_row
assert 216 == 250    test_core_weight_falls_back_to_default_when_row_is_missing
```

The fifth (`test_core_weight_falls_back_on_empty_catalog`) passes either way; it pins the fresh-install path so a later change cannot make an unseeded catalogue raise.

Two more added from review, both for behaviour called out as deliberate in the notes above and neither previously covered:

```
sqlalchemy.exc.MultipleResultsFound   test_core_weight_is_deterministic_when_the_name_is_duplicated
                                      (verified by removing .order_by(...).limit(1))
                                      test_core_weight_matches_the_row_name_case_insensitively
```

After the change: `68 passed` for `test_spool_tag_matcher.py`. `ruff check` and `ruff format --check` clean across the backend. Run in the repo's own `backend-test` image (no local venv — `python3-venv` is not installed on this host).

## Checklist


maziggy (changes requested):
It is the fix I asked for, and the reasoning in the description
is better than the issue it closes -- particularly falling back to the constant
rather than to another row, which is the part that stops this recurring.

One correction, and it lands on your backfill question rather than the diff.

The "catalog dropdown rendered empty" symptom is not what happens, and I was the
one who told you it was -- my issue comment said the same thing. SpoolWeightPicker
falls back to the first catalog row matching the weight, and auto-selects when
exactly one row matches. DEFAULT_SPOOL_CATALOG has exactly one 216 g row. So the
dropdown shows "Bambu Lab - Plastic High Temp", and opening the form and saving
any unrelated field writes core_weight_catalog_id = <High Temp id> to the
database. The arbitrary pick gets laundered into what looks like a deliberate
user selection while core_weight stays wrong. After your change the same
mechanism writes the correct row, so it becomes harmless.

Which is why the backfill guard you offered -- core_weight_catalog_id IS NULL --
would miss in both directions. It skips any spool whose form was ever opened and
saved, and those are exactly the ones that have already put a wrong tare through
a weighing. And it overwrites a user who hand-typed a corrected weight matching
no catalog row, because that same effect clears their id to NULL. That is the
case you wrote the guard to protect.

Key it on the value instead: data_origin = 'rfid_auto' AND core_weight = 216.
That selects the wrong spools and only the wrong spools. Yes please, if you want
to add it -- separate PR is fine.

Two small things:

- The case-insensitive match and the order_by(id).limit(1) tiebreak are both
  called out as deliberate in your notes but neither has a test. The second is
  worth one: without limit(1), scalar_one_or_none() raises on a duplicate-name
  row, and nothing would catch that.
- Scope is one caller short. POST /inventory/spools/from-slot also calls
  create_spool_from_tray and is not gated on Spoolman mode server-side; the UI
  routes Spoolman installs elsewhere, so your conclusion holds. No action.

I will add the changelog entry on merge. Thanks for the measurements correction
on the issue as well -- that is what kept 216 and 253 out of this.

ojimpo:
Both tests added in 839cf2c2, and thank you for the correction — I had the form's
behaviour backwards and it changes the follow-up.

**Tiebreak.** You were right that it needed one, and it is worse than I had it in
the notes. Without `order_by(id).limit(1)` a duplicate-name row does not just make
the pick arbitrary again — `scalar_one_or_none()` raises `MultipleResultsFound`, so
the AMS read fails outright. I verified that by removing the clause and watching the
new test raise rather than mis-assert:

```
sqlalchemy.exc.MultipleResultsFound: Multiple rows were found when one or none was required
FAILED test_core_weight_is_deterministic_when_the_name_is_duplicated
```

The test asserts the lowest id wins. Case-insensitivity has one too, seeding the row
as `bambu lab - plastic low temp` with an edited weight so it fails if the match ever
tightens back to exact case.

**The laundering.** I checked `AdditionalSection.tsx` against your description and it
is exactly as you say: `displayValue` falls back to `matchingEntries[0].name`, and the
effect auto-selects when `matches.length === 1`. `DEFAULT_SPOOL_CATALOG` has exactly
one 216 g row — one at 216, one at 250, out of 91 — so both conditions hold and the
form writes `core_weight_catalog_id = <High Temp id>` on the next save of any field.

That is a better description of the defect than mine was, because it explains why the
column being NULL is the *less* damaged state. I have corrected that paragraph in the
PR description rather than leaving it standing.

**Backfill.** Agreed, and I had it backwards for the reason above. `IS NULL` skips
every spool whose form has been opened — the ones that have already put a wrong tare
through a weighing — and clearing to NULL is precisely what happens to a user who
hand-types a corrected weight matching no row, so the guard would have overwritten the
case it existed to protect. `data_origin = 'rfid_auto' AND core_weight = 216` selects
the wrong spools and only the wrong spools. Coming as its own PR.

**`/inventory/spools/from-slot`.** Noted, thank you — I had traced only the MQTT
handler. Understood as no action here.

ojimpo:
Backfill is up as #2925, keyed on `data_origin = 'rfid_auto' AND core_weight = 216` as you specified. The three cases from your review — never-opened, laundered-id, hand-corrected-weight — are one test each, so the guard that would have got it wrong is pinned as well as the one that gets it right.

ojimpo:
Rebased onto current `dev` in 409bd147 — b38022ec landed in the middle of
`create_spool_from_tray`, so the two edits met head-on. Both are kept: your
`effect_type` fallback runs where you placed it, and the catalogue lookup below
it is this PR's name-matched version rather than the prefix query.

I have closed #2925. Your repair is the better one and I said why over there —
it fixes `weight_used`, which I had not touched at all, and it selects on any of
the other Bambu rows rather than the literal 216, which is the consistent line I
had failed to draw.

The two do not interact: the migration selects a `core_weight` matching one of
the *other* Bambu rows, so a spool this PR creates correctly at 250 g is never a
candidate for it.

Backend suite on the merge: 10672 passed, 22 skipped. `ruff check` and
`ruff format --check` clean.

maziggy (approved):

Issue #2907, opened 2026-08-22
[Bug]: Spoolman filament auto-match ignores the AMS sub-brand — PLA Matte spools are linked to the PLA Basic filament of the same colour


ojimpo:
### Product

Bambuddy

### Area

Filament / AMS / Spoolman

### Bug Description

When a Bambu RFID spool is first seen in an AMS tray, `_find_or_create_filament`
(`services/spoolman.py`) matches it to a Spoolman filament using **material + `color_hex` only**.
The AMS sub-brand (`tray_sub_brands`, e.g. `PLA Matte`, `PETG Translucent`) is used only as an
exact-name tie-break:

```python
chosen = next(
    (f for f in bambu_candidates if (f.get("name") or "").strip().lower() == sub_brand),
    bambu_candidates[0],
)
```

Bambu Lab's entries in the external library are named by **colour** (`Jade White`, `Black`,
`Translucent Teal`), while `tray_sub_brands` is a **product line** (`PLA Matte`). Those two never
match, so for Bambu Lab spools the tie-break always falls through to `bambu_candidates[0]`.

Because Bambu reuses the same hex across product lines, this reliably picks the wrong variant:

| Physical spool (from RFID) | Filament created / linked |
|---|---|
| PLA Matte — Ivory White (`FFFFFF`) | PLA Basic **"Jade White"** |
| PLA Matte — Charcoal (`000000`) | PLA Basic **"Black"** |
| PETG Translucent — clear | **"Black"** (PETG) — see note |

Three out of seven genuine spools were mislinked on first load. The spool ends up carrying the
wrong density, wrong temperature defaults and the wrong product identity, which then feeds cost
and length calculations.

Note on the third row: that spool is colourless clear, so something also went wrong with the
colour itself before the match ran. I'm still confirming that part (it looks like the alpha byte
of `tray_color` is being dropped) and will file it separately if it holds up — the two defects
compound but have different causes. The variant-matching defect described here reproduces on its
own with the PLA Matte cases above, which need no colour handling at all.

**The data needed to disambiguate is already in Bambuddy.** The seeded `color_catalog` table
carries the sub-brand in its `material` column:

| manufacturer | color_name | hex_color | material |
|---|---|---|---|
| Bambu Lab | Black | #000000 | **PLA Basic** |
| Bambu Lab | **Charcoal** | #000000 | **PLA Matte** |
| Bambu Lab | Jade White | #FFFFFF | **PLA Basic** |
[...]

Issue #1227, opened 2026-05-07
[Bug]: New Filament appears as wrong colour


elit3ge:
### Component

Bambuddy

### Bug Description

When I put a new Ivory white roll of filament in, Bambuddy adds it as a Jade White. It's happened for the last 4 rolls.

### Expected Behavior

Appear as Ivory White not Jade

### Steps to Reproduce

1. Insert new roll of filament (Ivory White) into AMS
2. Check what new roll has been added.

### Printer Model

P2S

### Bambuddy Version

Dev

### Printer Firmware Version

01.02.00.00

### Installation Method

Manual (git clone)

### Operating System

Linux (Ubuntu/Debian)

### Screenshots

<img width="818" height="833" alt="Image" src="https://github.com/user-attachments/assets/6e565d71-2bb5-490b-b285-bf27ab2be680" />

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I am using the latest version of Bambuddy
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled

maziggy:
Available/Fixed in branch dev and available with the next release or daily build. Please let me know if it works for you.

-----
If you find Bambuddy useful, please consider giving it a ⭐ on [GitHub](https://github.com/maziggy/bambuddy) — it helps others discover the project!

PR #2924, opened 2026-08-23
Let a clear spool stay clear on the way to Spoolman (issue #2912)

### LINKED ISSUE #2912: [Bug]: Clear filament syncs to Spoolman as pure black — the #1545 alpha fix covered the built-in inventory only
### Product

Bambuddy

### Area

Filament / AMS / Spoolman

### Bug Description

A Bambu **PETG Translucent (colourless clear)** spool loaded into an AMS tray was auto-created in
Spoolman as a **black** filament:

```
filament #8  name="Black"  material=PETG  color_hex=000000  density=1.25
```

The AMS reports the tray correctly. Raw value straight off `GET /api/v1/printers/1/status`:

```
AMS 0 / tray 0   tray_color="00000000"   tray_type="PETG"   tray_sub_brands="PETG Translucent"
                 tray_info_idx="GFG01"   tray_id_name="G01-C0"
                 tag_uid="CC9954E400000100"
                 tray_uuid="2482EBF77B824E6B8A5FD896DA5580B5"
```

and the matching create in the log:

```
2026-08-22 17:37:09,708 INFO [backend.app.services.spoolman] [-] Creating new spool in Spoolman for PETG Translucent (tag: 2482EBF77B824E6B...)
```

The Spoolman sync path drops the alpha byte, so `00000000` becomes `000000` — indistinguishable
from opaque black:

| location (1.2.6b1) | code |
|---|---|
| `services/spoolman.py:1204` | `color_hex = tray.tray_color[:6]  # Strip alpha channel` (`_find_or_create_filament`, the Bambu-RFID path) |
| `services/spoolman.py:1154` | `color_hex=tray.tray_color[:6]` (non-Bambu RFID path) |
| `services/spoolman.py:354` | `color_hex = color_hex[:6] ...  # Strip alpha channel if present` |

For contrast, the opaque spools in the same AMS report a real alpha byte, and empty trays report
an empty string rather than `00000000` — so `00000000` is a genuine reported colour here, not a
sentinel:

```
AMS 0 / tray 0   tray_color="00000000"   PETG Translucent   (clear)
AMS 0 / tray 1   tray_color="7F7E83FF"   PETG Basic         (grey)
AMS 0 / tray 2   tray_color=""           (empty slot)
AMS 0 / tray 3   tray_color=""           (empty slot)
```

**This is the same defect #1545 fixed for the built-in inventory, and the fix commit made it
worse for Spoolman mode.**

`63233495` (`fix(inventory): support transparent / clear filament end-to-end (#1545)`) removed the
cream rewrite from `parse_ams_tray`:

```diff
-        if tray_color == "00000000":
-            tray_color = "F5E6D3FF"  # Light cream/natural color
+        # ... we pass `00000000` through verbatim — the inventory row keeps the
+        # AMS-reported colour and the frontend resolves the name to "Clear" ...
```

That rewrite came from PR #191 (*"Fix Spoolman sync for transparent/natural filament spools"*,
fixes #190), and `parse_ams_tray` is shared by both inventory modes — it feeds the Spoolman sync
sites at `api/routes/spoolman.py:310`, `:527`, `:1202` and `main.py:2512`. The same commit added
an explicit alpha guard on the **built-in** side (`services/spool_tag_matcher.py:103-110`):

```python
# Transparent filament (#1545): the AMS reports alpha=00 for clear spools.
# Skip the catalog lookup — the catalog only stores RGB so 000000 would
# resolve to "Black" ... Store "Clear" directly
if rgba and len(rgba) == 8 and rgba[6:8].lower() == "00":
    color_name = "Clear"
```

No equivalent guard exists on the Spoolman side. So for Spoolman users the history reads:

| | clear spool in Spoolman mode |
|---|---|
| before #191 | skipped as an invalid colour (#190) |
| #191 (2026-01-30) | `F5E6D3` — cream, not accurate but not wrong-looking |
| **#1545 (2026-05-29) → today** | **`000000` — pure black** |

Spoolman itself can store the alpha byte. From its own OpenAPI schema (0.26.1):

```json
"color_hex": {"type": "string", "minLength": 6, "maxLength": 8,
  "description": "Hexadecimal color code of the filament, e.g. FF0000 for red.
                  Supports alpha channel at the end. ..."}
```

So the value Bambuddy is discarding is one the backend would have accepted verbatim.

Knock-on effect: because the colour collapses to `000000`, the external-catalogue lookup then
matches it against the Bambu PETG entries at that RGB (`Black` / `HF Black` / `PETG-CF Black`) and
names the filament **"Black"**. That second step is the sub-brand matching defect I filed
separately; the two are independent (that one reproduces on opaque PLA Matte spools with no
colour handling involved), but together a clear spool ends up as a black one with a black name.

### Expected Behavior

A clear/translucent spool should not be recorded in Spoolman as opaque black. Spoolman accepts
`RRGGBBAA`, so the straightforward fix is to pass the AMS value through instead of truncating it,
matching what #1545 already does for the built-in inventory.

### Steps to Reproduce

1. Spoolman mode, `auto_add_unknown_rfid` on (default).
2. Load a Bambu **PETG Translucent (clear)** spool into an AMS tray.
3. `GET /api/v1/printers/{id}/status` → the tray reports `tray_color: "00000000"`.
4. Check the filament Bambuddy created in Spoolman → `color_hex=000000`, name `Black`.

Step 4 is what I observed at the time of the sync; I have since corrected that record by hand, so
the current state of my Spoolman instance no longer shows it.

- Captured the raw `tray_color` from `/printers/{id}/status` rather than inferring it, to confirm
  the AMS really reports `00000000` and not e.g. `FFFFFF00`.
- Confirmed empty trays report `""`, so `00000000` is not being used as an "empty" sentinel on
  this printer.
- Searched existing issues for transparent / translucent / alpha / `tray_color` / `color_hex` and
  found #1545, #190/#191, #1394 and #1154. See the Bug Description for how this differs.
- Verified against Spoolman's OpenAPI schema that `color_hex` accepts 8 characters.

### Printer Model

X2D

### Bambuddy Version

1.2.6b1 (daily channel, `ghcr.io/maziggy/bambuddy:daily`, digest `sha256:9b4f52b9…`, pulled
2026-08-22)

### Printer Firmware Version

01.02.00.00 (AMS `n3f`, sw_ver 05.00.22.19)

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

No support package. Debug logging is off on this instance as a matter of policy, and it was not
needed here — the raw AMS value is available at INFO level via `/printers/{id}/status`, and the
create is logged at INFO. Both are quoted verbatim above. Happy to produce a debug-logged repro if
useful.

### Additional Context

**Relationship to past issues (so this doesn't re-open a closed problem):**

- **#1545** — same symptom, fixed for the **built-in inventory** only; the closing comment scopes
  it to "Built-in inventory now supports transparent end-to-end", and the reporter stated they
  were not using Spoolman. This report is the Spoolman half of that.
- **#190 / PR #191** — the original Spoolman-mode fix, whose cream fallback #1545 removed. Not a
  request to restore cream: cream was a workaround for a swatch renderer that couldn't draw
  alpha, which is no longer the constraint.
- **#2907** (sub-brand matching) — independent; see above.
- **#1394** (missing PLA Translucent in the filament list) — a catalogue-content request, not the
  sync path.

**I'm happy to open the PR.** Which shape do you prefer?

1. Pass the full `RRGGBBAA` through to Spoolman at the three sites above, and compare on the
   RGB prefix when matching (smallest change; keeps the AMS value intact, which is what #1545
   chose for the built-in inventory).
2. Mirror `spool_tag_matcher`'s guard: detect `alpha == 00` in the Spoolman path and handle it
   explicitly (e.g. keep `00000000` and skip the external-catalogue colour match so it can't
   resolve to "Black").

Option 1 looks closer to the intent of #1545, but option 2 is more surgical if you'd rather not
change what Bambuddy writes into `color_hex` for every spool.

ojimpo:
## Description

Bambuddy validated an alpha byte it never intended to store. `_validate_rgba` accepts six or eight characters and raises a message saying so, both `rgba` fields are declared `max_length=8` advertising `RRGGBBAA` — and then every write truncated to six and every read rejected anything that was not exactly six. A clear spool reads as `00000000` from the AMS, so it went in as `000000` and came back out as opaque black. Entering "fully transparent" by hand in the UI landed on the same place.

This stops truncating on write, accepts eight on read, and normalises every colour comparison to the RGB prefix so widening the stored value cannot mint duplicates.

### The four things that had to move together

**1. Write — `create_filament` is the chokepoint.** Every create funnels through it, so the caller-side `[:6]` slices were redundant on top of it and are gone. Eight characters are written **only when the alpha byte is not `FF`**. Blanket pass-through would rewrite the `color_hex` of every opaque spool on its next touch, churning records in people's Spoolman for no benefit and widening the duplicate-match exposure from clear spools to the whole inventory. Restricting it to genuinely translucent spools leaves existing data byte-identical.

**2. Read — the regex and the `+ "FF"` assembly.** `_COLOR_HEX_RE` now accepts 6 or 8, and `rgba` only appends the opaque byte when the value is not already carrying one. Fixing the writes alone would have turned clear spools grey rather than black.

**3. Comparisons — the upgrade hazard, and the reason this is not just "stop truncating".** Colours are compared with full-string equality, and **every filament already in a user's Spoolman is stored six characters**. Had the compared value grown to eight, `_find_or_create_filament` would have fallen through on every spool and the next AMS sync would have minted a duplicate filament for each one. Comparisons now go through `color_match_key`, which is **the shape the value would be stored as** — so two colours match exactly when storing them would produce the same value:

| value | key | consequence |
|---|---|---|
| `000000` | `000000` | existing six-character data |
| `000000FF` | `000000` | still matches it — the upgrade guard |
| `00000000` | `00000000` | its own filament, in both directions |

See "The trade in the key" below for why the third row is not the RGB prefix.

**4. The edit route's change detection needed a third answer.** Neither option in (1)–(3) works at `spoolman_inventory.py`:

| comparison | opaque no-op edit | alpha-only edit |
|---|---|---|
| raw strings | PATCHes the filament every time (`FF0000` vs incoming `FF0000FF`) | lands |
| bare RGB prefix | correct | **silently dropped** — the edit never reaches Spoolman |
| stored shape (this PR) | correct | lands |

So it compares what each side *would be stored as*, which is false only when the stored value would actually differ. Both directions are pinned by tests.

### The trade in the key

I first had `color_match_key` on the bare RGB prefix and deferred clear-finds-opaque to #2907. @maziggy's review shows the reverse direction does not wait for it, because **before this PR no 8-character filament could exist at all**. With one stored as `00000000`, an opaque `000000FF` tray matches it, then renders through `getSwatchStyle` as the transparency checkerboard and is named Clear by `resolveSpoolColorName`. Whichever roll syncs first decides and the other is mislabelled — a new user-visible defect introduced by this PR, not an old one carried forward.

Folding the alpha into the key when it is not `FF` settles both directions without touching colour-name resolution, so I took it.

**The cost, stated plainly: a user already carrying a wrongly-opaque filament from this bug gets a second one the next time that roll is auto-added.** They end with two records — the old wrong black one still linked to the old spool, and a new correct clear one. I think that is the better trade, for two reasons. The duplicate is bounded and self-correcting: one per affected roll, and the new record is the right one from then on. The alternative is unbounded in the other direction — every black roll on an instance that has a clear filament of the same RGB is a coin flip on which name and swatch it inherits, and that keeps happening. It is also the failure that is hard to diagnose from the UI, because a black spool showing a checkerboard reads as a rendering bug rather than a match bug.

The upgrade guard is untouched by this: `000000FF` still keys to `000000`, so nothing that exists today stops matching.

### What this deliberately does not do

**It does not give a clear spool its own filament when the internal search finds an opaque one of the same RGB — that case is now handled**, per the above. What remains scoped out is the **sub-brand** short-circuit you identified on #2907 (`spoolman.py:1257-1266`): a PLA Matte tray still matches a PLA Basic filament of the same colour, because separating those needs the colour-name resolution that PR is for.

**It leaves `SkippedSpool.color` alone** (`api/routes/spoolman.py:386/397/600/611`). Those four `[:6]` slices feed a "skipped" report for display and never reach Spoolman, so widening them changes nothing except the diff.

### Also caught, not in the original report

`spoolman_inventory.py:711` builds the edit form's default `rgba` as `cur_color + "FF"` from the current filament. With eight-character values now possible that produces a ten-character string, so it only appends when the value is not already carrying an alpha.

`SpoolFormModal.tsx:452` is the frontend twin of the read-side regex — `/^[0-9A-F]{6}$/` then append `FF`, so picking a now-storable clear filament out of the Spoolman catalogue prefilled the form with `808080FF`. Found by @maziggy in review; it was unreachable while Bambuddy never wrote eight characters and this PR makes it reachable. Two frontend tests, one per branch.

`_create_filament_from_external` (`spoolman.py:1315`) passed the alpha-preserving value as the *default* of `external.get("color_hex", ...)`, so on a fresh Spoolman with the external library reachable a clear tray still matched SpoolmanDB's opaque "PLA Basic Black" on prefix and was created at `000000` — the reported path, still reproducing. Also found in review. The alpha-aware key closes it: the candidate list is built with the same key, so an opaque external entry is no longer a candidate for a translucent tray and the filament is created from the tray data instead. Pinned by `test_clear_tray_does_not_take_a_same_rgb_external_entry`, which fails on the prefix-only key.

### On the `spool_tag_matcher.py` parity note

Thank you for pointing at `spool_tag_matcher.py:95-99` — I had not spotted that the built-in side already filters the colour catalog by `tray_sub_brands` citing #1227. I have used it as the reference for #2907 rather than designing that resolution from scratch, and I have taken the standing expectation about parity between the two modes as read.

## Related Issue

Fixes #2912

## Documentation

**Pick one**:
- [x] No docs update required — reason: bug fix with no new setting, endpoint or UI element. The API contract is unchanged; `rgba` already documented and validated `RRGGBBAA`, and this makes the storage layer honour what it advertises.

## Type of Change

## Changes Made

- `backend/app/utils/color_utils.py` — new `spoolman_color_hex()` (what to store: 8 chars only for a non-`FF` alpha) and `color_match_key()` (RGB prefix, for comparisons)
- `backend/app/services/spoolman.py` — `create_filament` stores the normalised shape instead of truncating; `find_or_create_filament` splits the match key from the value it creates with; `_find_or_create_filament` does the same and both of its comparison loops go through `color_match_key`; the non-Bambu RFID path and the external-entry fallback pass the tray colour through instead of slicing it
- `backend/app/api/routes/_spoolman_helpers.py` — `_COLOR_HEX_RE` accepts 6 or 8; `rgba` no longer appends `FF` to a value that already has an alpha
- `backend/app/api/routes/spoolman_inventory.py` — create and edit routes store the normalised shape; the edit form's default `rgba` no longer double-appends; `metadata_unchanged` compares stored shapes
- Tests in four files (below)

## Screenshots

Not applicable — no frontend change. A translucent spool's swatch starts rendering with its real alpha because the API now returns it, but nothing in the frontend was touched.

## Testing

- [x] I have tested with my printer model: **X2D** (the `tray_color="00000000"` in the report is a raw MQTT read from it)

**Ten tests fail on `dev` before the change**, for the reasons in the report and the review:

```
assert '000000'   == '00000000'   test_create_filament_stores_alpha_for_a_translucent_spool
assert '000000'   == '00000000'   test_clear_tray_creates_a_translucent_filament
assert '000000'   == '00000000'   test_find_or_create_filament_matches_on_prefix_and_creates_with_alpha
assert '808080FF' == 'FF000080'   test_eight_char_color_hex_is_read_back_with_its_alpha
assert '808080FF' == '00000000'   test_fully_transparent_color_hex_survives_the_read
assert '000000'   == '00000000'   test_create_spool_keeps_a_clear_colour_translucent
assert '808080FF' == 'FF000080'   test_update_alpha_only_edit_reaches_the_filament
assert '000000'   == '00000000'   test_clear_tray_does_not_attach_to_the_black_filament
assert '00000000' == '000000'     test_black_tray_does_not_attach_to_a_clear_filament
assert '000000'   == '00000000'   test_clear_tray_does_not_take_a_same_rgb_external_entry
```

The last three fail on the prefix-only key as well, which is how I checked that the change in (3) is the thing doing the work rather than the widened write. Frontend, against `dev`:

```
AssertionError: expected '808080FF' to be '00000000'
  SpoolFormModal > prefills a translucent filament with its own alpha (#2912)
```

The remaining new tests **pass both before and after on purpose** — they pin what must *not* change: an opaque colour stays six characters, an 8-char tray colour still matches an existing 6-char filament (the duplicate-minting guard), and a round-tripped opaque `rgba` is still a no-op edit. Those are the regressions a naive "stop truncating" would introduce, so they are worth having even though they are green on `dev`.

**One existing test was changed, deliberately.** `test_eight_char_color_hex_falls_back` asserted that an 8-char `color_hex` greys out, with the comment "Only 6-char hex is valid from Spoolman". Spoolman stores what it is given — I verified that by writing `1234ABCD` through Spoolman's own API before filing. It now pins the value coming back with its alpha, and I added a 7-character case so the malformed path stays covered.

Full suites after the change: backend **10655 passed, 22 skipped**; frontend **3230 passed** across 239 files, with `tsc`, `npm run lint` and the locale-parity check clean. `ruff check` and `ruff format --check` clean. Run in the repo's own `backend-test` / `frontend-test` images.

## Checklist


maziggy (changes requested):
Reviewed and tested locally — full backend suite 10803 passed, ruff check and format clean. The
decomposition here is right, and the duplicate-minting guard in (3) is the part that makes this
safe to ship: keeping comparisons on the RGB prefix while widening the write is exactly the call.
Restricting the widened write to a non-FF alpha is right too — blanket pass-through would have
churned every opaque record in people's Spoolman. Three things before merge.

1. The reported path still stores opaque when SpoolmanDB has a same-RGB Bambu entry.
`_create_filament_from_external` (spoolman.py:1315) passes the alpha-preserving value as the
*default* of `external.get("color_hex", ...)`, so it only fires when the external entry has no
colour at all — and every SpoolmanDB entry has one. The candidate list is built with
`color_match_key`, so a clear tray reading 00000000 matches the external "PLA Basic Black" on
prefix 000000 and the new filament is created at 000000. I confirmed it:

    created with color_hex = 000000

Unlike the internal short-circuit you scoped out, nothing is preserved by that — the record is
new. On a fresh Spoolman with the external library reachable, the issue still reproduces.

2. Prefix matching now runs in both directions, and the new direction is user-visible. 00000000
and 000000FF are the same RGB by definition, so clear PLA and black PLA are one filament. You
documented clear-finds-opaque and deferred it; the reverse is new, because before this PR no
8-char filament could exist. With a filament stored 00000000, an opaque 000000FF tray matches it,
and that black roll then renders through `getSwatchStyle` as the transparency checkerboard and is
named Clear by `resolveSpoolColorName`. Whichever roll syncs first decides, and the other is
mislabelled. `color_match_key` can settle both directions without touching colour-name resolution,
so it needn't wait for #2907 — fold the alpha into the key only when it is not FF:

    "000000"   -> "000000"     existing data
    "000000FF" -> "000000"     still matches it, upgrade guard intact
    "00000000" -> "00000000"   its own filament, never conflated

The cost is real and worth weighing in the PR either way: a user already carrying a wrongly-opaque
filament from this bug gets a second one the next time that roll is auto-added. Your call which
trade you prefer, but the inverse case is a reason to decide it here rather than next time.

3. One frontend guard is the twin of the read-side regex you fixed. SpoolFormModal.tsx:452 does
`/^[0-9A-F]{6}$/.test(rawHex) ? rawHex : '808080'` and then appends FF, so picking the now-storable
clear filament out of the Spoolman catalogue picker prefills 808080FF. It was unreachable while
Bambuddy never wrote 8 characters; this PR makes it reachable. Everything else on the frontend
already handles alpha, so this is the only gap I found.

Minor: the `spoolman_color_hex` docstring says a malformed value is reported rather than reshaped,
but that only holds below 6 characters — a 7-char value is truncated to 6 while the read side greys
the same value out. Not reachable through `_validate_rgba`, so it is the docstring, not the code.

The rest I checked and it holds: `_normalize_color_for_id`, `_spool_color_to_hex` and the label
renderer all already accept 8 characters, the `metadata_unchanged` table is correct with both rows
pinned by tests, and the four `SkippedSpool.color` slices are display-only and right to leave.
Rewriting `test_eight_char_color_hex_falls_back` rather than deleting it was the right way to
handle that one.

ojimpo:
All three addressed in 1d867de8. Two of them were mine to have caught, and the
second one is the useful lesson.

**(2) — taken, with the trade.** You are right that the reverse direction does not
wait for #2907, and I had missed why: before this PR no 8-character filament could
exist, so an opaque tray matching a clear filament is a defect this PR *creates*
rather than one it carries forward. `color_match_key` is now the shape the value
would be stored as, so two colours match exactly when storing them would produce
the same value — same table you gave.

I chose to accept the duplicate. The cost is bounded and self-correcting: one extra
filament per affected roll, and the new record is the right one from then on. The
inverse is unbounded — every black roll on an instance that has a same-RGB clear
filament is a coin flip, and it keeps happening. It is also the one that cannot be
diagnosed from the UI: a black spool rendering as a checkerboard reads as a display
bug, not a match bug. Written up in the description under "The trade in the key"
rather than left as a commit message.

**(1) — and this closed it.** You are right that it was still reproducing, and the
`.get` default is the reason. Worth saying where that came from: I had it as `or`
originally and changed it to `external.get("color_hex", <fallback>)` on #2923's
review, on the grounds that preserving `.get`'s semantics kept the diff easier to
read. That was the wrong axis to decide it on — what mattered was what the call
actually returns, and a SpoolmanDB entry always has a colour, so the fallback was
dead. The alpha-aware key from (2) removes the opaque external entry from the
candidate list, so the tray data is used. `test_clear_tray_does_not_take_a_same_rgb_external_entry`
fails on the prefix-only key, which is how I checked (2) is what does the work.

**(3) — fixed, with both branches tested.** Confirmed the symptom by reverting:

```
AssertionError: expected '808080FF' to be '00000000'
```

**Docstring** — corrected rather than the code, as you said. It now says what
actually happens to each length and notes that neither the sub-6 nor the 7-char
path is reachable through `_validate_rgba`.

Ten backend tests fail on `dev` before the change now, plus the frontend one.
Full suites after: backend 10655 passed / 22 skipped, frontend 3230 passed across
239 files, `tsc` and `npm run lint` clean and locales in parity.

The wider point I am taking from this: widening what the system can *store* means
auditing how existing code reads the new state, not only checking that it still
reads the old one. I had done the second and stopped.

ojimpo:
Cross-reference: #2944 (issue #2907) touches the same two comparison lines in `_find_or_create_filament`, so these two will conflict on whichever lands second. Both are based on `dev`.

They are independent in behaviour — this one decides which colours count as the same colour, #2944 decides which filament a colour belongs to — so the merge is a textual one either way. Say which you want first and I will rebase the other and re-run the suite before you look at it.

maziggy (changes requested):
Round two. All three are closed, and (2) landed the way I'd have taken it — the trade
you wrote up under "The trade in the key" is the right one and stating the cost in the
description rather than burying it in a commit message is the right call. Verified on
94cd48af: ruff clean, backend 10887 passed, frontend 3254 across 240 files, lint and
build clean.

Two things before merge, one optional.

1. The round-one vocabulary is still on the code, and it now describes the inverse of
what the code does. Three places say the match key drops alpha:

    spoolman.py:763   "`color` is the match key and stays on the RGB prefix"
    spoolman.py:1258  "Stored value may carry alpha; the match key never does (#2912)."
    test_spoolman_service.py  test_find_or_create_filament_matches_on_prefix_and_creates_with_alpha
                              "the match key drops alpha, the value handed to
                               create_filament keeps it"

That is precisely the invariant round two exists to overturn, and it is the one a future
reader must not get wrong — restoring prefix matching on the strength of those comments
reintroduces the mislabelling in both directions.

The mechanical corollary is the same root cause. `color_match_key` is now
`spoolman_color_hex(x) or ""` by definition, so `color`/`stored_color` (765-766) and
`color_hex`/`color_upper` (1259-1261) hold the same string for every input that reaches
them — they differ only for a falsy or whitespace-only value. `cur_rgba` at
spoolman_inventory.py:714 is the same shape of leftover: appending FF and then
normalising at 743 lands on the identical string either way. Nothing here is wrong, it
is round-one scaffolding for a distinction round two removed. Collapse the pairs, or
keep them and make the comments say what is true.

2. The upgrade guard is pinned on one path, and the other one now hands over a raw
eight characters. test_opaque_tray_still_matches_an_existing_six_char_filament covers
`_find_or_create_filament`. The public `find_or_create_filament` has its own separate
match loop, and spoolman.py:1208 now passes `tray.tray_color` where it passed
`tray.tray_color[:6]` — the non-BL RFID auto-create. The existing tests in
test_spoolman_inventory_methods.py only ever pass six-character values, and the new
test_find_or_create_filament_matches_on_prefix_and_creates_with_alpha stubs
`get_filaments` to `[]`, so despite the name it never exercises matching at all; it
pins the create value only.

So if the FF fold ever comes off that key, every non-Bambu RFID spool on the instance
mints a duplicate filament on the next sync and nothing goes red. One test closes it:
an existing filament at FF0000, call with FF0000FF, assert the id comes back and
create_filament was not called.

3. Optional, display only. Two swatches hand-roll the colour instead of going through
`getSwatchStyle`. SpoolmanFilamentPicker.tsx:89 and :146 do
`backgroundColor: '#' + hex`, so a clear filament — which is exactly what this PR puts
into that picker — renders as #00000000, a fully transparent swatch, i.e. blank.
LinkSpoolModal.tsx:129 does the same on filament_color_hex. utils/colors.ts:353 already
returns the checkerboard for an alpha of 00 and the RGB prefix otherwise. Reachable
before this PR only for filaments some other tool wrote; this PR makes it the normal
rendering for clear rolls. Fine as a follow-up if you'd rather keep this PR to the
storage layer, but say so rather than leaving it.

Nit: `_create_filament_from_external` (spoolman.py:1315) still uses
`external.get("color_hex", <tray fallback>)`. I agree the alpha-aware key is what closed
the reported path and I confirmed it does. The fallback expression itself is still dead
for any entry that carries a colour, and it returns None rather than the fallback for an
entry whose color_hex is explicitly null. Not reachable in a way that matters — a
null-colour entry is only a candidate when the tray colour is empty too — so a plain
`or` or a comment, not a code change.

Checked and holds: internal mode already stores the full eight characters
(spool.rgba is String(8)), so this brings Spoolman mode up to internal-mode parity
rather than past it. `_spool_color_to_hex` drops alpha for the archive rewrite
deliberately and documents it, so a clear roll still lands as #000000 in the colour
graph and that is correct. `_validate_rgba` admits 6 or 8, both metadata_unchanged
directions are pinned, and rewriting test_eight_char_color_hex_falls_back rather than
deleting it was right the first time and still is.

On merge order: take this one first. #2944 is the smaller rebase of the two, and the
comparison lines it touches will read more clearly once the key semantics here are
settled and the comments match them.

ojimpo:
Round three, on `d7e6ce3b`. All four addressed, including the optional one — taken here rather than as a follow-up, with the reason below.

**1. The round-one vocabulary.** Right, and it was the worse half of the problem: a comment that states the overturned invariant is an instruction to reintroduce the bug. All three are rewritten to say the key *is* the stored shape, and the mechanical corollary is collapsed rather than re-commented. `color`/`stored_color`, `color_hex`/`color_upper` and `cur_rgba` are one name each now. I checked the falsy/whitespace divergence you noted before removing it: no caller reaches those branches with such a value, and neither branch handled it deliberately, so there was nothing to preserve.

**2. The other match loop.** Confirmed, and I had not spotted it — `spoolman.py:1208` hands over `tray.tray_color` whole now, and the only test whose name claimed to cover matching stubs `get_filaments` to `[]`, so it pinned the create value and nothing else. Renamed to what it actually does, and added `test_find_or_create_filament_matches_an_existing_six_char_filament`: existing filament at `FF0000`, called with `FF0000FF`, asserts the id comes back and `create_filament` was not called. Verified it fails as expected if the fold comes off the key.

**3. The swatches — taken in this PR.** The argument for a follow-up would be scope, but the reachability is the wrong way round for that: before this PR a `#00000000` filament only existed if some other tool wrote it, and after it that is the normal storage for every clear roll. Shipping the storage change without the two call sites means the first thing a clear-spool user sees is a blank circle in the picker they select it from. `SpoolmanFilamentPicker.tsx` (both swatches, local `colorStyle` helper deleted) and `LinkSpoolModal.tsx` go through `getSwatchStyle`, and each has a test pinning the checkerboard and asserting no `backgroundColor` is set.

**Nit.** Changed to a plain `or`, with a comment saying it is a tidy and not a live path — an entry whose `color_hex` is explicitly null is only a candidate when the tray colour is empty too.

Verification on `d7e6ce3b`:

```
backend/tests/unit    7918 passed, 22 skipped
frontend              3256 passed, 240 files
ruff check            All checks passed!
ruff format --check   804 files already formatted
eslint .              clean
npm run build         clean (tsc -b && vite build)
```

Frontend is 3256 against your 3254 — the two new swatch tests.

On merge order: understood, and this one is ready. I will rebase #2944 onto `dev` once this lands and re-run the full suite there before flagging it.

Issue #2614, opened 2026-07-19
[Bug]: Selected plate from multi-plate 3MF records whole-file filament usage in statistics


Jostxxl:
### Bambuddy version

`dev` at `69c0effb` (native/systemd deployment with PostgreSQL)

### Describe the bug

When one selected plate from a sliced multi-plate 3MF is printed through the queue, the completed print log/statistics can store the filament usage of the **entire 3MF** instead of the selected plate.

Our reproducible production example is a sliced file named `heart 3.gcode.3mf` containing 22 plates. Each queue job targets one plate, and the archive rows correctly have different `plate_id` values. Nevertheless, completed print-log rows all received exactly the same whole-file value:

```text
archive 359, plate_id 1  -> 12006.49 g
archive 360, plate_id 2  -> 12006.49 g
archive 368, plate_id 3  -> 12006.49 g
archive 393, plate_id 10 -> 12006.49 g
archive 402, plate_id 11 -> 12006.49 g
archive 450, plate_id 17 -> 12006.49 g
archive 451, plate_id 18 -> 12006.49 g
```

There are 13 affected completed print-log entries in this installation, totaling `156084.37 g`. This makes the lifetime/user statistics incorrect by roughly 156 kg from this file alone.

As a control, plates uploaded today as physically separate, plate-specific files (for example `heart 3_plate_17.gcode.3mf`) show plausible per-plate values of roughly 230-610 g. Those uploads do **not** prove that selection from the original 22-plate 3MF is fixed, because those files only contain one plate.

### Steps to reproduce

1. Add a sliced multi-plate 3MF to the library (our example contains 22 plates and has a whole-file filament total of about 12.0 kg).
2. Create a queue item for one selected plate, preserving the original multi-plate 3MF.
3. Dispatch and complete that one plate.
4. Open Print History / Statistics or inspect `print_log_entries.filament_used_grams`.

### Expected behavior

The completed print entry and all statistics should use only the selected plate's filament usage. For our plates, this is approximately 230-610 g depending on the plate.

### Actual behavior

Every completed plate receives the complete 22-plate 3MF total (`12006.49 g`). That whole-file amount is then summed again for every run in the statistics.

### Additional context / likely code path

This is separate from #2603. That fix persists and displays the queue item's selected `plate_id`, but it does not repair these already stored whole-file filament values, and we have not found evidence that it scopes the archive metadata value used by the completed print-log/statistics path.

[...]

PR #2925, opened 2026-08-23
Repair the tare on spools that took the arbitrary catalogue row (issue #2909)

ojimpo:
## Description

Follow-up to #2923, as agreed on that review. That PR stops `create_spool_from_tray` taking whichever `Bambu Lab%` row the database returns first; this repairs the spools created before it.

On SQLite the row it took is `Bambu Lab - Plastic High Temp` — 216 g against the 250 g the spool actually weighs. Left alone, a spool the user never opens the form for stays wrong indefinitely, and it is not a display error: `core_weight` is the tare in SpoolBuddy's weigh flow, so every scale weighing of an affected spool has reported 34 g of filament that is not on it and written `weight_used` 34 g low. Remaining weight, low-stock alerts and the pre-print filament check all read that number afterwards.

## Why it keys on the value

I offered `core_weight_catalog_id IS NULL` on #2923 and it is the wrong guard, for the reason @maziggy identified there. `AdditionalSection.tsx` displays the first catalogue row matching the weight and auto-selects it when exactly one matches — and `DEFAULT_SPOOL_CATALOG` has exactly one 216 g row (I counted: one at 216, one at 250, out of 91). So opening a spool's form and saving any unrelated field writes the High Temp id, and the arbitrary pick reads as a deliberate user selection from then on.

| spool | `IS NULL` guard | value guard (this PR) |
|---|---|---|
| never opened in the form | repaired | repaired |
| opened and saved once — id laundered to High Temp | **skipped**, and it is the one most likely to have been weighed | repaired |
| hand-typed 207 g, matching no row, so id cleared to NULL | **overwritten** — the case the guard was for | untouched |

`data_origin = 'rfid_auto' AND core_weight = 216` selects the wrong spools and only the wrong spools. All three rows above are pinned by tests.

## Shape

- **The weight comes from the catalogue row, not a constant.** An install that has corrected `Bambu Lab - Plastic Low Temp` to its own measurement gets that value, and `core_weight_catalog_id` is set to the row that supplied it — same rule as the forward fix, so the two cannot drift.
- **With no such row, nothing is touched.** There is then nothing to say what the right value is, and 216 is at least a number the user can see and edit. Logged rather than silent.
- **One-shot via a settings flag**, following `_migrate_scope_run_filament_to_plate` (#2614). Not for idempotency — the `UPDATE` is naturally idempotent once it has run — but because a user may deliberately put an `rfid_auto` spool on a High Temp core afterwards, and without the flag the next restart would silently revert it. That is pinned by a test too.

## Known limits, stated rather than papered over

- **A Postgres install may carry a different wrong value.** With no `ORDER BY` the pick was unordered there, so a spool could have landed on White (253) instead. Those are not selected: unlike 216 on SQLite there is no basis to say whether 253 was the arbitrary pick or a deliberate choice, and guessing would overwrite real data. SQLite is the default and the reported case.
- **An install that edited the High Temp row before its spools were created** carries that edited weight instead of 216 and is not selected, for the same reason.

Both leave the user where they were rather than moving the damage somewhere harder to see, and the forward fix means neither can grow.

## Related Issue

#2909 — follow-up to #2923 (the forward fix), agreed on that PR's review. Filing separately as requested.

## Documentation

**Pick one**:
- [x] No docs update required — reason: one-shot data repair. No new setting, endpoint or UI element; the value it writes is the 250 g default the wiki already documents (`docs/spoolbuddy/usage.md`).

## Type of Change

## Changes Made

- `backend/app/core/database.py` — `_migrate_repair_rfid_core_weight`, called at the end of the migration run. Reads the Low Temp row, updates `data_origin='rfid_auto' AND core_weight = 216` spools to its weight and id, marks a settings flag so it never runs twice
- `backend/tests/unit/test_rfid_core_weight_backfill_2909.py` — seven tests

## Screenshots

## Testing

- [x] I have tested with my printer model: **X2D** — the three affected spools on my instance are its AMS RFID reads:

```
id, material, subtype, colour, core_weight, core_weight_catalog_id, origin
(1, 'PLA',  'Basic', 'Black',       216, None, 'rfid_auto')
(2, 'PLA',  'Matte', 'Ivory White', 216, None, 'rfid_auto')
(3, 'PETG', 'Basic', None,          216, None, 'rfid_auto')
```

Seven tests, one per case in the table above plus the catalogue-missing and one-shot paths. Two are verifications rather than assertions of the happy path:

- `test_repairs_a_spool_whose_form_laundered_the_wrong_row` seeds the High Temp id on the spool — it is the case an `IS NULL` guard would skip
- `test_does_not_revert_a_216_set_after_the_repair` runs the migration, sets 216 back by hand, runs it again. I confirmed the gate is load-bearing by removing the early return and watching this one fail

Full backend suite: **10634 passed, 22 skipped**. `ruff check` and `ruff format --check` clean. Run in the repo's own `backend-test` image.

## Checklist

### DIFF
--- a/backend/app/core/database.py
+++ b/backend/app/core/database.py
@@ -4427,6 +4427,91 @@ async def run_migrations(conn):
         conn, "ALTER TABLE notification_providers ADD COLUMN on_ams_drying_suspended BOOLEAN DEFAULT TRUE"
     )
 
+    # Migration: repair the core weight of RFID-created spools that took the
+    # arbitrary "first Bambu Lab row" from the spool catalogue (#2909).
+    await _migrate_repair_rfid_core_weight(conn)
+
+
+async def _migrate_repair_rfid_core_weight(conn) -> None:
+    """Give RFID-created spools the tare of the spool they actually arrived on (#2909).
+
+    ``create_spool_from_tray`` looked the core weight up with a ``Bambu Lab%``
+    prefix query and took the first row back, with no matching step and no
+    ``ORDER BY``. On SQLite that is ``Bambu Lab - Plastic High Temp`` — 216 g
+    against the 250 g the spool actually weighs. The forward fix selects the row
+    by name; spools created before it keep the wrong value, and it is not
+    cosmetic: ``core_weight`` is the tare in SpoolBuddy's weigh flow, so every
+    scale weighing of one of these spools charges 34 g of filament that is not
+    there.
+
+    Keyed on the value, not on ``core_weight_catalog_id IS NULL``. The catalogue
+    picker in the spool form displays the first row matching the weight and
+    auto-selects it when exactly one does, and ``DEFAULT_SPOOL_CATALOG`` has
+    exactly one 216 g row — so opening a spool's form and saving any unrelated
+    field writes the High Temp id, laundering the arbitrary pick into what reads
+    as a deliberate user selection. A NULL guard would therefore skip precisely
+    the spools that have been through the form, which are the ones most likely to
+    have been weighed. It would also overwrite a user who typed a corrected weight
+    matching no catalogue row, because that same effect clears their id to NULL.
+
+    The weight comes from the catalogue row rather than a constant, so an install
+    that has corrected ``Bambu Lab - Plastic Low Temp`` to its own measurement
+    gets that value; ``core_weight_catalog_id`` is set to the row that supplied
+    it. With no such row the spools are left alone — there is nothing to say what
+    the right value is, and 216 is at least a number the user can see and edit.
+
+    One-shot via a settings flag. Without it a user who deliberately sets an
+    ``rfid_auto`` spool to 216 g would have it silently reverted on the next
+    restart.
+    """
+    from sqlalchemy import text
+
+    flag = "_backfill_2909_rfid_core_weight_done"
+
+    async with conn.begin_nested():
+        already = (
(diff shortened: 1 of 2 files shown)


ojimpo:
Superseded by b38022ec on `dev` — closing.

Your version is the better one and it is worth naming why, so the next person
reading this PR does not think it was dropped for being redundant.

It repairs `weight_used`, which mine did not touch at all. I had treated the tare
as the whole of the damage and stopped there, but a spool that has been on the
scale carries the 34 g through into its used weight, and because that error is a
constant it can be added back exactly however much has been printed since. Mine
would have left every weighed spool still wrong in the field that actually drives
remaining weight and the low-stock alerts.

And it selects on the weight of *any* of the other Bambu catalogue rows rather
than the literal 216. I had scoped the Postgres case out and said so in the
description, on the grounds that 253 could not be told apart from a deliberate
choice — but that argument applies equally to 216, which I had already accepted,
so it was not a consistent line. Reading both weights out of the catalogue rather
than hardcoding them is the part I should have taken from my own forward fix and
did not.

The one case that cannot be distinguished is stated in your version instead of
being guarded against badly, which is the right way round.

#2923 is still the forward half — `spool_tag_matcher.py` on `dev` is still asking
for the first `Bambu Lab%` row, so new spools are still created with whatever the
database returns first. It has gone CONFLICTING since b38022ec touched that file;
I will rebase it onto current `dev` and push.

Issue #2762, opened 2026-08-04
[Feature]: Dispatch uploads via the native port-6000 - H2* and P2S are using internal eMMC


bmfurtado:
### Problem

On H2-series printers (I tested an H2D on current firmware, LAN-only + Developer Mode), **FTPS on 990 only exposes external storage**. With no USB drive inserted, every directory listing comes back empty and every `STOR` fails with `553 Could not create file` — I verified this with a manual FTPS client (TLS session reuse on the data channel, `bblp` + access code, trying `/`, `/cache`, and other paths). Bambuddy's dispatcher (`bambu_ftp.py`) is FTPS-only, so queue dispatch to a stick-less H2D always fails:

```
FTP upload failed for /<file>.3mf: 553 Could not create file. (error code: 553)
```

Meanwhile, Bambu Studio sends to the same printer without any external storage. I watched a successful direct Studio send on the wire; the sequence was: bind/detect on port 3000 (×2), MQTT on 8883, then the actual transfer as **two TLS connections to port 6000** — the "BambuTunnelLocal" protocol — which writes to *internal* (eMMC) storage. FTPS was only touched briefly and carried no file.

### Request

Implement tunnel-based upload as a dispatch path: prefer the native `:6000` tunnel for printers that support it, and fall back to FTPS otherwise (mirroring Studio's own selection logic). This would make dispatch work on H2-series printers without requiring users to dedicate a USB stick and enable "Store Sent Files on External Storage".

### References

- [ClusterM/open-bambu-networking](https://github.com/ClusterM/open-bambu-networking) — an open-source (AGPL-3.0) reimplementation of Studio's network plugin with a working native `:6000` wire: chunked upload, eMMC browsing, `SUB_FILE`/thumbnail commands, and the documented FTPS fallback ("legacy printers without `:6000` use Studio SendJob → FTPS").
- Side benefit: the same tunnel serves internal-storage listing and 3MF thumbnails, which could also improve the file-manager features for H2/X2-generation printers.

### Current workaround

A USB drive in the printer plus "Store Sent Files on External Storage" — this works, but surprises users whose printers work fine in Studio without a stick.

maziggy:
New models H2* and P2S are now using internal eMMC storage per default when sending prints via slicer. This is a new behaviour.

maziggy:
Scope note after a day of measuring this on H2C, H2D and X1C hardware.

What this issue does NOT cover, so nobody picks it up expecting it:

[...]

Issue #1820, opened 2026-06-24
[Bug]: Optional AMS remain%-delta fallback for Spoolman weight on no-3MF ("Untitled") prints


ojimpo:
### Component

Bambuddy (Spoolman weight tracking)

### Summary

When a print starts that Bambuddy cannot fetch a 3MF for (the job shows up as "Untitled" / "名称未設定"), **no filament consumption is reported to Spoolman at all** for that print. The spool's remaining weight simply never decreases for that job.

This is an edge case, but it still comes up in **shared / multi-user setups**: every so often someone slices and prints an unsaved project, the printer reports `subtask_name: 名称未設定` and keeps no retrievable `.3mf` on its FTP, so the estimate-based path has nothing to read.

### Where it happens (current behavior, dev)

`store_print_data()` is the only weight writer for Spoolman since #1119. With no 3MF it early-returns and no `ActivePrintSpoolman` row is created, so completion reports nothing:

```
FTP download failed for /名称未設定.gcode.3mf: 550 Failed to open file. (not on printer)
... (all candidate paths 550) ...
Could not find 3MF file for print: /data/Metadata/plate_1.gcode
Created fallback archive 131 for 名称未設定 (no 3MF available)
[SPOOLMAN] No filament usage data in 3MF for archive 131
# at completion:
[SPOOLMAN] No tracking data for print (printer=1, archive=131)
```

(H2S, Bambuddy 0.2.5b1. Named jobs on the same printer track fine — they get `[SPOOLMAN] Stored tracking data`.)

### Relationship to past issues (so this doesn't re-open a closed problem)

- **#1119** intentionally made per-print 3MF tracking the *only* weight writer and dropped the AMS remain%-based writer, because AMS doesn't report `tray_weight` without RFID, so **non-BL spools** got silently mis-/un-tracked. This request is **not** "bring the old global AMS sync back."
- It is deliberately scoped to avoid the #1119 failure mode: the fallback should fire **only when (a) there is no 3MF for the print AND (b) the tray is an RFID/Bambu spool that actually reports `remain%` + weight.** Non-RFID trays (the #1119 case) are skipped, so they keep getting no bogus writes.
- Distinct from #1771 / #1793 (runout mis-attribution = over-count). This is the opposite: a whole print silently *not counted*.

### Proposed behavior

[...]

PR #2931, opened 2026-08-23
Fall back to the Bambu Cloud task's cover when the 3MF is not reachable over FTPS

### LINKED ISSUE #2910: [Feature]: Fall back to the Bambu Cloud task cover when the print's 3MF is not reachable over FTPS
### Problem or Use Case

On H2-series printers the active print's 3MF often is not on external storage, and
`GET /api/v1/printers/{id}/cover` cannot find it — so print cards and archives show no image.

Two independent reasons, both measured on an H2C (firmware `01.02.00.00`, LAN + Developer Mode):

**1. The file is on internal eMMC, and FTPS cannot reach it.** FTPS on :990 is chrooted to external
storage. `/data`, `/cache`, `/model`, `/Metadata` and `/sdcard` all answer `550`, and `CWD ..` lands
back at `/`. Only the card root and `/timelapse` are real. This is the same storage split reported in
#2762.

**2. `subtask_name` is not always a filename.** `_produce_cover_image` builds candidate 3MF names
from it — `<name>.gcode.3mf`, `<name>.3mf`, plus space→underscore variants — and tries each under
`/`, `/cache/`, `/model/`, `/data/`. For a print started from a MakerWorld profile, `subtask_name` is
the **profile title**. A job running here reports:

```
subtask_name = "PETG 0.2mm layer, 2 walls, 15% infill"
gcode_file   = "/data/Metadata/plate_1.gcode"
```

There is no filename in that string, so no number of path or naming variants can ever match it. The
lookup exhausts every path, `_cover_404_cache` is written, and the endpoint answers:

```
404 {"detail":"No cover available for 'PETG 0.2mm layer, 2 walls, 15% infill' (cached)"}
```

`PrintArchive.thumbnail_path` stays NULL for the same print, so print history has no image either.

On this deployment: **158 archives, 78 with no thumbnail.**

### Proposed Solution

Fall back to the Bambu Cloud task's cover, which is the sliced plate render itself and needs no
filename at all. Everything required is already in the codebase:

- `PrintArchive.subtask_id` is already parsed off MQTT and stored (`main.py`, the
  `raw_mqtt.get("subtask_id")` path, including the `"" / "0" → None` normalisation).
- `BambuCloudService` already holds the session and already handles token expiry
  (`is_expiry_401`, `validate_token`).

`GET https://api.bambulab.com/v1/user-service/my/task/{subtask_id}` returns, for the print above:

```
cover        https://makerworld.bblmw.com/makerworld/cache/<redacted>/3mf/1/<colour>/REP1/Metadata/plate_1.png
designTitle  "H2D / H2S / H2C Poop Chute"
designId     1265042      profileId 296523048      plateIndex 1
```

Measured:

| call | result |
|---|---|
| task lookup, no `Authorization` | **401** |
| task lookup, with the stored cloud token | **200** |
| the `cover` URL itself, no auth at all | **200**, 3654 bytes, `image/png`, 512×512 |

That `Metadata/plate_1.png` is exactly the asset the FTP path was trying to extract from the 3MF.

Shape of the change:

1. Add a `get_task(subtask_id)` to `bambu_cloud.py`.
2. In `_produce_cover_image`, after the FTP paths are exhausted and **before** `_cover_404_cache` is
   written, resolve the archive's `subtask_id` and try the task cover.
3. Optionally use the same call at print start to fill `PrintArchive.thumbnail_path`, so print
   history and the archive card benefit and not just the live printer card.

On this deployment that fills **71 of the 78** thumbnail-less archives.

Four things I would want to get right in the implementation, each of which bit me while measuring:

- `subtask_id` arrives as a **string**, so `!= 0` and truthiness both pass for `"0"`. The existing
  MQTT path already normalises this; the new one must use the same rule.
- `api.bambulab.com` rejects a default `urllib` User-Agent with error `1010`. Going through the
  service's own client avoids it.
- `view=top` and `view=pick` must keep returning 404 on this path rather than silently serving the
  default cover — the cloud task exposes one cover, not the three views.
- Coverage is conditional on the print having been dispatched through the cloud, not merely on the
  server being signed in. 19 of 158 archives here have no `subtask_id` at all and stay uncovered;
  those should keep the existing 404 rather than a misleading placeholder.

### Alternatives Considered

- **Port-6000 tunnel (#2762).** I tried it against this H2C. The tunnel does list the running print's
  3MF by its real name and size on internal storage, but `FILE_DOWNLOAD` is refused on firmware
  `01.02.00.00` and `SUB_FILE` is not recognised, so no image comes out of it here. That issue's
  "side benefit: the same tunnel serves 3MF thumbnails" may not hold on this firmware, which seems
  worth recording either way.
- **Anonymous `design-service/profile/{profileId}`.** Returns 200 with a `coverUrl` and **no**
  sign-in, so it would work for installs not connected to Bambu Cloud. But it is MakerWorld's
  marketing photo rather than the plate render, and it is wrong on multi-plate models. Reasonable as
  a second fallback, not as the primary.
- **A camera still.** `/camera/snapshot` works, but it shows the bed as it is now, not the model.
- **Reaching internal storage over FTPS.** Chrooted, as above.

### Feature Category

Print Archiving

### Priority

Would improve my workflow

### Mockups or Examples

The endpoint today, for the print described above:

```
GET /api/v1/printers/2/cover?token=...
404 {"detail":"No cover available for 'PETG 0.2mm layer, 2 walls, 15% infill' (cached)"}
```

The cover the cloud task hands back for that same print is a 512×512 transparent PNG of the sliced
plate — the correct model, in the print's own filament colour.

### Contribution

### Checklist

Closest prior art I found, and why this is not a duplicate: #2762 is about **upload dispatch** over
the port-6000 tunnel and mentions thumbnails only as a possible side benefit; #1420 is the FTP cover
loop whose negative cache this change would write into; #1820 covers no-3MF "Untitled" prints from
the Spoolman side. None of them proposes a cover source that does not depend on a filename.

sadontsev:
## Description

When the active print's 3MF cannot be read over FTPS, fall back to the Bambu Cloud task's cover so
the printer card still shows the model.

`get_printer_cover` finds the 3MF by **name**: it builds candidate filenames from `subtask_name` and
tries each under `/`, `/cache/`, `/model/` and `/data/`. Two things defeat that.

**The file is on internal storage.** On H2-series printers Bambu Studio uploads over port 6000 to
eMMC, and FTPS on :990 only exposes external storage — `/data`, `/cache`, `/model` and `/Metadata`
all answer 550, and `CWD ..` returns to `/`. Same storage split as #2762.

**`subtask_name` is not always a file name.** A print started from a MakerWorld profile reports the
*profile title*. Measured on an H2C:

```
subtask_name = "PETG 0.2mm layer, 2 walls, 15% infill"
gcode_file   = "/data/Metadata/plate_1.gcode"
```

There is no file name in that string, so no combination of paths or naming variants can ever match
it — this case fails however the file was stored.

Either way the search exhausts every path, `_cover_404_cache` is written, and the endpoint 404s.
`PrintArchive.thumbnail_path` stays NULL for the same print, so print history has no image either.
On my install that was **78 of 158 archives**.

`subtask_id` identifies the *job* rather than the file. It is already parsed off MQTT and stored, and
`BambuCloudService` already holds the session and its 401-expiry handling — so the fallback is one
`get_task` call plus a download. The task carries `cover`, which is the sliced plate render: the same
`Metadata/plate_N.png` that is inside the unreachable 3MF. It filled **71 of those 78**.

Measured: the task lookup answers 401 without the session and 200 with it; the CDN URL it returns
needs no credential at all.

**Relationship to `_running_print_archive_file`.** While rebasing I picked up your change that
looks for the archived copy of the running print's 3MF, anchored on `subtask_id`. That is the same
insight from the other side, and a local file beats a network round trip — so this runs strictly
*after* it, and only a print with no local copy and no readable file on the card reaches the cloud
at all. Rebased onto that; suite is green at 10708.

## Related Issue

Fixes #2910

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#76

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

## Changes Made

- `services/bambu_cloud.py` — `BambuCloudService.get_task()`, one GET against
  `/v1/user-service/my/task/{id}`, same shape as the other methods on the class.
- `api/routes/printers.py` — `_cloud_task_cover()`, tried only when `_produce_cover_image` raises
  404, and only for the default view.
- `backend/tests/unit/test_cloud_task_cover_2910.py` — 8 cases, all on the failure paths.

## Testing

- [x] I have tested with my printer model: H2C (firmware 01.02.00.00, LAN + Developer Mode)

`./test_backend.sh` — **10712 passed**, rebased onto current `dev`. `ruff check` and
`ruff format --check` clean on all three files.

12 cases. Eight cover `_cloud_task_cover`'s failure paths — which is the whole design, since every one of them has to return `None` so the existing 404 stands. The other four cover `get_task` directly: the first eight mock the service out, so without them the new method would have shipped with its error branches never executed.

`bandit` reports the same 5 findings as `dev` — this branch adds none. It briefly added a sixth, `B105` on the `cloud_token != ""` filter; that is a SQL emptiness test rather than a credential, and the filter earns its place (without it a row holding an empty token could win the `limit(1)` and abort the lookup while another user has a usable session), so it carries a `# nosec B105` with that reason, matching the convention already used in `core/auth.py`.

Verified against the live printer: `/printers/2/cover` returned 404 with
`No cover available for 'PETG 0.2mm layer, 2 walls, 15% infill' (cached)`, and returns a 3070-byte
PNG after this change, logged as `Cover: served from the cloud task <id>`.

## Checklist

## Additional Notes

Deliberate limits, so review does not have to guess whether they were considered:

- **Only the default view falls back.** The cloud task exposes one cover; serving the angled render
  for `view=pick` would hand the skip-objects modal a picture where it expects a coordinate mask. So
  `view=top` and `view=pick` keep their 404.
- **Every failure returns `None`** and lets the existing 404 stand — no session, an expired one, an
  unreachable Bambu, a task with no cover, a body that is not a PNG. A cover is decoration and a
  broken one is worse than none. That is what the tests pin.
- **A print that never went through the cloud has no `subtask_id`** and is unreachable this way — 19
  of 158 archives here. They keep the old 404, which is the honest answer.
- The negative cache entry is cleared on success, or the next request would short-circuit to 404
  before it could reach the fallback at all.

**One thing I would like your opinion on.** On an auth-enabled deployment the cloud session is keyed
to a *user*, and this route is authenticated by a camera stream token, so it has no user context. I
use the lowest user id holding a session — ordered rather than "whatever the database returns first",
so a deployment behaves the same way twice. It cannot leak between users, because the task endpoint
is scoped to the token's own account and a session that does not own the running job's task simply
gets nothing back. If you would rather it only used a global (auth-disabled) token and did nothing
when auth is on, that is a two-line change and I will make it.


maziggy (changes requested):
Rebased and re-checked against current dev. The code is still good.

Two things landed today in the same path: #2957 puts try_recover_fallback_archive
inside _produce_cover_image, and #1820 makes the storage verdict actually fire for
prints started from the printer's screen. Neither collides with this. You catch 404
specifically, both of #1820's paths still end there, so the fallback now runs in
milliseconds instead of after the sweep.

I did check whether caching the cloud PNG would starve #2957's recovery. It doesn't:
a cool-off raises 503, which you decline, and that is the only case #2957 retries.
Its recovery runs off the 3MF cache and the retry ladder in main.py, not this route.

ruff clean, full suite 11054 passed on the merge.

One note on the numbers: #1820 now probes five directories for screen-started prints,
and an H2S keeps recent jobs under /cache, so a few of your 78 will find a real local
3MF and the cloud's share will be smaller than you measured. Still worth having.

What I do want before merge is the other half of the issue. #2910 leads with "print
cards and archives show no image" and 78 of 158 archives, and step 3 is filling
thumbnail_path. This PR fixes the card only, so history stays blank for exactly the
prints it is about, and #2957 schedules no retry for a storage verdict, so nothing
else will ever fill them. Please cover it here rather than leaving it open.

It should be small: archive_dir() already handles file_path == "" by returning
archive_dir/<id>, so the PNG has somewhere to go, and you can anchor the row on
subtask_id the way _running_print_archive_file does. One edge to get right --
archive_print assigns existing.thumbnail_path unconditionally on recovery, so a 3MF
that turns up without a thumbnail would blank a good cloud one.

On your user question: keep what you have. Ordering by id beats "whatever the database
returns first", and the task endpoint being account-scoped is the real guarantee.
Global-token-only would just make this do nothing on every auth-enabled install.

sadontsev:
Covered here, as asked — the row now gets the cover, not just the card.

`_fill_cloud_cover_thumbnail` runs at the one place a no-3MF archive is created
(`main.py`, right after the fallback row commits), writes the PNG under
`archive_dir(archive)` — which, as you said, already resolves to `archive_dir/<id>`
for `file_path == ""` — and records `thumbnail_path`. Best-effort throughout: print
start must not fail because a decoration could not be fetched, so a raising fetch,
a missing cover and a missing session all leave the row exactly as it was.

I split `cloud_cover_bytes(subtask_id)` out of `_cloud_task_cover` rather than
copying it. The route half resolves `subtask_id` from live printer state; the
archive half has it on the row already. That keeps one copy of the token lookup,
the `get_task` call and the CDN download — and one copy of the `"0"` / `""`
normalisation, which is the part I did not want duplicated.

On the edge you named: `archive_print` assigned `existing.thumbnail_path`
unconditionally, so a 3MF turning up later without a thumbnail erased a good cloud
cover. It is `thumbnail_path or existing.thumbnail_path` now. That is the case I
most wanted pinned, since these rows are the least likely to ever get another
image — the test fails without the guard with
`assert None == 'archive/7/thumbnail.png'`.

Eight new tests: the write-and-record path, the `""` / `"0"` / `"  "` / `None`
declines (with `cloud_cover_bytes` asserted un-awaited, so the normalisation is
tested rather than assumed), no-cover and raising-fetch leaving the row alone, and
the recovery guard above.

Also took your correction on the numbers — with #1820 probing five directories for
screen-started prints, some of the 78 will now find a real local 3MF and the
cloud's share will be smaller than I measured. I have not re-measured; the row-fill
matters regardless of how the split lands.

Rebased on current `dev`. ruff clean, 8642 unit passing.

One unrelated thing worth flagging: `test_download_deadline_and_gate_2957.py::
test_a_late_transport_error_is_not_logged_as_loop_noise` fails for me on a clean
`upstream/dev` worktree as well as on this branch, so it is not from this work. It
asserts around a 0.05s cap, so it may just be this machine being slow — but if CI
is green on it, that is a difference worth knowing about.

Issue #1112, opened 2026-04-24
[Bug]: Where are External Folder files stored?


Carter3DP:
### Component

Bambuddy

### Bug Description

Where are files stored when using an external folder in file manager? For example say you have an external folder that links to a NAS, and you upload a file through the file manager in Bambuddy. You won't be able to see that file if you access that NAS through another computer.

### Expected Behavior

Files uploaded through the file manager to an external folder should show up in that external folder if you bring it up on another computer.

### Steps to Reproduce

1. Create an external folder that links to a SMB server.
2. Through Bambuddy, upload a .gcode.3mf file to the external folder.
3. Access the external folder outside of Bambuddy, the file will not be there.

### Printer Model

None

### Bambuddy Version

0.2.3.2

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

### Relevant Logs / Support Package

[bambuddy-support-20260424-094242.zip](https://github.com/user-attachments/files/27054232/bambuddy-support-20260424-094242.zip)

### Screenshots

<img width="1920" height="1004" alt="Image" src="https://github.com/user-attachments/assets/f787328a-b17c-47ba-9323-596e790b0f5d" />
<img width="1920" height="1031" alt="Image" src="https://github.com/user-attachments/assets/68783b65-7a79-4924-abb5-a3f0a54a9670" />

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I am using the latest version of Bambuddy
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled

maziggy:
Yup, found a small issue.

Available/Fixed in branch dev and available with the next release or daily build. Please let me know if it works for you now.

-----
If you find Bambuddy useful, please consider giving it a ⭐ on [GitHub](https://github.com/maziggy/bambuddy) — it helps others discover the project!

Carter3DP:
It does not appear to working still, now files no longer appear if uploaded to an external folder. You can still upload a file to a folder that exists on Bambuddy and then move it to the external folder, and it will appear on Bambuddy but not physically on the external folder.

This is on version 0.2.4b1

[bambuddy-support-20260427-095013.zip](https://github.com/user-attachments/files/27128129/bambuddy-support-20260427-095013.zip)

maziggy:
[...]

PR #2932, opened 2026-08-23
Serve a library file's geometry as STL (GET /library/files/{id}/mesh)

### LINKED ISSUE #2878: [Feature]: Serve a library file's geometry as STL (GET /library/files/{id}/mesh)
### Problem or Use Case

Bambuddy already parses model containers server-side — `stl_thumbnail` and `plate_thumbnail` both load a mesh with trimesh and render it. What it cannot do is hand the geometry itself to a client, so every client that wants an *interactive* 3D view has to parse the container on its own.

That is not hypothetical. `frontend/src/components/ModelViewer.tsx` (~1050 lines) ships JSZip, walks `Metadata/model_settings.config`, the plate JSONs and the `.model` XML by hand, and runs `STLLoader` — all to reach a mesh the backend can already read — and it does that work in the browser. That is the shape of #1412 ("3D preview in File Manager freezes the website"): a large container parsed client-side.

Third-party clients have it worse: a native app has to reimplement the same ZIP + 3MF handling from scratch to show the same model.

### Proposed Solution

`GET /api/v1/library/files/{file_id}/mesh` → binary STL (`model/stl`).

- Same auth and visibility as the adjacent `download_file` route (`LIBRARY_READ_ALL` / `LIBRARY_READ_OWN` + `_ensure_library_file_visible`).
- Loads with trimesh (already a dependency), simplifies toward a vertex budget, returns binary STL.
- Refuses a sliced `.gcode.3mf` with `400` — those carry toolpaths, so the G-code viewer is the right tool, and trimesh's 3MF loader raises `KeyError: 'world'` on Bambu Studio's sliced output anyway.
- `422` for a file that parses to no geometry, with a sentence fit to show a user.
- Parsing and decimation run in `asyncio.to_thread`, matching how `system.py` and `support.py` offload blocking work. Not theoretical: one real 3.1 MB model took 10.2 s to decimate.

What it buys the project beyond third-party clients: the web frontend could drop JSZip and the hand-written 3MF walk and simply `STLLoader` one response, already decimated server-side — which addresses #1412 at the cause rather than around it.

### Alternatives Considered

- **Server-rendered thumbnails (#156, already shipped).** Answers "which model is this?" but not "let me turn it around" — a PNG is not interactive.
- **Keep parsing in every client.** The status quo. It works, but it is the same ZIP + XML parser written once per client, and in the browser it competes with the UI thread.
- **A JSON vertex payload.** Roughly double the bytes of binary STL, and every 3D toolkit already reads STL.

The honest trade-off is size: STL is a wasteful transport (50 bytes per triangle, every vertex tripled), so the export can be *larger* than the container. Measured on a real 3.9 MB project 3MF — 100k vertices → 10.0 MB, 50k → 5.0 MB. There is no `GZipMiddleware`, so the client fetches the uncompressed size; I settled on a 50k budget for that reason, but it is a number worth arguing about.

### Feature Category

API

### Priority

Would improve my workflow

### Mockups or Examples

Measured against a real library, with the budget above:

| file | source size | STL | faces |
|---|---|---|---|
| plain `.3mf` | 770 KB | 80 KB | 1,596 |
| plain `.3mf` | 3.1 MB | 7.3 MB | 146,914 |
| plain `.3mf` | 3.9 MB | 5.0 MB | 99,988 |
| sliced `.gcode.3mf` | 6.1 MB | *refused* | — |

The middle one lands at 147k faces rather than the ~100k the budget asks for — quadric decimation cannot always reach its target, so a simplification that misses, throws, or empties the mesh falls back to delivering what it has rather than refusing.

### Contribution

The implementation is already written and tested: service + route + 15 unit tests, `./test_backend.sh` green (10,084 passed), verified against real library files in a running container.

I opened it as #2866 before reading CONTRIBUTING.md properly, and it was rightly closed — no assigned issue, and it targeted `main` instead of `dev`. That is on me. Raising it here first as the process asks; if you assign it I will re-open from a branch off `dev`, and if you would rather not add this surface at all, say so and I will drop it with no hard feelings.

### Checklist

Searched before filing: #156 (STL thumbnails — completed), #1412 (3D preview freeze), #2816, #2787, #2725. None of them asks the server to serve geometry.

sadontsev:
## Description

Adds `GET /library/files/{id}/mesh`, returning a library file's geometry as a binary STL.

A `.3mf` is a ZIP holding an XML mesh, and Bambuddy already opens it to render thumbnails. Serving
the mesh saves every client from shipping its own ZIP + 3MF handling to reach geometry the server
can already load — which is what a client otherwise has to do just to show a model in 3D.

This is the endpoint from #2866. That PR was rightly closed: I opened it with no assigned issue and
targeted `main`. Raised properly as #2878, which you have since assigned, so this targets `dev`.

## Related Issue

Fixes #2878

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#72

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

## Changes Made

- `services/mesh_export.py` — `is_exportable()` and `export_mesh_stl()`. trimesh only, already a
  dependency; no new packages.
- `api/routes/library.py` — the route, behind the same `require_ownership_permission(LIBRARY_READ_ALL,
  LIBRARY_READ_OWN)` pair the neighbouring file routes use.
- `tests/unit/services/test_mesh_export.py` — 17 cases.

## Testing

- [x] I have tested with my printer model: H2C (this endpoint is printer-independent — it reads
      library files, not the printer)

`./test_backend.sh` — **10729 passed**, rebased onto current `dev`.

`pytest --cov=backend.app.services.mesh_export` — **100%**, 17 cases, plus 12 route cases in
`test_library_mesh_route.py` and `TestLibraryOwnershipPermissions` covering the gate, the
statuses, both headers and the conditional request. The last two branches to get cover were the ones only reachable by making trimesh itself misbehave: an export that raises (which decides whether the route answers 422 or 500) and the `str`→`bytes` normalisation, which would otherwise fail in the route when it sets a byte length, well away from its cause.

`bandit` on the changed files reports the same 5 findings as `dev` — this branch adds none. `ruff check` and `ruff format --check` clean.

## Checklist

## Additional Notes

Decisions a reviewer would otherwise have to infer:

- **Sliced `.gcode.3mf` is refused with 400**, on the name, before touching disk. Those carry
  toolpaths — the G-code viewer is the right tool — and trimesh cannot read their scene graph
  anyway. Refusing early means a caller gets a specific status instead of a generic failure.
- **422, not 500, for a file with no geometry.** The request was well formed and the server is
  healthy; this particular file has nothing to give, and the message says which.
- **`asyncio.to_thread`**, because parsing and decimating is CPU-bound and can take seconds on a
  dense model. That is how `system.py` and `support.py` already offload blocking work, so this
  follows suit rather than adding a second way.
- **`MAX_VERTICES = 50000`, half of `stl_thumbnail`'s.** They are not the same question: a thumbnail
  is rendered once server-side, this is shipped over the wire to a phone.
- **`Cache-Control: private, no-cache` plus an `ETag`.** "Immutable" holds for managed files but
  not external ones: `scan_external_folder` keeps the row for a path it already tracks, so a model
  replaced over the mount has the same id and different geometry. The validator is built from
  `fs_modified_at` and `file_size`, and the conditional check runs before the parse so a
  revalidation costs a row read rather than seconds of CPU. `private` because the bytes sit
  behind a per-user permission check.
- **`Content-Disposition` via `safe_download_filename` then `build_content_disposition`.** Headers
  are latin-1, so an accented or Cyrillic filename would raise `UnicodeEncodeError` inside
  Starlette; and an external name is never validated on the way in, so a newline in a POSIX
  filename would survive the helper's quote stripping. Same order archives.py and printers.py
  use. Both cases are covered at the route.


maziggy (changes requested):
Good piece of work — the service is clean, the STL-over-JSON reasoning holds up, and the
"Additional Notes" section is the right way to hand a reviewer the decisions. I checked
the two things the vertex budget silently depends on and both are fine: fast-simplification
is pinned in requirements.txt, and trimesh 4.12's first positional parameter is `percent`,
so the decimation call means what the variable name says rather than a face count.

Verified on 33b37edb: ruff and format clean, backend 10878 passed, your 19 cases pass.

Two things before merge.

1. `Cache-Control: private, max-age=86400` with no validator is wrong for external files.

The justification — a pure function of an immutable stored file — holds for managed files:
they land on disk as `<uuid><ext>` and there is no replace endpoint. It does not hold for
external ones. `scan_external_folder` keys rows on the on-disk path, and for a path it
already tracks it refreshes `fs_modified_at` and `continue`s (library.py:1868-1881). Replace
the model over the mount — which is what #1112 exists for — and the same `file_id` now has
different geometry while every client that already fetched it keeps the old mesh for up to a
day, with nothing to revalidate against.

`LibraryFile` already carries `fs_modified_at` and `file_size`, so an ETag built from those
plus an `If-None-Match` check keeps the bandwidth win and closes the staleness. Dropping to
`no-cache` would also work, at the cost of the thing the header was for.

2. The route has no tests.

65 of the 483 lines are a new public endpoint with an ownership check, a 400, a 404, a 422
mapping and two response headers. Nothing under backend/tests/ references `/mesh` — the
service is at 100% and the route is at 0%. Every decision in your Additional Notes is a
route decision and none of them are pinned: swap the 422 for a 500, drop the Cache-Control
header, or hand `LIBRARY_READ_ALL` where `require_ownership_permission` belongs, and the
suite stays green.

The ownership one is the one I care about. A library route that returns file bytes to the
wrong user is a failure mode this project has shipped advisories for, and it is one test:
user B requests user A's file, expect 403.

3. TestTheDownloadHeader tests a helper dev already covers, and its docstring says otherwise.

Both cases call `build_content_disposition` directly, and the comment "raises if the route
ever goes back to an f-string" is not true — nothing in that test reaches library.py. The
existing backend/tests/unit/test_http_utils.py already has test_header_is_latin1_encodable,
test_quotes_and_backslashes_stripped_from_ascii_fallback and test_disposition_param_is_respected,
so the class is duplicating dev and pinning nothing new.

The trap you found is real and catching it in self-review was the right instinct — aim the
test at the code that would regress. One route test with a non-ASCII LibraryFile.filename,
asserting the response header encodes as latin-1, does what this class claims to.

Minor, and I would take either as a comment rather than a change:

- The filename is external-origin and skips `safe_download_filename`. Uploaded names pass
  through `validate_print_filename`, which rejects control characters and bounds length.
  Names discovered by `scan_external_folder` do not, and a POSIX filename may contain a
  newline; `build_content_disposition` strips only quotes and backslashes, so the ASCII
  fallback keeps it. archives.py:2495 and printers.py:2075 run externally-sourced names
  through `safe_download_filename` first for exactly this. `download_file` has the same
  exposure through Starlette so it is not new, but this is the route hand-building a header.

- The output is bounded and measured; the input is not. Library uploads have no size cap,
  `trimesh.load` holds the whole model in RAM, and `Response(content=blob)` buffers the
  export too — in a shared default executor thread that every other `to_thread` caller
  queues behind. `stl_thumbnail` has the same posture but runs once at upload time; this is
  a GET any reader can drive repeatedly and in parallel. A size pre-check or a small
  semaphore would bound it.

Notes, no action:

- `.ply` is in SUPPORTED_SUFFIXES but not `_SCANNABLE_EXTENSIONS`, so it can only ever
  arrive by upload, never from an external folder. Just asymmetric, not wrong.
- The service's own `.gcode.3mf` guard cannot fire for a managed file: a sliced upload lands
  on disk as `<uuid>.3mf`, so `export_mesh_stl` sees an exportable name and reaches trimesh,
  which then fails into "could not be read as a 3D model". The route's check on
  `file.filename` is what actually enforces it — correct as written, but the service-level
  guard is not the backstop the module docstring implies.
- `stem[: -len(".gcode")]` is dead for `.gcode.3mf` but earns its keep for `foo.gcode.stl`.

sadontsev:
Thank you — all three land, and the first one is a real bug I would not have found by reading my own diff. Pushed as `f5c33436`.

**1. `Cache-Control` without a validator.**

You are right and I had the wrong mental model of "immutable". I checked `scan_external_folder` at library.py:1868-1881 and it is exactly as you describe: a path it already tracks keeps its row, refreshes `fs_modified_at` and `continue`s, so the same `file_id` genuinely can serve different geometry.

Now `private, no-cache` plus a strong `ETag` built from `fs_modified_at` and `file_size` — the two fields that move when the bytes do, where the id does not. `no-cache` stores but revalidates, so the bandwidth saving survives as a 304.

The conditional check runs *before* the parse rather than after, which turns a revalidation into a row read instead of seconds of `asyncio.to_thread`. That felt like the more useful half of the fix.

**2. The route has no tests.**

Fair, and it was the wrong way round: the service at 100% and the route at 0% is coverage aimed where it was easy rather than where it mattered. New `backend/tests/integration/test_library_mesh_route.py`, 10 cases — 200 with `model/stl`, the 400, both 404s, the 422, both headers, and the two conditional-request paths.

On the ownership one: I wrote it as **404, not 403**, because that is what this codebase already promises. `_ensure_library_file_visible` returns 404 for a row the caller may not see, and its docstring says why — "not 403, to avoid id-enumeration leaks". The delete/update siblings 403 because they gate on the write permission before fetching the row; a read gates on visibility. Both cases are now in `TestLibraryOwnershipPermissions` beside the existing library ones: the owner gets 200 for a file that really exists on disk, another operator gets 404 for the same row and no bytes. Happy to change the expectation if you would rather the read routes 403 too, but that looked like a decision already made.

**3. `TestTheDownloadHeader` duplicated `dev` and its docstring was untrue.**

Correct on both counts — it called `build_content_disposition` directly and could not have caught a route regression, and `test_http_utils.py` already covers the helper three ways. Deleted. The replacement is a route test with a Cyrillic `LibraryFile.filename` that asserts the response header encodes as latin-1, which is what the class claimed to do.

**Minor — `safe_download_filename`.**

Taken as a change rather than a comment; it was a one-line omission with a precedent I had not looked for. The name now goes through `safe_download_filename` before `build_content_disposition`, matching archives.py:2495 and printers.py:2075. There is a test with a newline in an external filename.

**Verification.** `./test_backend.sh` — 10729 passed. `ruff check` and `ruff format --check` clean. `mesh_export.py` still 100%; the route is covered by the 12 new cases.

One note on scope, since you flagged `download_file` has the same exposure through Starlette: I have left it alone. It is a pre-existing path, unrelated to this issue, and folding it in would make this PR about two things. Say the word and I will open it separately.

sadontsev:
Thanks — and an apology for the confusion on the state of this one. The branch you
reviewed at 10:57 was not the branch that was on `origin` at 12:30; I force-pushed the
fixes without saying so, which left your review pointing at code that no longer
existed. That is on me.

So most of this was already addressed before your review landed, and I would rather
say that plainly than re-present it as new work:

**1. Cache-Control.** Now `private, no-cache` with an `ETag` built from `fs_modified_at`
and `file_size`, and an `If-None-Match` check that answers 304 *before* the parse — so a
revalidation costs a database row rather than seconds of CPU. Your read of the external
case is exactly why: `scan_external_folder` keeps the row for a path it already tracks
and refreshes `fs_modified_at`, so the id survives a replacement and the geometry does
not.

**2. Route tests.** Ten of them now, plus two in `test_ownership_permissions.py`. The
ownership pair is there — the cross-user one asserts **404**, not 403, and says why in
its docstring: `_ensure_library_file_visible` answers 404 for a row the caller may not
see, so the endpoint does not confirm which ids exist. The sibling delete/update routes
403 because they gate on the WRITE permission before fetching the row; a read gates on
visibility. If you would rather it were 403 for consistency, say so and I will change
the helper rather than special-case this route.

**3. `TestTheDownloadHeader`.** Gone. You were right that it tested a helper `dev`
already covers and that its docstring claimed something untrue. Replaced by two route
tests: a non-ASCII `LibraryFile.filename` asserting the header encodes as latin-1, and a
newline in an external name, which is the case `safe_download_filename` exists for.

**4. The filename.** Through `safe_download_filename` before
`build_content_disposition`, same order `archives.py:2495` and `printers.py:2075` use.

**The one thing that was genuinely still open is the unbounded input**, and I have taken
it as a change rather than a comment — `05c71020`. A GET any reader can drive in
parallel, holding the whole model in RAM in the shared executor, is a bound worth having
even though `stl_thumbnail` gets away without one by running at upload time. Checked off
`stat()` before the read, 200 MB, reusing `makerworld._MAX_3MF_BYTES`'s number rather
than inventing a second one. Two tests: refused before trimesh is reached, and the bound
is inclusive.

Left alone deliberately: `.ply` being upload-only, the service-level `.gcode.3mf` guard
being unreachable for managed files, and `stem[: -len(".gcode")]`. All three read as
correct-but-asymmetric to me too, and I would rather not churn them inside this PR.

Merged current `dev`. ruff clean, 8641 unit and the 29 mesh cases passing.

One unrelated flag: `test_download_deadline_and_gate_2957.py::
test_a_late_transport_error_is_not_logged_as_loop_noise` fails for me on a clean
`upstream/dev` worktree as well as on this branch, so it is not from this work. It
asserts around a 0.05s cap, so it may just be a slow machine — but if CI is green on it,
that difference is worth knowing.

Issue #2938, opened 2026-08-23
[Bug]: test_nothing_derivable_before_the_first_midnight fails for 31 minutes a day — the snapshot is offset from the wall clock, not the local midnight it is testing


ojimpo:
### Product

Bambuddy

### Area

Other / not sure

### Bug Description

`backend/tests/unit/services/test_plug_energy_history.py::test_nothing_derivable_before_the_first_midnight` fails on `dev` for **31 minutes of every day** and passes for the other 23.5 hours.

I hit it running the full suite for the #2906 PR at 22:10 UTC, checked it was not mine, and measured the window rather than guessing at it.

**It is not my branch.** With my changes stashed, on clean `dev` @ `537b4d25`:

```
FAILED backend/tests/unit/services/test_plug_energy_history.py::test_nothing_derivable_before_the_first_midnight
1 failed, 7 passed in 2.56s

E   assert 1.5 is None
```

Four earlier full-suite runs today, between 08:00 and 13:00 UTC, were green.

**The mechanism.** The module pins the reporter's timezone in an autouse fixture:

```python
@pytest.fixture(autouse=True)
def berlin(monkeypatch):
    monkeypatch.setenv("TZ", "Europe/Berlin")
```

so an outer `TZ` makes no difference — I ran it under `TZ=UTC`, `TZ=Asia/Tokyo` and `TZ=Europe/Berlin` and all three failed identically.

The test then writes its snapshot at a raw wall-clock offset:

```python
now = datetime.now(timezone.utc)
# Snapshot taken this morning, after midnight — no baseline for the day.
await _snapshot(db_session, plug.id, now - timedelta(minutes=30), 102.0)
```

The comment is the premise, and it is only true away from the boundary. Between local midnight and 00:30, `now - 30 minutes` is *before* local midnight, so it is a perfectly good baseline: `_counter_at(midnight_today)` finds it and `today` comes back `103.5 - 102.0 = 1.5` instead of `None`.

I measured the window with the real `local_day_start`, one minute at a time:

```
window (UTC, CEST): 22:00 .. 22:30  (31 minutes)
```

which is 00:00–00:30 Berlin local — and 22:10 UTC is when I ran the suite.

**The production code is right.** A snapshot from before local midnight *is* a valid baseline for today, and `derive_today_yesterday` treats it as one. Only the test's premise breaks at the boundary.

[...]

Issue #2903, opened 2026-08-22
[Bug]: Enabling Spoolman from Settings skips extra-field registration — first AMS sync always fails with HTTP 400


ojimpo:
### Environment

- Bambuddy 1.2.5.3 (`ghcr.io/maziggy/bambuddy@sha256:c670164aaa3b0c5af715ca00e9745cdf3a4d7d337fc11d96fc85180371952698`)
- Spoolman 0.26.1
- Printer: Bambu Lab X2D, firmware 01.02.00.00, one AMS
- Both Bambuddy and Spoolman were **fresh instances with empty databases**.

### What happens

Enable Spoolman in Settings → Filament Tracking, set the URL, then run "Sync AMS Data". Every slot fails:

```
Synced 0 spools with 3 errors
OJIMPO-X2D-01 AMS 0:0: Spoolman rejected spool creation (HTTP 400)
OJIMPO-X2D-01 AMS 0:1: Spoolman rejected spool creation (HTTP 400)
OJIMPO-X2D-01 AMS 0:2: Spoolman rejected spool creation (HTTP 400)
```

Vendor and filament creation succeed — only spool creation fails. From Spoolman's access log:

```
POST /api/v1/vendor    200
POST /api/v1/filament  200
POST /api/v1/spool     400
```

Spoolman's response body is:

```json
{"message": "Unknown extra field tag."}
```

### Root cause

Spoolman rejects any spool payload containing an `extra` key that has not been registered
as a custom field. Bambuddy stores the tray UUID in `extra.tag`, so the `tag` field has to
exist in Spoolman before the first spool is created.

Bambuddy already has `ensure_tag_extra_field()` for exactly this
(`backend/app/services/spoolman.py:935`), but it is only reached from two places:

| Path | Location | Registers fields? |
|---|---|---|
| `POST /api/v1/spoolman/connect` | `backend/app/api/routes/spoolman.py:145` | yes |
| application startup, when `spoolman_enabled` is already true | `backend/app/main.py:8069` | yes |
| **`PUT /api/v1/settings/spoolman`** (enabling it from the Settings UI) | `backend/app/api/routes/settings.py:521-577` | **no** |

The Settings UI enables Spoolman through `PUT /settings/spoolman`, and that handler contains
no `ensure_*` call at all.

So the natural first-run sequence — enable Spoolman in Settings, then sync — always fails on
a fresh Spoolman instance. Restarting Bambuddy afterwards makes it work, because the startup
[...]

PR #2939, opened 2026-08-23
Keep the consumed-counter reset out of Spoolman's own numbers (issue #2906)

### LINKED ISSUE #2906: [Bug]: "Reset usage to 0" on a Spoolman-backed spool wipes the measured remaining weight — the spool jumps back to full
### Product

Bambuddy

### Area

Filament / AMS / Spoolman

### Bug Description

In Spoolman mode, the Inventory page's **Reset counter** action (the one next to the
"Total Consumed / Since tracking started" card) does not just zero the displayed counter — it
**resets the spool's remaining weight back to full in Spoolman**. Any measured remaining weight
is destroyed, silently, with no confirmation that mentions weight.

The internal-mode endpoint gets this right. In #1644 you explained exactly why:

> /reset-usage only ever stamped weight_used_baseline = weight_used, so the displayed
> "Total Consumed" widget (weight_used - baseline) zeros while the lifetime weight_used field
> stays at whatever it was. That's the right shape — **zeroing weight_used would falsely make
> remaining (label_weight - weight_used) jump back to "full" even though you've already pulled
> filament out**

The Spoolman-mode path does precisely that. `services/spoolman.py:658`:

```python
async def reset_spool_usage(self, spool_id: int) -> dict:
    """Reset a spool's used_weight to 0 in Spoolman. ..."""
    response = await self._request_spool(
        "PATCH",
        spool_id,
        json_body={"used_weight": 0},
        operation="reset-usage",
    )
    return response.json()
```

Spoolman derives `remaining_weight = initial_weight - used_weight`, so `used_weight = 0`
means `remaining_weight = initial_weight` — full spool.

The caller's docstring (`api/routes/spoolman_inventory.py:930`) claims the opposite:

```
... so the Inventory page's `weight_used - baseline` display lands at 0 while remaining
(= label - weight_used) is preserved — parity with the internal-mode endpoint (#1390, see also
backend/app/api/routes/inventory.py::reset_spool_consumed_counter).
```

`remaining` is not preserved. The display lands at 0 only because both sides collapse together:
after the reset `remaining == label`, so `weight_used = label - remaining = 0` and
`baseline = 0 - 0 = 0`. The counter reads 0 for the wrong reason, and the spool now claims to
be full.

### Expected Behavior

Resetting the consumed counter should zero the displayed "Total Consumed" figure and leave
`remaining_weight` exactly as it was — same semantics as internal mode's
`weight_used_baseline`, as described in #1644.

### Steps to Reproduce

Reproducible directly against the Spoolman REST API (this is what Bambuddy sends):

1. Spoolman mode enabled, at least one spool with a known remaining weight.
2. Note the spool's state:

   ```bash
   curl -s http://<spoolman>/api/v1/spool/<id>
   # initial_weight=1000.0  remaining_weight=737.0  used_weight=263.0
   ```

3. Send the same request `reset_spool_usage` sends:

   ```bash
   curl -X PATCH http://<spoolman>/api/v1/spool/<id> \
     -H 'Content-Type: application/json' -d '{"used_weight": 0}'
   ```

4. Observe:

   ```
   initial_weight=1000.0  remaining_weight=1000.0  used_weight=0.0
   ```

   The spool is now "full". 737 g of measured remaining weight is gone.

Equivalently, from the UI: Inventory → any Spoolman-backed spool → **Reset counter**.

| | initial_weight | remaining_weight | used_weight |
|---|---|---|---|
| before | 1000.0 | **737.0** | 263.0 |
| after  | 1000.0 | **1000.0** | 0.0 |

- Read the running container's source (`docker exec bambuddy`) rather than GitHub HEAD, so the
  line numbers above are from the exact build that reproduces (1.2.6b1, daily).
- Searched open AND closed issues for `reset consumed counter` / `reset usage`. Found **#1644**
  (closed 2026-06-05), which is the internal-mode version of this and where the correct
  semantics were established — see the "Relationship to past issues" section below.
- Verified against Spoolman 0.26.1's own REST API with a throwaway spool, so the result is not
  an artifact of Bambuddy's read mapping. The throwaway spool was deleted afterwards.
- Confirmed the workaround below restores the intended behaviour on all 8 of my spools.

### Printer Model

X2D

### Bambuddy Version

### Printer Firmware Version

01.02.00.00

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

No support package. Debug logging is deliberately off on this host (it produced ~790 MB/day and
filled 34 GB on another instance of mine), and this defect leaves no log trace — the request
succeeds and returns 200.

The evidence above is stronger than a log bundle for this bug: it is a direct before/after probe
of Spoolman's own REST API using the exact request body Bambuddy sends, plus the source of the
running build. Happy to attach a debug-logged run if you still want one.

### Additional Context

**Suggested fix — one line.** Pin `initial_weight` to the current remaining weight in the same
PATCH:

```python
async def reset_spool_usage(self, spool_id: int) -> dict:
    current = await self.get_spool(spool_id)
    remaining = current.get("remaining_weight")
    response = await self._request_spool(
        "PATCH",
        spool_id,
        json_body={"initial_weight": remaining, "used_weight": 0},
        operation="reset-usage",
    )
    return response.json()
```

Verified on Spoolman 0.26.1:

```
before: initial=1000.0 remaining=737.0 used=263.0
PATCH  {"initial_weight": 737, "used_weight": 0}
after:  initial=737.0  remaining=737.0 used=0.0
```

Bambuddy's own read mapping then produces the intended display without any further change:

```
weight_used = label(1000) - remaining(737) = 263
baseline    = 263 - spoolman_used(0)       = 263
displayed   = 263 - 263 = 0        # counter zeroed
remaining   = 737 / 1000 = 73.7%   # label-based, still correct
```

`initial_weight` then means "how much filament this spool had when tracking started", which is
what the "Since tracking started" subtitle already promises.

I'm happy to open a PR for this if you'd like — say the word and whether you'd prefer the
`get_spool` round-trip above or passing the already-known `remaining_weight` down from
`spoolman_inventory.py` (the route already holds the spool dict in the bulk path).

**Relationship to past issues (so this doesn't re-open a closed problem)**

- **#1644** (closed 2026-06-05, 0.2.5b1) — same button, **internal mode**. That issue was about
  the endpoint *name* being misleading and was fixed by renaming `/reset-usage` to
  `/reset-consumed-counter`; the internal implementation was already correct
  (`weight_used_baseline`). The Spoolman-mode implementation added later never adopted that
  shape. **This report is the Spoolman path, still broken in 1.2.6b1.**
- **#1390** — cited in the current docstring. That was about an *edit* to an unrelated field
  overwriting `remaining_weight` after a reset. That fix is in place and is not what this is.
  This is the reset call itself.
- **#1119 / #1793 / #1820 / #1276** — my earlier Spoolman reports, all about *print-time*
  consumption attribution. This one is not print-related; it fires on a manual UI action.

ojimpo:
## Description

"Reset usage to 0" PATCHed `used_weight = 0`. That is not a baseline: Spoolman derives `remaining_weight` as initial minus used, so zeroing the used weight put the spool back to full and threw away the measured remaining filament. A roll with 737 g left read 1000 g, and the next print charged against a spool Bambuddy believed was untouched.

The baseline lives in `spool.extra` now — the mechanism internal mode has always used for its `weight_used_baseline` column. `_map_spoolman_spool` folds the stored value into the baseline it reports, because the frontend's arithmetic is fixed at `weight_used - weight_used_baseline` and that is where a baseline has to arrive.

### Why not rewrite `initial_weight`

That was my proposal on the issue and you were right to refuse it. It repairs Bambuddy's display by overwriting a field the user owns: `initial_weight` goes 1000 → 737 with no way back, ratchets down again on every subsequent reset, and Spoolman's own views then compute 737/737 and show the spool as full. For someone running Spoolman *because* Spoolman is their system of record, that moves the wrongness rather than removing it — and Bambuddy quietly rewriting a user-owned field is the complaint the issue is about.

Writing to `extra` touches no native field. `initial_weight`, `remaining_weight` and `used_weight` all survive, so nothing moves on the Spoolman side at all. `test_reset_does_not_touch_any_native_spoolman_field` pins that.

### The copy needs no change

`en.ts:4853` already promises "The spool itself, its remaining weight calculation, and your settings are not changed." You flagged that as an explicit assurance given immediately before the wipe, in thirteen locales. **This fix is what makes it true** — so no locale edits, and the string that was the strongest evidence of the bug becomes the specification.

### Things that came free or turned out simpler

- **Registration.** `_ensure_extra_fields` has been payload-keyed since #2903, so the write carrying `bambu_weight_used_baseline` registers it on its own. No list to update, exactly as you said.
- **The bulk path.** You noted that `bulk_reset_spool_consumed_counter` iterates bare ids and cannot hand a spool dict down. It does not need to: the client method fetches the spool itself, which it needs anyway to read the used weight it is about to record. The bulk route is a one-word change.
- **Concurrency.** The reset takes the same per-spool `extra_lock` as `merge_spool_extra`, so a concurrent tag or colour write cannot read the extra dict between the fetch and the PATCH and put the old one back.

### `remaining_weight` being None

Guarded, as you asked. Legacy spools and spools with a filament linked but never primed carry `None`, and the old code PATCHed straight through it. The reset now records the baseline without touching the field, and the read mapping's existing `remaining_raw is None` branch takes the stored baseline too, so the counter still reads 0. `test_reset_survives_a_spool_with_no_remaining_weight` covers both halves.

### What this does not repair

**A spool already wiped by the old implementation stays wiped.** Its `remaining_weight` was overwritten in Spoolman and the measured value is gone — there is nothing left in either database to recover it from. Those spools need re-weighing. I would rather say so than ship a migration that guesses.

## Related Issue

Fixes #2906

## Documentation

**Pick one**:
- [x] No docs update required — reason: bug fix with no new setting, endpoint or UI element. The button, its copy and the API shape are unchanged; the stored representation behind them is what moves, and the wiki does not describe it.

## Type of Change

## Changes Made

- `backend/app/services/spoolman.py` — `reset_spool_usage` (PATCH `used_weight = 0`) replaced by `reset_spool_consumed_counter`, which records the current used weight under `bambu_weight_used_baseline` in `spool.extra` under the per-spool extra lock. The old method is removed rather than left in place; nothing else called it and what it does is the bug
- `backend/app/api/routes/_spoolman_helpers.py` — `_extract_extra_float` (the numeric twin of `_extract_extra_str`), which accepts both the string and the number spelling so an install that already stored the number form keeps working; `_map_spoolman_spool` folds the stored baseline into `weight_used_baseline` in both the remaining-known and remaining-None branches, clamped at both ends
- `backend/app/api/routes/spoolman_inventory.py` — the single and bulk routes call the new method; docstring corrected, since it described the old derivation as deliberate
- `backend/tests/unit/services/test_spoolman_reset_baseline_2906.py` — thirteen tests against a fake Spoolman that enforces three of the real server's rules, not one
- `backend/tests/integration/test_spoolman_inventory_api.py` — fixture correction, below, and the staged baseline moved to the string form for the same reason

## Screenshots

## Testing

- [x] I have tested with my printer model: **X2D**

### Review round two

All four points taken. The blocking one first.

**The write is stored as a JSON string now, `json.dumps(str(baseline))`.** You were right that it never landed: Spoolman registers an unseen extra key as `field_type` `"text"`, and `validate_extra_field_value` then requires the value to decode to a `str`, so the JSON number `263.0` came back "Value is not a string." and the PATCH 400'd. I had read `_ensure_extra_fields` registering the key as sufficient and never asked what type it registered it *as*.

The part that makes this worse than a 400 is the one I want to confirm I have understood: `add_or_update_extra_field` refuses to change a field's type, so a single numeric write pins the key to `text` on that install forever and no later release can repair it. The storage type has to be right the first time, which is why this is a string rather than a float field.

**The read side takes the other half.** `_extract_extra_float` decoded `'"263.0"'` to the `str` `'263.0'`, failed the `isinstance` check and returned `None` — so the string form would have been written correctly and then silently ignored. It now tries `float(decoded)` when the decode yields a `str`, and still returns `None` for anything that is not a number. Both spellings read alike, so an install that already stored the number form is not stranded.

**`or 0.0` is gone.** Spelled out as `0.0 if stored is None else max(0.0, stored)`, which keeps the distinction the docstring claims and closes the negative case in the same line.

**The negative clamp.** Worth reporting that my first test for this was vacuous — I staged a spool where `used_weight` and `remaining_weight` agree, and there the outer `max(0.0, ...)` absorbs the negative on its own and the test passes with or without the fix. It only bites when the two disagree, which is exactly when the baseline is carrying a real correction rather than a cancelling pair. The test now stages `used_weight` 263 against a `remaining_weight` saying 300 g has gone: without the clamp the 37 g correction is dropped and consumed reads 300 instead of 263. That is the same mistake you caught in the original fixture, so I checked it the way you checked mine rather than assuming.

**The fake now applies three rules instead of one.** Registration is real — `GET /field/spool/<name>` 404s for an unregistered key, `POST` creates it, and a second `POST` with a different type is refused with "Field type cannot be changed." Extra values are validated against the registered type. The four keys Bambuddy writes today are pre-registered as `text` and `bambu_weight_used_baseline` is not registered at all, which is the state you described on your instance.

Putting the old numeric write back against that fake:

```
11 failed, 2 passed
```

Reverting only the read-side `str` branch:

```
6 failed, 7 passed
```

Reverting only the negative clamp:

```
1 failed, 12 passed
```

**The fixture correction.** `test_spoolman_inventory_api.py:66` mocked the reset as `{**SAMPLE_SPOOLMAN_SPOOL, "used_weight": 0}` — `used_weight` 0 next to `remaining_weight` 750.0, which real Spoolman cannot return since it recomputes remaining from initial minus used. Against a real server that is 1000.0, and the assertion at line 590 (`label_weight - weight_used == 750`, "remaining unchanged") fails. The one check that would have caught this was cancelled out by its own fixture. It now returns what the endpoint actually produces: every native field unchanged, with the baseline in `extra`.

**And the same rule is enforced rather than mocked.** Following the shape of the #2903 tests, the new unit tests drive the real `SpoolmanClient` against a fake Spoolman that recomputes `remaining_weight = initial_weight - used_weight` on every PATCH. A reset that writes `used_weight` cannot pass there. **All eight of the original tests fail against `dev`:**

```
assert 1000.0 == 737.0                                    test_reset_leaves_the_measured_remaining_weight_alone
assert 0 == 263.0                                         test_reset_does_not_touch_any_native_spoolman_field
assert 'bambu_weight_used_baseline' in {'tag': ...}       test_reset_records_the_baseline_in_extra
                                                          test_reset_preserves_other_extra_keys
assert (1000.0 - 0.0) == 737.0                            test_displayed_consumed_reads_zero_while_remaining_holds
assert (313.0 - 0.0) == 50.0                              test_consumption_after_a_reset_counts_from_the_baseline
                                                          test_reset_survives_a_spool_with_no_remaining_weight
                                                          test_second_reset_moves_the_baseline_forward
```

The first line is the reported symptom exactly: the spool jumping back to full.

Current numbers, on this branch:

```
the two touched test files together      179 passed
backend/tests/unit + that integration    8198 passed, 22 skipped
ruff check                               All checks passed!
ruff format --check                      820 files already formatted
```

**The #2938 note from the previous revision is dropped.** You confirmed it is fixed on `dev` in `ed84f0f74`, and the run above is clean, so there is no failure left to explain. Thank you for picking it up.

## Checklist


maziggy (changes requested):
The design call is right and the write-up made this quick to check. One blocking
problem: the write Spoolman actually receives is rejected, so the reset never
lands.

_ensure_extra_fields registers any unseen key via ensure_extra_field(name), whose
field_type defaults to "text". You send json.dumps(263.0), which is the JSON
number 263.0, and Spoolman's validate_extra_field_value requires a text field to
decode to a str -- "Value is not a string." I ran your payload through Spoolman's
own validator: '263.0' and '0.0' both rejected, '"263.0"' accepted. My instance
has no bambu_weight_used_baseline registered, and all four keys Bambuddy writes
today are text.

Sequence against a fake that applies both of Spoolman's real rules:

  GET  /spool/42
  GET  /field/spool/bambu_weight_used_baseline   -> 404
  POST /field/spool/bambu_weight_used_baseline   -> created as "text"
  PATCH /spool/42                                -> HTTP 400, extra unchanged

This matters more than a normal 400, because add_or_update_extra_field raises
"Field type cannot be changed." The first reset on an install registers the key
as text permanently, so a later fix that registers it as float cannot repair that
install. The storage type has to be right the first time.

Please store the string form, json.dumps(str(baseline)) -- what tag,
bambu_color_name and both slicer keys already do, and what a text field accepts.
The read side needs the other half: _extract_extra_float decodes '"263.0"' to the
string '263.0', fails the isinstance check and returns None, so have it try
float(decoded) when the decode yields a str.

On the tests: the fake answers GET /field/spool/<anything> with 200, so
registration never runs, and it never type-checks extra. Same shape as the
fixture you are fixing -- it models the rule that bit last time, not the one that
bites this time. Teaching it those two rules is what stops this repeating, and
test_reset_records_the_baseline_in_extra currently pins the rejected format.

Two small ones while you are in there. _extract_extra_float(...) or 0.0 collapses
None and 0.0, which the docstring says the function exists to keep apart --
harmless arithmetically, but pick one. And a negative stored baseline inflates
the displayed counter: the write clamps to >= 0, the read does not, and
min(baseline, used_weight) only bounds the other end. A max(0.0, stored) closes
it.

The rest holds up, and I checked the parts I expected to break. extra over
rewriting initial_weight is the right call. update_spool_full sends only the
fields passed and does not take extra_lock, so nesting inside it is safe. The
arithmetic gives real_used - stored_baseline in both branches and survives a
re-weigh through sync_spool_weight. The new min() clamp is a no-op for spools
that were never reset, so no regression. The fixture correction is real. And one
you did not mention: ForecastPanel drops usage history for any spool with a
non-zero baseline, so Spoolman resets now stop polluting the forecast rate the
way internal-mode ones already did.

Suite on your branch merged with dev: 11050 passed, ruff clean. #2938 is fixed on
dev in ed84f0f74, so that failure is gone.

maziggy:

ojimpo:
Conflict resolved and the branch is up to date with `dev`.

**I merged rather than rebased, and I want to flag that explicitly since you asked for a rebase.**
`4d736133` on this branch is your own `Merge branch 'dev'` from 08-25. Rebasing would drop it
and the push would have to be a force push over a commit you made, so I took the merge. Say the
word if you would rather have the linear history and I will rebase and force push — it is your
branch state I would be rewriting, which is why I did not decide it on my own.

The conflict itself was purely additive: your `parse_spoolman_multi_colors` and my
`_extract_extra_float` / `BAMBU_WEIGHT_USED_BASELINE_KEY` landed at the same point at the end of
the extra-field helpers in `_spoolman_helpers.py`. Both are kept, with `_extract_extra_float`
next to `_extract_extra_str` it mirrors and your function following unchanged.

**One test needed a real change, not just conflict resolution.**
`test_the_baseline_key_is_registered_before_it_is_written` asserted
`"GET /field/spool/bambu_weight_used_baseline" in fake.log`. `dev` changed how
`_ensure_extra_fields` decides a key is missing — it probed the key directly when I wrote that,
and lists the fields now (#2983) — so the test failed on a change that does not touch the
behaviour it exists to cover. It was pinning the shape of the probe, which was never the claim.

What it is actually about is the ordering: Spoolman answers 400 for an extra key it was never
told about, so registration has to land before the PATCH that carries the value. That is what it
asserts now, by index into the request log rather than by membership, so the two cannot be
satisfied out of order:

```python
registered = fake.log.index("POST /field/spool/bambu_weight_used_baseline")
written = fake.log.index("PATCH /spool/42")
assert registered < written
```

Measured, on the merged tree:

```
head                                                          13 passed
_ensure_extra_fields call removed from update_spool_full      11 failed, 2 passed
```

The mutation fails the test above along with ten others in the file — `FakeSpoolman` rejects the
unregistered key with a 400, which is the live failure this covers.

Full backend unit suite on this branch with the `dev` merge under it: **8623 passed, 22 skipped**.

maziggy (changes requested):
Reviewed in full. The approach is right and I checked the parts that could bite rather than taking the description's word for it.

I drove the old implementation and the round-two numeric write against your FakeSpoolman: the first returns remaining 1000 where 737 went in, the second 400s. The fake reproduces both, so these tests will keep catching them. ensure_extra_field defaulting to field_type "text" is confirmed in the client, which is what makes the string encoding mandatory rather than stylistic.

Traced the baseline through every other write path. update_spool sends extra=None unless the tag is being cleared, both tag paths re-fetch and merge inside the lock, and spool creation builds a fresh extra dict, so a copy does not inherit a baseline. The edit form binds weight_used to the mapped value and only sends it when touched, so an edit after a reset cannot reintroduce the original bug. No caller of reset_spool_usage survives anywhere.

One change before merge: BAMBU_WEIGHT_USED_BASELINE_KEY is declared in both services/spoolman.py and _spoolman_helpers.py with nothing asserting they match. Your comment explains why the duplication is there and I agree with the reasoning, but a divergence fails silently - the reset returns 200, the key lands, and the counter just never zeroes. Since the storage type cannot be changed after the first write, please add the one-line assert that the two are equal.

One observation, no action needed. ForecastPanel drops usage history for spools with a non-zero baseline. Under the old implementation a Spoolman reset left the mapped baseline at 0, so pre-reset events kept inflating the forecast rate; this fix quietly repairs that too and brings Spoolman in line with internal mode.

Last, a wording nit rather than a code one. The clamp comment calls a baseline above used_weight "the user's edit", but the AMS sync writes remaining_weight from the AMS percentage, so used_weight can drop below the baseline with no user involvement and the counter then reads 0 until it climbs back. Same behaviour as internal mode and much better than before, when the next sync undid the reset outright - worth saying so in the comment.

Thanks for the round-two writeup, particularly for reporting that your first negative-clamp test was vacuous. That is the hard thing to say and it is why this one is worth trusting.

Issue #2602, opened 2026-07-18
Queue upload concurrency is batch-scoped and does not refill free slots


Jostxxl:
## Summary

`queue_max_concurrent_uploads` is currently applied only to the items selected in one scheduler pass. `check_queue()` then awaits `_dispatch_selected()`, which awaits `asyncio.gather()` until the slowest FTP upload in that batch finishes.

As a result, unused concurrency slots are not refilled while any upload from the current batch remains active. On a large farm this makes the setting behave like a per-batch cap rather than a continuously refillable worker pool.

## Environment

- Current `dev` build: `059b733c`
- Native systemd deployment with PostgreSQL
- 93 configured printers
- `queue_max_concurrent_uploads=16`
- Prinjekt farm; printers are routinely powered on/off

## Reproduction / live evidence

The service correctly loaded the setting:

```
17:47:00 Dispatching 2 queue item(s) with up to 16 concurrent upload(s)
18:04:13 Dispatching 1 queue item(s) with up to 16 concurrent upload(s)
18:19:45 Dispatching 1 queue item(s) with up to 16 concurrent upload(s): [177]
```

Queue item 177 uploaded a 95,988,809-byte 3MF for 513.6 seconds:

```
18:19:45 FTP upload starting ... heart_3.3mf
18:28:25 FTP upload complete ... 95988809 bytes in 513.6s
```

Only after that upload completed did the next two eligible items start:

```
18:28:28 Dispatching 2 queue item(s) with up to 16 concurrent upload(s): [174, 175]
18:28:30 Queue item 174: FTP upload starting
18:28:30 Queue item 175: FTP upload starting
```

For those 8.5 minutes, 15 configured upload slots were unused. Jobs becoming eligible during the long upload could not be selected.

## Root cause

`check_queue()` ends with:

```python
if dispatch_ids:
    await self._dispatch_selected(dispatch_ids, upload_limit)
```

and `_dispatch_selected()` awaits the entire batch:

```python
results = await asyncio.gather(...)
```

The comment says this preserves the no-overlapping-pass invariant because rows remain `pending` until upload completion. That safety property is valid, but it couples scheduling progress to the slowest FTP server.

## Expected behavior

Treat the configured limit as a global refillable pool:

1. Reserve selected queue rows against duplicate dispatch.
2. Track each active upload independently.
3. Return control to the scheduler while uploads continue.
4. On each fast scheduler tick, fill any free slots up to the configured limit.
[...]

PR #2940, opened 2026-08-23
Give the Low Filament notification something that fires it (issue #2913)

### LINKED ISSUE #2913: [Bug]: "Low Filament" notification can never fire -- on_filament_low has no call site
### Product

Bambuddy

### Area

Notifications

### Bug Description

`NotificationService.on_filament_low()` exists, is wired through the model, the schemas, the API
route, the notification template system and the UI -- but **nothing in the backend ever calls it**.
Enabling the "Low Filament" toggle on a notification provider persists fine and shows a badge on
the provider card, and the wiki lists it as a supported trigger, but the notification can never be
sent.

The provider model has 31 event columns (`backend/app/models/notification.py`). Enumerating the
call sites in the running 1.2.6b1 image gives 26 distinct producer methods:

```
$ docker exec bambuddy sh -c \
    'grep -rn "notification_service\.on_" /app/backend/app/ | grep -v pycache | grep -v "def on_"' \
  | sed 's/.*notification_service\.\(on_[a-z_]*\).*/\1/' | sort -u
on_ai_failure_detection
on_ams_drying_suspended
on_ams_ht_humidity_high
on_ams_ht_temperature_high
on_ams_humidity_high
on_ams_temperature_high
on_bed_cooled
on_billing_charge_failed
on_first_layer_complete
on_ha_sensor_alert
on_maintenance_due
on_plate_clear_required
on_plate_not_empty
on_print_complete
on_print_missing_spool_assignment
on_print_progress
on_print_start
on_printer_error
on_printer_offline
on_queue_completed
on_queue_job_added
on_queue_job_assigned
on_queue_job_failed
on_queue_job_skipped
on_queue_job_started
on_queue_job_waiting
```

Diffing that against the 31 columns leaves five names:

```
on_filament_low
on_print_failed
on_print_stopped
on_stock_break_alert
on_stock_reorder_alert
```

`on_print_failed` and `on_print_stopped` are fine -- they have no method of their own because
`on_print_complete()` selects the event field from the print status
(`notification_service.py:1213-1223`: `completed` -> `on_print_complete`, `failed` ->
`on_print_failed`, `aborted`/`stopped`/`cancelled` -> `on_print_stopped`).

That leaves three columns with no producer at all: **`on_filament_low`**, and the two stock
alerts (which are a separate case -- see Additional Context).

What *is* present for `on_filament_low`:

| Layer | Location | State |
|---|---|---|
| Provider column | `backend/app/models/notification.py:75` | present |
| Create / Update schema | `backend/app/schemas/notification.py:52`, `:150` | present |
| API route mapping | `backend/app/api/routes/notifications.py:52`, `:136` | present |
| Service method | `backend/app/services/notification_service.py:1566` | present, **never called** |
| MQTT relay method | `backend/app/services/mqtt_relay.py:661` | present, **never called** |
| Default template | `backend/app/models/notification_template.py:101` | present (`"{printer}: Slot {slot} at {remaining_percent}%"`) |
| Template preview variables | `backend/app/schemas/notification_template.py:79`, `:192` | present |
| i18n (EN / DE) | `backend/app/i18n/__init__.py:25`, `:57` | present |
| UI toggle + badge | `frontend/src/components/AddNotificationModal.tsx`, `NotificationProviderCard.tsx` | present |
| Wiki | https://wiki.bambuddy.cool/features/notifications/ -- "Low Filament: Filament running low" | documented |
| **Producer** | -- | **missing** |

There is also no threshold anywhere to decide what "low" means: searching the backend for
`filament_low_threshold` and `low_filament` returns nothing. So there is no configuration entry
point for the feature either.

This is not a regression I can date -- as far as I can tell the producer was never written.

### Expected Behavior

Either of these would be fine, and I do not have a preference:

1. Something in the backend calls `notification_service.on_filament_low(...)` when a tray's
   remaining filament crosses a threshold, with a setting for that threshold, or
2. If the event is not implemented, the toggle is removed from the provider UI (and from the wiki)
   so it does not read as a working feature.

What should not happen is the current state: the toggle persists, shows a badge, has an editable
notification template, and is documented -- but is silently inert.

### Steps to Reproduce

I am deliberately giving the static reproduction rather than a runtime one. There is no code path
that evaluates "low", so a runtime observation can only ever be "the notification never arrives",
which is weak evidence. The absence of the call site is the actual defect and is exact:

1. Check out `main` (or exec into the running container).
2. Enumerate the producers:

   ```bash
   grep -rn "notification_service\.on_" backend/app/ \
     | grep -v "def on_" \
     | sed 's/.*notification_service\.\(on_[a-z_]*\).*/\1/' | sort -u
   ```

   Diff that against the event columns in `backend/app/models/notification.py`. Five names have
   no producer method: `on_filament_low`, `on_print_failed`, `on_print_stopped`,
   `on_stock_break_alert`, `on_stock_reorder_alert`. The two print ones are dispatched inside
   `on_print_complete()`, so the real gaps are `on_filament_low` and the two stock alerts.
3. Confirm the method itself exists and is reachable only from its own definition:

   ```bash
   grep -rn "on_filament_low" backend/app/
   ```

   Results are the model column, the schemas, the route mapping, the service definition and the
   MQTT relay definition. No caller.
4. Confirm no dynamic dispatch could reach it:

   ```bash
   grep -rn "getattr(notification_service" backend/app/
   ```

   No results.
5. Confirm there is no threshold to trigger it:

   ```bash
   grep -rn "filament_low_threshold\|low_filament" backend/app/
   ```

   No results.

From the user side the toggle is at Settings -> Notifications -> add a provider -> enable
"Low Filament". Reading the code rather than clicking it: the field is in
`NotificationProviderUpdate`, the route maps it, and the card issues
`mutate({on_filament_low: ...})`, so it persists and the badge renders. I have not clicked
through this myself -- see Additional Context.

- Read the running 1.2.6b1 source in the container (`docker exec bambuddy`) rather than the GitHub
  HEAD, so the line numbers above are from the build that is actually running.
- Confirmed the same on `main` via GitHub code search: `on_filament_low` appears in the model,
  schemas, route, service definition, MQTT relay definition, the frontend, and tests -- never in
  `main.py`, `print_scheduler.py` or `bambu_mqtt.py`. Searching `remaining_percent` likewise
  surfaces no producer.
- Checked `backend/tests/unit/services/test_notification_service.py`: the only occurrence is a
  fixture assigning `provider.on_filament_low = False`. There is no behavioural test, which is
  probably why the missing producer has not shown up as a failure.
- Checked the wiki notifications page, which documents the trigger.
- Searched open and closed issues for `filament low notification`, `on_filament_low` and
  `Low Filament`. Nothing matches. The only hits for "Low Filament" are #2669 (a print-options
  overlay clipping filament usage) and #1762 (a feature request about AMS filament matching),
  both unrelated.
- No in-app diagnostic applies -- this is not a connection problem.

### Printer Model

Not printer-related

### Bambuddy Version

1.2.6b1 (`ghcr.io/maziggy/bambuddy:daily`, digest
`sha256:c3938d7ed41e1aa8c6ae0a69d5924d881813f4f3263528ff30dfc3b903401021`, image built
2026-08-19, pulled 2026-08-22). Also verified against `main` as of 2026-08-22.

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian) -- Ubuntu 24.04

No support package, and no logs would help: the defect is that a code path does not exist, so
there is nothing for it to log. Debug logging would not add anything either. The evidence is the
source enumeration above, which reproduces in a clean checkout in a few seconds. Happy to attach
one if you want the surrounding configuration.

### Additional Context

**What I have and have not observed.** Everything above is read from source: the missing call
site, and the fact that the toggle persists and renders a badge. I have *not* enabled the toggle
in the UI, and I have *not* watched a spool run down and observed the absent notification -- my
instance is only now getting its first notification providers, and my API key is denied
`NOTIFICATIONS_CREATE`/`UPDATE`, so I cannot script it either. I am saying so explicitly rather
than implying runtime observations I did not make. The static evidence stands on its own: with no
caller and no threshold, there is no input that could produce this notification.

**This is not a re-run of an existing issue.** I searched open and closed issues (queries listed
under Troubleshooting) and found nothing covering the filament-low trigger. #2669 and #1762 both
mention filament but are a UI clipping bug and an AMS matching feature request respectively.
This is also unrelated to the AMS environmental alarms (#2905), which do fire.

**There is a second, adjacent case, and I would rather ask than bundle it.**
`on_stock_reorder_alert` and `on_stock_break_alert` have the same shape -- implemented at
`notification_service.py:2188` and `:2213`, toggles in the provider UI, no caller anywhere. They
are worse in one respect: unlike `on_filament_low`, they are absent from
`NotificationProviderBase` / `NotificationProviderUpdate` and from `_provider_to_dict()`, so the
values the UI sends are dropped by Pydantic and never appear in the GET response either. Tell me
whether you want that as a separate issue and I will open one rather than growing this one.

**Happy to write the PR.** Which shape do you want?

- **(a) Implement the producer.** The natural site is the AMS tray update path in `main.py`,
  next to the existing AMS alarm checks, using the tray `remain` percentage with a new
  `filament_low_threshold` setting and per-printer/per-slot debouncing along the lines of
  `_notified_hms_errors`. This needs a decision on whether it keys off AMS `remain` (which is
  coarse and, on my machine, was off by up to 56% against a scale) or off Spoolman remaining
  weight when Spoolman mode is on.
- **(b) Remove the toggle** from `AddNotificationModal` / `NotificationProviderCard` and drop the
  wiki line, leaving the column in place for a later implementation.

(a) is more work but matches what the wiki already promises. I would start there unless you would
rather not add another setting.

ojimpo:
## Description

Every layer of this feature existed except the one that produces the event: a column on `notification_providers`, a schema field, a route, a message template, thirteen locales and a UI toggle — and `_get_providers_for_event(db, "on_filament_low", ...)` appearing only inside the method that would be called. The toggle could be switched on and could never fire.

This is the producer, in the shape you specified.

### No new setting

`low_stock_threshold` (default 20%) with the per-spool `low_stock_threshold_pct` override is already exactly this decision — already configurable, already surfaced, already driving the Inventory page's Low Stock count. Reusing it means the alert and the badge cannot disagree about what "low" means.

The AMS `remain` percentage is not consulted, for the reason I gave on the issue: I have measured it up to 56 points out against a scale, which is not a number to page someone on.

### One asymmetry worth naming

You wrote that the threshold is "already meaningful in both modes". The global half is. The per-spool half is not: `low_stock_threshold_pct` is a column on Bambuddy's own `spool` table with no Spoolman equivalent, and `_map_spoolman_spool` does not emit the key at all. So a Spoolman-mode slot takes the global threshold and only that.

I have implemented it that way rather than inventing a Spoolman-side override, because it is what the Inventory page already does there — `s.low_stock_threshold_pct ?? lowStockThreshold` falls through to the global for every Spoolman spool. The two surfaces stay consistent. Say the word if you would rather it lived in `spool.extra` like the other Bambuddy-only fields.

### Where it runs, and why not on the AMS push

The scheduler's 30 s pass, next to `_check_auto_drying`. Resolving a Spoolman slot means asking Spoolman what the spool holds; doing that per MQTT push would be a request storm for a value that only moves when a print finishes. One `get_all_spools` call covers every slot on a pass.

Both paths out of `_check_queue` call it, for the same reason both call `_check_auto_drying`: an empty queue does not mean the spools already in the printers stopped mattering.

### Debounce

`_notified_hms_errors` copied, as you said — a set keyed by `(printer_id, ams_id, tray_id)`, cleared when the slot goes back **above** its threshold rather than on a timer. A spool sitting just under the line stays quiet; a refilled slot can alert again.

Two details that are not obvious from the pattern:

- **The key is added before the send, not after.** A provider that is slow or down must not let the next pass repeat the alert.
- **A pass that resolves no slots at all leaves the set alone.** A Spoolman that was briefly unreachable would otherwise re-alert on every spool the moment it came back.

Slots with no assigned spool produce nothing — there is no remaining weight to stand behind for one, and inventing a figure is how the remain percentage would have got back in.

## Related Issue

Fixes #2913

## Documentation

**Pick one**:
- [x] No docs update required *from me* — reason: you said on the issue you would take the three wiki pages once the producer lands, and I have not touched them. I verified they are the three you named and that the third is phrased as an instruction: `features/notifications.md:382` ("Low Filament | Filament running low"), `features/ams.md:734` ("When filament is running low"), `features/spoolman.md:359-361` ("Notifications for low filament: … Enable **Low Filament** event"). Happy to write them instead if you would rather not.

No code-side docs change: no new setting, endpoint or UI element. The toggle and its copy are unchanged — they start working.

## Type of Change

## Changes Made

- `backend/app/services/print_scheduler.py` — `_check_filament_low`, called from both exits of `_check_queue`; `_filament_low_slots_spoolman` for the Spoolman-mode resolution; `_emit_filament_low` for the per-slot debounce and send; `_get_low_stock_threshold`; module-level `_remaining_percent` and `_ams_slot_label`
- `backend/tests/unit/services/test_filament_low_producer_2913.py` — fourteen tests

## Screenshots

## Testing

- [x] I have tested with my printer model: **X2D**

Fourteen tests: the threshold crossing in both directions, a slot with no assigned spool, the debounce not repeating across passes, re-arming once the slot goes back above, the debounce being per-slot rather than per-printer, the per-spool override beating the global, the global setting being honoured, a failing provider not taking the pass down, and the arithmetic and slot labels on their own (AMS-A T1, HT-A, External).

**Being straight about what these do and do not prove.** The producer is new code, so its tests cannot fail against `dev` — there is no `_check_filament_low` there to call. They pin the behaviour you specified; they are not regression tests, and I would rather say that than present them as if they were.

**The last one is the assertion that was missing.** You identified why this shipped: the only occurrence of `on_filament_low` in `test_notification_service.py` is a fixture setting the flag to `False`, so nothing ever checked the event reaches a provider. `test_the_event_reaches_a_provider_with_the_toggle_on` does — a provider with the toggle on is handed the rendered message and the variables. **It passes against `dev` too**, because the service method was always there; it was the caller that was not. Its absence is why nobody noticed.

**A guard I did not add, and why.** The check that would actually catch the next one of these is an enumeration test — every event column on `notification_providers` has a producer somewhere. I did not write it because it fails today for `on_stock_reorder_alert` and `on_stock_break_alert`, which is the separate plumbing bug we agreed to split out, so it would land red. Worth adding once that is fixed, with the two print events allowlisted since they are dispatched from inside `on_print_complete`. Happy to take it then.

Full backend suite: **10679 passed, 22 skipped**. `ruff check` and `ruff format --check` clean, and no new warnings.

**On my last PR's suite run** — #2939 reported one failure, which I filed as #2938: a test in `test_plug_energy_history.py` that fails for 31 minutes a day. This run was at 22:48 UTC, outside that window, and it passed. That is the same explanation confirming itself rather than a different result.

## Checklist


maziggy (changes requested):
Good producer, and the two design calls at the top are right -- reusing
low_stock_threshold instead of inventing one, and refusing to key off remain.
_remaining_percent is character-for-character the Inventory page's expression and
pct < threshold matches its comparison, so the badge and the alert genuinely
cannot disagree about the arithmetic. And test_the_event_reaches_a_provider is
the assertion whose absence caused this. Three things need fixing before it goes
in.

The slot label. _ams_slot_label gets three real slot kinds wrong, measured
against the helper that already exists for this on the notification side
(_slot_label_from_global_tray composed with _global_tray_from_assignment in
spool_assignment_notifications.py):

  ams=254 tray=0   yours='HT-¿'      canonical='Ext-L'
  ams=255 tray=0   yours='External'  canonical='Ext-L'
  ams=255 tray=1   yours='External'  canonical='Ext-R'
  ams=  6 tray=0   yours='AMS-G T1'  canonical='Lite-1'

External assignments store ams_id=255 with tray_id picking left or right
(inventory.py:1771, ext_id = data.tray_id + 254), so on your own X2D both
externals get the same label. ams_id=254 rows exist in the wild --
InventoryPage.tsx:286, usage_tracker.py:55 and _global_tray_from_assignment all
handle them -- and fall into your >= 128 branch as HT- plus chr(191). A2L
AMS-Lite normalises to unit 6. Please call the existing pair rather than adding a
fifth spelling; it fixes all four lines at once.

The 3 s pass. run() sleeps _fast_check_interval, 3 seconds, on any productive
pass, and the early-return path reports productive while an upload is in flight
(#2602). So in Spoolman mode get_all_spools -- the whole collection, and there is
no cache anywhere in SpoolmanClient -- fires every 3 seconds for the length of a
batch drain. That is the request storm the design note is arguing against, just
moved. And when Spoolman is down _get_with_retry burns three attempts at a 5 s
connect timeout with 0.5 s backoff, ~16 s, inside check_queue holding the
scheduler's session, so a third-party outage throttles the print queue. Gate the
Spoolman resolution on elapsed time, or run it only on the idle pass.

The debounce needs the spool id. Pull a low spool, put a different part-used one
in the same slot, and it never alerts -- I ran it:

  low spool alerts                                 -> 1
  assignment deleted, second 10% spool assigned    -> still 1, expected 2

The key survives the pass where the slot resolves to nothing, and the new spool
never goes above the threshold to clear it. (printer_id, ams_id, tray_id,
spool_id) closes it with no pruning needed.

Three smaller ones. Archived spools alert -- I ran that too, and the Inventory
count skips them (InventoryPage.tsx:1089), while get_all_spools without
allow_archived already excludes them in Spoolman mode, so internal is the odd one
out; one where(Spool.archived_at.is_(None)). SpoolmanUnavailableError lands in
the broad except with exc_info=True, so an unreachable Spoolman writes a full
traceback every pass, every 3 s while dispatching -- catch it in
_filament_low_slots_spoolman and return {}, which your no-slots branch already
handles correctly. And _filament_low_slots_spoolman has no tests at all: fourteen
tests and none touches the branch with the client, the mapping and the
missing-spool skip, nor the "a pass that resolves nothing leaves the set alone"
behaviour you call out as non-obvious. Given the threshold is meant to be
meaningful in both modes, that half needs its own coverage.

Two notes rather than requests. Marking before the send means one transient
provider failure loses the alert for good -- defensible, but it is the opposite
failure mode from the one the comment argues against, so say it is a choice. And
mqtt_relay.on_filament_low still has no caller; it is one of five dead relay
methods so it is not uniquely yours, but your own table in #2913 listed it, so
either wire it or say in the PR that it stays dead, so the issue does not close
over it.

Declining the enumeration guard because it would land red on the stock alerts is
the right call, and I would rather have that said plainly than have it land red.
Suite on your branch merged with dev: 11056 passed, ruff check and ruff format
--check clean.

ojimpo:
Thanks — all three blockers and the three smaller ones are done, and both notes are answered.

**The slot label now defers to the pair you named.**

`_ams_slot_label` is `_slot_label_from_global_tray(_global_tray_from_assignment(ams_id, tray_id))`.
That fixes all four of your lines at once, and it changes normal AMS slots to `A1` / `B4` too,
which is the point — a fifth spelling was the actual defect, not the four wrong rows. The test
covers every kind:

```
_ams_slot_label(0, 0)   == "A1"
_ams_slot_label(1, 3)   == "B4"
_ams_slot_label(128, 0) == "HT-A"
_ams_slot_label(254, 0) == "Ext-L"
_ams_slot_label(255, 0) == "Ext-L"
_ams_slot_label(255, 1) == "Ext-R"
_ams_slot_label(6, 0)   == "Lite-1"
```

**The 3 s pass is gated.**

`_check_filament_low` returns immediately unless 30 s have passed — the idle interval, since a low
spool does not need answering at the fast cadence. I gated the whole check rather than just the
Spoolman resolution: the internal path was also re-querying every assigned spool every 3 s, and
the ~16 s of `_get_with_retry` inside `check_queue` you describe goes away with it.

**The debounce key carries the spool id.**

`(printer_id, ams_id, tray_id, spool_id)`. I reproduced your scenario and it now alerts twice —
`test_a_replacement_spool_in_the_same_slot_can_alert` walks it through: low spool alerts, the slot
resolves to nothing for a pass while the roll is swapped, the replacement at 10% alerts. No
pruning needed, as you said.

**The three smaller ones.**

Archived spools are excluded with one `where(Spool.archived_at.is_(None))`.
`SpoolmanUnavailableError` is caught in `_filament_low_slots_spoolman` and returns `{}`.
And `_filament_low_slots_spoolman` has tests now — the client and the mapping, a spool deleted out
from under an assignment, an outage, and the "a pass that resolves nothing leaves the set alone"
behaviour, asserted by showing an outage does not re-alert everything on recovery.

One thing worth flagging on the outage test. Asserting "no notification" would have been vacuous:
without the catch, `SpoolmanUnavailableError` reaches the broad `except` and the result is *also*
no notification. The observable difference is the traceback, so the test asserts
`"Low-filament check failed" not in caplog.text`. Removing the catch fails it.

Same class of problem with the gate: the existing multi-pass tests would have been swallowed by it
and passed without exercising the debounce at all. They now go through a `_pass()` helper that
steps over the gate deliberately, and the gate has its own test asserting no work is done rather
than no notification sent.

Every change measured against a mutation:

```
head                                          21 passed
old _ams_slot_label restored                   4 failed, 17 passed
time gate removed                              1 failed, 20 passed
archived filter removed                        1 failed, 20 passed
SpoolmanUnavailableError catch removed         1 failed, 20 passed
spool_id dropped from the debounce key         1 failed, 20 passed
```

**Your two notes.**

Marking before the send is a choice and the comment says so now, with the trade named: one
transient provider failure loses the alert until the spool crosses the threshold again, versus
re-alerting every pass while a provider is down. I kept the current behaviour because a spool that
is low stays low, so another real signal is not far away, whereas a provider stuck retrying is a
signal that does not stop — but it is a judgement, not a forced move.

`mqtt_relay.on_filament_low` **stays dead, and I would rather say so than wire it here.** Its
contract is spool-shaped — `(spool_id, spool_name, remaining_weight, remaining_percent)` — and this
producer is slot-shaped. There is no convention in the backend for building a spool's display name
(`spool_csv._spool_key` is a dedup key, not a label), so wiring it would mean inventing one in a
bug-fix PR. It is one of the five dead relay methods and I would rather it were done deliberately,
across all of them, than have this PR set the precedent for the name. Flagging it here so #2913
does not close over it.

Full backend unit suite on this branch with the `dev` merge under it: **8631 passed, 22 skipped**.

maziggy:

PR #2944, opened 2026-08-24
Match a Bambu roll to its own product line, not just its colour (issue #2907)

### LINKED ISSUE #2907: [Bug]: Spoolman filament auto-match ignores the AMS sub-brand — PLA Matte spools are linked to the PLA Basic filament of the same colour

ojimpo:
## Description

Material plus colour is not an identity. PLA Basic Black and PLA Matte Charcoal are both PLA at `#000000`, so a Matte roll linked to the Basic filament and took its name. The product line the printer reports as `tray_sub_brands` is the missing axis, and — as you said — a Spoolman filament has no field for it, so `name` is the only place it can live. That is why resolving the colour name is the core of this rather than the cheap alternative to it.

Your simplification holds: the catalogue's `material` column already carries `"PLA Matte"`, the same string `tray_sub_brands` does, so `(Bambu Lab, hex, sub-brand)` resolves the name directly with no rebuilding from a type plus a subtype.

**Your reproduction correction changed the patch, as you said it would.** The external tie-break is real and fixed here, but the internal search short-circuits before it — matching on material and colour with no line at all — so a tie-break-only patch would not have fixed the case I filed. Both loops go through the new criterion.

## One addition to the shape you asked for, and it is load-bearing

Taking the colour name as the third criterion **on its own** mints a duplicate filament for every spool on an existing instance. Bambuddy names its own creations after the sub-brand (`name=tray.tray_sub_brands or tray.tray_type`, `spoolman.py:1296`), so `"PLA Basic"` never equals `"Black"` and nothing already on disk would match.

So the criterion is *"does this name identify the same product line"*, satisfied by **either**:

- the catalogue's colour name — what this path creates from now on, and what an entry taken from the external library is already called;
- the sub-brand itself — what Bambuddy has been calling its own creations since before this fix.

Both encode the line, so both separate Basic from Matte:

| spool | before | after |
|---|---|---|
| PLA Matte Charcoal roll, `"Black"` (PLA Basic) filament on disk — **the reported case** | linked to it | new filament, `"Charcoal"` |
| PLA Basic roll, Bambuddy-created `"PLA Basic"` filament | linked | linked — **no duplicate** |
| PLA Basic Black roll, external-derived `"Black"` filament | linked | linked — **no duplicate** |

**The residual case, stated rather than hidden:** an external entry named something like `"PLA Basic Black"` satisfies neither spelling and still duplicates. It is bounded to one per affected roll, and the new record is the correct one from then on. I judged that better than the alternative, which is every spool on every instance duplicating on the next sync — but it is your call and I will take whichever you prefer.

When the catalogue has no row — it is seeded from Bambu's published list and lags new colours — the filament is built from the tray data rather than attached to whichever candidate came first. That part is unchanged from what we agreed.

## `db` is required, not optional

The colour catalogue lives in Bambuddy's database and `SpoolmanClient` had no DB access at all — every other method on it is pure HTTP. A session here is a real cost to that boundary, so it is worth saying why it is not optional: an optional one leaves the defect in whichever caller forgets to pass it, which is the shape #2903 was about. Every existing test failed with `TypeError` until it was passed, which is the property working.

Four call sites updated: `main.py:2585`, and `routes/spoolman.py` at 358, 576 and 1263.

## Extracted rather than copied

You pointed me at `spool_tag_matcher.py` as a reference implementation to copy. I extracted it instead, into `services/color_catalog_lookup.py`.

The reason is the drift you named on #2912: that logic has been inline since #857 and three fixes have landed there without crossing to this side — the sub-brand filter (#1227), the alpha guard (#1545), the translucent case. This would have been the fourth copy of it. One implementation is the only thing that stops a fifth.

**It is not adopted on the built-in side in this PR** — #2923 is open against that same file and I would rather not hand you a conflict. Follow-up once that lands, if you want it.

## A small correction

Your comment says 40 PLA Matte rows are seeded. There are 38 across all manufacturers, and **25 of them are Bambu Lab** — which is the number that matters here, since the lookup filters on manufacturer. Does not change anything about the design; noting it because I checked.

## Related Issue

Fixes #2907

## Documentation

**Pick one**:
- [x] No docs update required — reason: bug fix with no new setting, endpoint or UI element. The wiki describes Spoolman filament matching as automatic and does not document the match key, so there is nothing that becomes wrong.

## Type of Change

## Changes Made

- `backend/app/services/color_catalog_lookup.py` — new. `resolve_bambu_color_name(db, rgba, sub_brand)`, carrying the #1545 alpha guard over from the built-in path, and `filament_matches_product_line(name, expected_color_name, sub_brand)`
- `backend/tests/unit/services/test_color_catalog_lookup.py` — new. 13 direct tests against a real session
- `backend/app/services/spoolman.py` — `_find_or_create_filament(tray, db)`: resolves the expected colour name, uses it as the third criterion in the internal match, as the external candidate selector, and as the name on creation; `sync_ams_tray` takes and forwards `db`
- `backend/app/main.py`, `backend/app/api/routes/spoolman.py` — the four call sites
- `backend/tests/unit/services/test_spoolman_service.py` — four new tests, two rewritten, a `tray_matte` fixture promoted out of a local, and a `_Catalog` / `_NoCatalog` stub pair so a test can say which state it is exercising
- `backend/tests/unit/services/test_spoolman_extra_field_registration_2903.py`, `backend/tests/integration/test_spoolman_api.py` — signature updates

## Screenshots

## Round two

**1. "The equality never held" was wrong, and you are right that it sat in four places.** You replayed it correctly: my fixture contained an entry named exactly `"PLA Basic"`, the equality matched it, and `candidates[0]` was `"Black"` listed first. The old test passed because the tie-break worked. All four occurrences are corrected or deleted.

What is true is narrower, and here is the query you asked for. Against a current SpoolmanDB via `GET /api/v1/external/filament` on a live Spoolman 0.26.1 — 6957 entries, **269 Bambu Lab**:

```
material values (Bambu Lab):
  PLA 131, PETG 37, ABS 17, TPU 13, ABS-GF 8, PA6-GF 8, ASA 7, PC 7,
  PLA-CF 7, PETG-CF 6, PLA+WOOD 6, TPU-95A 6, ...

entries whose name is a sub-brand string:
  name == 'PLA Basic'   -> 0        name == 'PLA Silk'  -> 0
  name == 'PLA Matte'   -> 0        name == 'PLA Tough' -> 0
  name == 'PETG HF'     -> 0        name == 'ABS'       -> 0

names carrying 'PLA' at all:  3 / 269   (all "Support for PLA/PETG ...")
```

Every Bambu Lab entry is colour-named, with the line folded into the id. The rows at `#000000` under `material: PLA`:

```
name='Black'                       id='bambulab_pla_black_1000_175_n'
name='Matte Charcoal'              id='bambulab_pla_mattecharcoal_1000_175_n'
name='Tough+ Black'                id='bambulab_pla_tough+black_1000_175_n'
name='Support for PLA/PETG Black'  id='bambulab_pla_supportforpla/petgblack_500_175_n'
```

So the accurate statement is: **against the real library the equality cannot match, because no Bambu entry is named for a sub-brand — and `material` is `"PLA"`, never `"PLA Basic"`.** It matched only in my fixture, which staged a row SpoolmanDB does not contain. That is the same defect you caught in the #2906 fixture, so the fixture is the thing I changed: the test now uses the real pair above, and there the *old* tie-break genuinely fails —

```
E   AssertionError: assert 'Black' == 'Matte Charcoal'
```

a PLA Matte roll created as PLA Basic Black, which is the report.

**On the direction reversing.** You are right that where both spellings exist the new selector prefers the colour-named entry, and that this is the opposite of #1309. On the evidence above that case does not occur in the external library — there is nothing line-named to lose to. It can occur among *internal* filaments, and there `filament_matches_product_line` accepts both spellings rather than ranking them, so an instance that already has `"PLA Matte"` from Bambuddy's own creations keeps matching it. What #1309 asked for is preserved where it applies.

The instance also shows why this matters for existing installs. Its own Bambu filament reads:

```
id=6  name='Black'  material='PLA Basic'  vendor='Bambu Lab'
```

Colour in `name`, line in `material` — the external-derived shape, which the old `name == tray_sub_brands` equality would never have matched.

**2. The #1545 alpha guard is restored, and I am saying so out loud.** You are right that dropping it made the two modes name the same roll differently — "Clear" from the built-in path, "Black" from this one — which is precisely what the module exists to stop. Naming a genuinely translucent roll from the catalogue probably *would* beat "Clear", but that changes internal mode's behaviour since #1545 and belongs in its own issue, not smuggled through an extraction. The docstring now states the carry-over and the reasoning. I have not filed the translucent idea — say the word and I will.

It is pinned twice: `test_an_alpha_00_roll_is_clear_and_never_reaches_the_catalogue` in the new module tests, seeding the exact trap row, and `test_a_clear_tray_is_not_named_from_the_catalogue_even_when_it_can_answer` at the `_find_or_create_filament` level with a catalogue that *can* answer — the gap you identified, since the existing clear-tray test passes `_NoCatalog()`.

**3. The module has direct tests now** — `backend/tests/unit/services/test_color_catalog_lookup.py`, 13 against a real session with seeded rows, covering the manufacturer filter, the sub-brand filter, the id tie-break, the six-character prefix, case-insensitivity on both sides, the alpha guard, and both spellings `filament_matches_product_line` accepts.

Worth reporting that my first version of the manufacturer test was vacuous: it seeded the other manufacturer's row with `material: "PLA"`, so the sub-brand filter dropped it before the manufacturer filter was ever consulted, and the test passed with the filter deleted. It now leaves the sub-brand empty — the third-party case, where manufacturer is the only thing separating the rows — and fails when the filter goes. The docstring says why the sub-brand is omitted, so it does not get "helpfully" added back.

**On the two notes.**

The lossy-proxy point is now in the `filament_matches_product_line` docstring, naming what it cannot separate: `#000000` is "Black" under PLA Basic, PLA Tough, PLA-CF, ABS, ASA, PETG HF, PETG-CF and TPU 90A, so two PLA lines whose catalogue rows carry the same colour name still collapse. Pre-existing, not made worse, but "both encode the product line" reads stronger than it is.

And you are right that the residual duplicate case is wider than I wrote — it is any roll whose colour the catalogue has no row for, not only an entry named "PLA Basic Black". **It stays bounded to one per newly registered roll**, because `_find_or_create_filament` only runs when the tag is not already in Spoolman: an instance's existing spools are untouched, and a roll already registered never comes back through here.

### Verification of each of the three

```
alpha guard removed                        2 failed, 66 passed
manufacturer filter removed                1 failed, 68 passed
sub-brand filter removed                   2 failed, 66 passed
old tie-break restored                     1 failed, 68 passed   (assert 'Black' == 'Matte Charcoal')
```

## Testing

- [x] I have tested with my printer model: **X2D** — the three mislinked spools in the report are its AMS reads

**Two existing tests changed, both deliberately.**

`test_returns_existing_internal_bambu_lab_filament` is the one you warned me about. It pinned the short-circuit using a filament named `"Black"` standing in for a PLA Basic roll — which is the defect. Its own docstring already described the fallout ("once a wrong name is on disk, subsequent AMS reads keep reusing it") as a known cost of the #1309 workaround. It is now three tests: the same-line reuse still short-circuits, a filament named for the catalogue colour also matches, and a different-line filament does not.

`test_prefers_external_entry_matching_tray_sub_brands` is rewritten — see **Round two** above for why my description of it was wrong.

```
the touched test files, on this branch          134 passed
the new module's own tests                       13 passed
ruff check                                      All checks passed!
ruff format --check                             831 files already formatted
```

I also ran the wider suite, but on a rebase onto current `dev` rather than on this branch — **8269 passed, 22 skipped** — because I had rebased before noticing you had merged `dev` in here yourself. The four files this PR touches are byte-identical between the two trees, so the numbers describe the same change; I am reporting both rather than presenting the wider run as something it was not. Say the word if you would rather this branch were rebased than carrying your merge, and I will redo it that way.

## Additional Notes

**The #2924 conflict is resolved.** You merged it on 08-25 and said this one should rebase afterwards; it is rebased onto current `dev` and re-run. The two were independent in behaviour as expected — #2924 decides *which colours are the same colour*, this one decides *which filament a colour belongs to* — and the merge needed no reconciliation between them.


maziggy:
Plese rebase to resolve confict.

ojimpo:
Rebased onto dev.

The conflict was #2924 landing in between. It replaced the raw `.upper()` colour comparison in `_find_or_create_filament` with `color_match_key`, and this branch was written against the old expression, so both loops collided.

Resolved by keeping dev's key and putting the product line behind it:

```python
color = color_match_key(tray.tray_color)
sub_brand = (tray.tray_sub_brands or "").strip()
expected_color_name = await resolve_bambu_color_name(db, tray.tray_color, sub_brand)
...
    if fil_material.upper() != material_upper or color_match_key(filament.get("color_hex")) != color:
        continue
    if filament_matches_product_line(filament.get("name"), expected_color_name, sub_brand):
        return filament
```

The catalogue lookup reads `tray.tray_color` rather than the match key on purpose: it keys on the six-character prefix, so a clear roll — eight characters in the key since #2924 — still has to resolve its name like any other. There is a comment on that line saying so.

The five `TestColorHexAlphaHandling` tests #2924 added call `_find_or_create_filament` directly and now need the session. I did not want to hand them all a catalogue that answers nothing, because that would let the product-line check reject a filament the colour key was supposed to reject, and the test would pass for the wrong reason. So each gets the catalogue that leaves its own assertion as the only thing that can fail:

| test | session | why |
|---|---|---|
| `test_clear_tray_creates_a_translucent_filament` | `_NoCatalog()` | nothing to match; the catalogue has no bearing |
| `test_opaque_tray_still_matches_an_existing_six_char_filament` | `_Catalog("Black")` | the existing filament is named `Black`, so the line criterion is satisfied and the six/eight-character key is the only thing under test |
| `test_clear_tray_does_not_attach_to_the_black_filament` | `_Catalog("Black")` | same — the black filament passes the line check, so only the key excludes it |
| `test_black_tray_does_not_attach_to_a_clear_filament` | `_Catalog("Clear")` | the inverse, same reasoning |
| `test_clear_tray_does_not_take_a_same_rgb_external_entry` | `_Catalog("PLA Basic Black")` | the catalogue names the colour exactly what the external entry is called, so the selector would take it; only the colour key keeps it out of the candidate list |

One line of the commit message changed with the rebase: it said #2923 was open against `spool_tag_matcher.py`, which is why the resolution is extracted rather than adopted on the built-in side. #2923 has merged since; crossing the two paths is still a follow-up rather than an omission.

Suite on this branch merged with dev:

```
backend/tests/unit         8062 passed, 22 skipped
backend/tests/integration  2 failed, 2989 passed, 1 error
```

`ruff check` and `ruff format --check` are clean.

The three integration failures are not from this branch, and I do not think they are from any branch. They are in `test_ownership_permissions.py` and `test_github_restore_api.py`, nowhere near this change. I ran those two files three times to check:

| run | tree | result |
|---|---|---|
| 1 | bare `dev` @ 935d4b5b | 2 failed — `test_operator_cannot_delete_folder_with_files`, `test_operator_cannot_write_ownerless_archive_subresource[source_upload]` |
| 2 | this branch | 122 passed |
| 3 | bare `dev` @ 935d4b5b | 2 errors — `test_operator_cannot_update_others_library_file`, `test_every_missing_permission_is_named_at_once` |

Same code, a different set each time. Every one of them is the same failure: the `auth_setup` fixture gets a login response with no `access_token` in it, so `KeyError: 'access_token'` at setup. I have not worked out why — the per-IP login rate limit in `auth.py:475` is the obvious suspect, but `test_engine` is function-scoped and drops its tables, so the event table should be empty for each test, and I did not want to hand you a guess. Happy to open it separately if you have not already seen it.

maziggy (changes requested):
Thanks @ojimpo - the design is right and the reported case is properly fixed. I verified the
premise: the catalogue's material column carries "PLA Basic" and "PLA Matte", the same strings
tray_sub_brands reports, and there are zero rows with material exactly "PLA". Your count
correction is right too - 38 PLA Matte rows, 25 of them Bambu Lab. Full suite passes here
(11159) and ruff is clean.

Three things before merge.

1. "The equality never held" is not true, and it is repeated in four places - the PR body, the
   comment in spoolman.py, and the rewritten test's docstring and inline comment. I replayed the
   old selector against your own fixture: tray_pla_black.tray_sub_brands is "PLA Basic", the
   fixture contains an entry named exactly "PLA Basic", so the equality matched it and returned
   density 1.26 - which is what the old test asserted. candidates[0] was "Black", listed first,
   not second. The old test passed because the tie-break worked.

   That may not change the fix: if real SpoolmanDB entries are colour-named, the old tie-break is
   dead in the field and yours is better. But that is now the only thing holding the change up and
   nothing in the PR evidences it - please show a real external-library query. Note the direction
   reverses where both spellings exist: the new selector prefers the generic colour-named entry
   over the line-named one, which is the opposite of what #1309 asked for. And the comments need
   correcting either way, since they currently misdescribe the code they sit next to.

2. resolve_bambu_color_name drops the #1545 alpha guard, which the module's own docstring names as
   one of the three fixes it exists to stop re-losing. Measured against seeded rows:
   ("00000000", "PLA Basic") resolves to "Black", while create_spool_from_tray returns "Clear" for
   the same roll - so a clear roll can create a Spoolman filament called "Black", and the two modes
   now name the same roll differently. For a real PLA Translucent roll the lookup is better than
   "Clear", so this may be an improvement you meant - but decide it out loud, say so in the
   docstring, and pin it. No test currently lets the catalogue answer for an alpha-00 roll:
   test_clear_tray_creates_a_translucent_filament passes _NoCatalog().

3. The new module has no direct tests. _Catalog.execute ignores the query, so the manufacturer
   filter, the sub-brand filter, the id tie-break and the six-character prefix are all unpinned - I
   had to seed rows by hand to find the alpha behaviour above. A handful of tests against a real
   session would cover it.

Two notes, neither blocking:

- The criterion is the colour name, which is a lossy proxy for the product line. #000000 is "Black"
  under PLA Basic, PLA Tough, PLA-CF, ABS, ASA, PETG HF, PETG-CF and TPU 90A, so lines that share a
  base material and a colour name still collapse. Pre-existing and not made worse here, but the
  docstring says "both encode the product line" and someone will later read that as stronger than
  it is. One sentence on what it cannot separate.

- The residual duplicate case is wider than the PR states. It is not only an entry named
  "PLA Basic Black": whenever the catalogue has no row - the lagging case you cite yourself - an
  external-derived filament named "Black" no longer matches and a "PLA Basic" one is minted beside
  it. Still bounded to one per newly registered roll, since _find_or_create_filament only runs when
  the tag is not already in Spoolman, which is worth saying in the PR because it makes the tradeoff
  look much better than it currently reads.

The extraction itself is the right call and db being required rather than optional is correct - the
whole suite failed loudly until every caller passed it, which is the property working. No CHANGELOG
needed on your side.

maziggy (changes requested):
Thanks @ojimpo - all three points are addressed and I checked each rather than taking the body
for it. The alpha guard is back with the reasoning stated and pinned on both levels, the module
has real-session tests, and the equality claim is corrected everywhere - no trace of it left in
the code or the test docstrings. Your mutation numbers reproduce here exactly:

  alpha guard removed          2 failed, 132 passed
  manufacturer filter removed  1 failed, 133 passed
  sub-brand filter removed     2 failed, 132 passed
  old tie-break restored       1 failed, 133 passed   (assert 'Black' == 'Matte Charcoal')

And the wider suite on this branch as it stands, with the dev merge under it: 11535 passed,
ruff clean. No need to caveat that one any further.

Your external-library query settles point 1. Thank you for going and getting it - it is the
only thing that could have, and it changes the tie-break from an improvement into a correction.

One thing before merge.

1. The decision not to fall back to an arbitrary external entry is untested, and it is the
   biggest behaviour change in here outside the reported bug. Both halves of it survive:

     catalogue has no row -> restore the old candidates[0] fallback   134 passed
     catalogue answers, no external entry carries that name           134 passed

   test_falls_back_to_create_when_no_bambu_match_anywhere looks like it covers the first, but
   its external entry is 3DJAKE - the manufacturer filter drops it before bambu_candidates is
   built, so it arrives at that branch with an empty candidate list and never exercises the
   decision. The case that matters is a real Bambu candidate present and deliberately not used.

   This one fails under the mutation and passes on head:

     async def test_a_colour_the_catalogue_lags_is_built_from_the_tray_not_a_bambu_entry(
         self, client, tray_pla_black
     ):
         external = [{"id": "bambulab_pla_black_1000_175_n", "manufacturer": "Bambu Lab",
                      "name": "Black", "material": "PLA", "color_hex": "000000",
                      "density": 1.31}]
         ... _find_or_create_filament(tray_pla_black, _NoCatalog())
         assert mock_create.call_args.kwargs["name"] == "PLA Basic"
         assert "density" not in mock_create.call_args.kwargs

   Worth a sibling for the second half too - catalogue answers "Matte Charcoal", the library
   offers only "Black", and the roll is still built from the tray.

Two notes, neither blocking.

2. test_no_sub_brand_falls_back_to_the_lowest_id does not pin the tie-break. Deleting the
   order_by leaves 134 passing; only reversing it to .desc() fails. On SQLite the id is the
   rowid, so an unordered scan comes back in id order anyway - I tried assigning ids against
   insertion order to separate them and it still passed. The suite is SQLite-only and there is
   no Postgres job, so nothing in CI can notice if that ORDER BY goes, and the docstring's
   promise that the two modes cannot disagree about which row wins is exactly a determinism
   claim. I do not think a better test exists here; saying so in the docstring is the honest
   version.

3. Small correction to that test's reasoning. The third-party roll it describes cannot reach
   the module from its only caller: is_bambu_lab_spool gates non-Bambu rolls out, and
   parse_ams_tray substitutes tray_type when tray_sub_brands is empty (spoolman.py:1116-1118),
   so sub_brand is never empty on this path. The no-sub-brand branch becomes live when the
   built-in side adopts the module, which is a good reason to keep it - just not that one.

The extraction reads well and the docstrings are doing real work now. No CHANGELOG needed on
your side.

ojimpo:
Thanks — all three are done, and you were right that the fall-through was the untested part.

**1. Both halves of the decision now have a test.**

I took your test and added the sibling. `test_a_colour_the_catalogue_lags_is_built_from_the_tray_not_a_bambu_entry`
is the first half: a real Bambu Lab candidate present, catalogue with no row, roll built from the
tray anyway. `test_a_name_the_library_does_not_carry_is_built_from_the_tray_too` is the second:
catalogue answers "Matte Charcoal", the library offers only "Black", and the roll is still built
from the tray — which is the entry the old tie-break would have taken.

Both assert `"density" not in kwargs` as well as the name. `density` reaches `create_filament`
only through `_create_filament_from_external`, so its absence is what actually separates the two
paths; the name alone would not, since the candidate is called "Black" for reasons of its own.

Measured on the merged tree:

```
head                                             71 passed
old candidates[0] fallback restored              2 failed, 69 passed
```

Only the two new tests fail under that mutation — `test_falls_back_to_create_when_no_bambu_match_anywhere`
still passes, which is exactly the gap you described.

**2. The tie-break docstring says it is untested, because it is.**

I reproduced what you found: deleting the `order_by` leaves it passing and only `.desc()` fails,
for the rowid reason you gave. I could not build a better test either. The docstring no longer
claims the two modes cannot disagree about which row wins — it states the ordering is not pinned
here, why no useful test exists, and that nothing in CI would notice the `ORDER BY` going.

**3. Corrected — the third-party roll cannot reach the module.**

I checked rather than taking it: `parse_ams_tray` substitutes `tray_type` when `tray_sub_brands`
is empty at `spoolman.py:1116-1118`, and `is_bambu_lab_spool` gates non-Bambu rolls out before
that. So `sub_brand` is never empty on this path and the reason I gave was wrong. The docstring
now says the branch goes live when the built-in side adopts the module, and keeps it on that
basis.

Full backend unit suite on this branch with the `dev` merge under it: **8628 passed, 22 skipped**.

maziggy (changes requested):
Round three reviewed. Two things to change, and the first one matters more than what I raised last round.

1. The colour name is going into the identity field, and that breaks parity with internal mode.

I ran the same PLA Matte Charcoal roll through both sides:

                                material  subtype          color_name
internal mode                   PLA       Matte            Charcoal
spoolman BEFORE (from scratch)  PLA       Matte            Matte      (synthesized)
spoolman AFTER  (from scratch)  PLA       Charcoal         Charcoal   (synthesized)
spoolman external path          PLA       Matte Charcoal   Matte Charcoal

Before this PR Spoolman mode had the subtype right and the colour wrong. After it, the two are swapped. Neither state matches internal mode on both fields, so the patch moves which one is broken rather than fixing it.

The cause is that #1357 already decided where a colour name lives on this side: Spoolman has no color_name on Filament, so it goes in spool.extra.bambu_color_name and the filament's name carries identity. _map_spoolman_spool reads that key first and only synthesizes from the subtype when it is absent. The AMS-sync path never writes it - create_spool(extra={"tag": json.dumps(spool_tag)}) is the only extra it sets - so the colour ends up in filament.name because the field built for it is sitting unused two lines away.

The fix does not weaken #2907. Keep creating the filament as tray_sub_brands, which filament_matches_product_line already accepts through its sub-brand spelling, and add bambu_color_name: json.dumps(expected_color_name) to that same create_spool extra dict. _ensure_extra_fields has been payload-keyed since #2903, so it registers itself. Both fields then land where internal mode has them, and the line stays in the filament name where your match key needs it.

2. The catalogue does not say what two of the tests say it says.

catalog_defaults.py seeds ('Bambu Lab', 'Charcoal', '#000000', 'PLA Matte'). test_prefers_the_external_entry_the_catalogue_names stages _Catalog("Matte Charcoal") and its comment states "The catalogue calls #000000 'Matte Charcoal' for PLA Matte". It does not, and here the string decides which branch runs. I seeded the real DEFAULT_COLOR_CATALOG into a session and drove _find_or_create_filament with the external trio from your PR body:

PLA Basic roll -> name='Black'     density=1.24  [external library]
PLA Matte roll -> name='Charcoal'  density=None  [created from scratch]
PLA Tough roll -> name='Black'     density=1.24  [external library: the PLA Basic entry]

So the reported roll never reaches the external branch and loses the library's density and weight; the only test covering that branch for a multi-line colour is staged on a value the catalogue cannot produce; and a PLA Tough roll selects the PLA Basic entry while bambulab_pla_tough+black_1000_175_n is in the same candidate list.

The cause is structural rather than a typo. The catalogue splits the line into material and keeps a bare colour in color_name; SpoolmanDB folds the line into name. Comparing one against the other can only fire for the line whose external names carry no prefix, which is PLA Basic. You have the live 0.26.1 instance, so the query that settles it is whether the external name equals the catalogue color_name for Bambu entries generally or only for Basic. If only for Basic, the selector wants a different key - the external id already carries the line (bambulab_pla_mattecharcoal_1000_175_n). Your call on the shape; what I am asking for is that the test stop staging a catalogue value the seed contradicts.

3. The extraction cannot yet absorb the caller it was extracted for.

The internal block takes three things off that catalogue row - color_name, extra_colors and effect_type - the last two added so an AMS-identified gradient renders like a hand-picked one. resolve_bambu_color_name returns the name alone, so it cannot replace the internal caller as written. The #2923 conflict is not what blocks that adoption, the signature is. And since the new Spoolman path reads the same row and drops the stops, an AMS-identified gradient roll still renders flat in Spoolman mode and gradient in internal mode. Pre-existing, but this is the PR that walks past it with the row already in hand.

One note, no action needed unless you want the docstring to say it. The alpha guard returns the constant "Clear" for every line, so the new criterion is inert for translucent rolls - I measured a PLA Matte clear roll reusing a PLA Basic "Clear" filament. Identical to the old behaviour, so nothing regressed, but the docstring covers translucency only from the naming side and this is the other half.

Sizing, which argues for the change. Of 159 Bambu hexes only 8 are shared across product lines, and most of those cross base materials, which the material check already separates. The same-material ambiguities are #000000, #FFFFFF and #8E9089. This fixes Basic against Matte, Silk and Sparkle; Silk against Tough at white and Basic against Tough at black still collapse, exactly as your docstring says.

On the direction of travel, this is Spoolman mode catching up rather than diverging, and that is worth saying plainly. Internal mode has treated the product line as part of identity since it was written - create_spool_from_tray stores it as subtype and find_matching_untagged_spool filters on material, rgba and subtype with the NULL-subtype row ranked second. Adding the line to the Spoolman match key is the same criterion. The corrections above are about where the resulting values land, not about whether the criterion belongs there.

Everything else checks out. All four call sites updated with none missed, the session in on_ams_change is a live async_session, your row-count correction is right at 25 Bambu PLA Matte of 180 Bambu rows, and the internal-path tests use the correct "Charcoal". The two modules do not import circularly in either direction, so the deferred import inside _find_or_create_filament can move to the top if you want it there. ruff clean; the touched files are 136 passed.

PR #2951, opened 2026-08-24
Resolve dispatch plate_id from the archive instead of assuming 1

### LINKED ISSUE #2947: [Bug]: queue item without plate_id defaults to plate 1, wedges printer if archive has a different plate
**What happens**

A queue item with no `plate_id` always dispatches plate 1
(`print_scheduler.py:6011`, `bambu_mqtt.py:5409`/`:5635`:
`plate_id=item.plate_id or 1`). If the archive's G-code isn't plate 1, the
printer accepts the print command, can't find the file, throws an HMS error,
and sits in IDLE until power-cycled. The queue eventually shows "Check if SD
card is inserted...", the card is fine.

**Why**

A single-plate export keeps its original plate number in
`Metadata/plate_N.gcode`. We export one plate per job, roughly half land on
plate 2. Those never match the `or 1` default. Verified across 28 archives, no
exceptions.

A sibling model's docstring already documents `plate_id: null` as
"single-plate file" for this same field (`schemas/print_queue.py:361`,
`PrintBatchPlateTarget`), but dispatch (`print_scheduler.py:5571`, `:5899`,
`:5998`, `:6011`) still assumes plate 1 instead of checking the archive.

Distinct from #2603 (a selected plate lost on cancellation), here nothing was
ever selected.

**Fix**

When `plate_id` is null, resolve it from which `Metadata/plate_N.gcode` the
archive contains, instead of assuming 1. `default_plate_gcode_name()`
(`threemf_tools.py`) already does exactly this for `GET /{archive_id}/gcode`,
just not on the dispatch path. Happy to send a PR.

**Evidence**

Same job, same printer, retried by hand 13 min later with `plate_id=2`,
prints fine:

```
12:39:57 param: Metadata/plate_1.gcode
12:40:03 HMS error
12:43:00 wedged post-acceptance, reverted to pending
12:44:30 FTP upload failed: Connection reset by peer
...
12:53:09 param: Metadata/plate_2.gcode, prints normally
```

Archive's `Metadata/` has `plate_2.gcode` only, plus leftover
`plate_1.json`/`.png` from the source project.

**Impact**

Roughly half our prints, until we started sending `plate_id` ourselves. Not
urgent for us now, but the default fails silently instead of erroring, and the
error message blames the wrong thing.

**Version**

`ghcr.io/maziggy/bambuddy@sha256:c670164aaa3b0c5af715ca00e9745cdf3a4d7d337fc11d96fc85180371952698`
(latest, built 2026-08-15)

sgiffhorn:
## Description

A queue item with no explicit `plate_id` always dispatched plate 1
(`item.plate_id or 1`, four call sites in `print_scheduler.py`). A
single-plate export cut out of a larger project keeps its original plate
number, so a right-side-only archive whose one G-code member is
`Metadata/plate_2.gcode` was dispatched as plate 1: the printer accepted the
`project_file` command, couldn't find the G-code, raised an HMS error, and
stayed wedged in IDLE until power-cycled (issue #2947).

The four call sites agreed on plate 1 only by accident. With G-code
injection on, `inject_gcode_into_3mf` already falls back to the archive's
own default plate internally whenever the plate id it's handed isn't in the
file, so it could silently inject into a different plate than the one the
print command itself asked for — reproduced and confirmed on #2947.

`default_plate_number()` (new, `threemf_tools.py`) sits beside the existing
`default_plate_gcode_name()` and composes it with the numeric index its
chosen member encodes. `_effective_plate_id()` (new, `print_scheduler.py`,
module-level, mirroring `_nozzle_mismatch_message` right above it) resolves
the plate once in `_start_print`, right after the `file_path.exists()`
guard and before the G-code injection block, and all four call sites that
previously repeated `item.plate_id or 1` independently now reuse that one
value.

## Related Issue

Fixes #2947

## Documentation

**Pick one**:
- [x] No docs update required, reason: internal dispatch fallback, no config
  key, no endpoint, no observable behaviour change for a caller that already
  sets `plate_id`.

## Type of Change

## Changes Made

- `backend/app/utils/threemf_tools.py`: add `default_plate_number(names)`,
  composing `default_plate_gcode_name()` with the numeric plate index.
- `backend/app/services/print_scheduler.py`: add module-level
  `_effective_plate_id(explicit_plate_id, file_path)`, resolve
  `effective_plate_id` once in `_start_print`, replace the four
  `item.plate_id or 1` occurrences with it.
- `backend/tests/unit/test_plate_gcode_selection.py`: add
  `TestDefaultPlateNumber`, covering the single-plate-numbered-2 shape that
  wedged the printer, plus the "no plate number to dispatch" fallback case.
- `backend/tests/unit/test_scheduler_effective_plate_id.py` (new): covers
  `_effective_plate_id`'s file I/O + fallback-to-1 behaviour (unreadable
  file, missing file, no gcode members, no plate_N naming).

## Screenshots

N/A, backend-only change, no UI.

## Testing

- With `_effective_plate_id`'s archive-reading branch temporarily neutered
  to always return 1 (simulating the pre-fix `item.plate_id or 1` path),
  exactly the one test asserting a non-1 result fails; restoring it turns
  it green again.
- `ruff check` and `ruff format --check` clean on the four changed/added
  files.
- Full backend suite (`pytest backend/tests/ --ignore=backend/tests/unit/test_bambu_ftp.py -n 10`):
  10874 passed, 0 failed (14 pre-existing FTP-teardown thread warnings,
  unrelated to this change).
- Also ran the dispatch-focused suites directly (every unit test file that
  references `extract_rack_plan_from_3mf`, `extract_slot_extruders_from_3mf`,
  or `_start_print` — 15 files, 263 tests): all passed.

- [ ] I have tested with my printer model: n/a, no printer path is touched
  beyond the value it dispatches with.

## Checklist

## Additional Notes

Targeted at `dev` per CONTRIBUTING. Shape (public `default_plate_number`
taking `names` rather than reaching for `_plate_number_of`, single
module-level resolution point) follows your review comment on #2947.


maziggy (changes requested):
Reviewed, and the shape is right — thanks for following the outline from #2947 exactly.
Verified the injection claim (threemf_tools.py:1293) and confirmed dispatch is the only
caller of printer_manager.start_print, so those four sites really are the whole surface.
Full backend suite green here, ruff clean. Nice side effect you may not have noticed:
dispatched_plate_id now gets the real plate, so a plate-2-only file stops showing the
plate-1 thumbnail on the printer card.

Three things before I merge.

**1. plate_id=0 now dispatches plate 0.**

`item.plate_id or 1` mapped 0 to 1; `if explicit_plate_id is not None` returns it:

```
item.plate_id=0    -> old=1  new=0
item.plate_id=None -> old=1  new=1
```

That sends `Metadata/plate_0.gcode`, which is the exact wedge this PR removes. Nothing
validates it (`plate_id: int | None = None`, no `ge=1`) and the frontend never sends it,
but the API accepts it, and the rest of the code already reads 0 as "no plate"
(print_queue.py:483, 511, 713, 986, 995). Please make the explicit branch truthiness-based
(or `is not None and > 0` if you want negatives too), with the case added to the test.

**2. The wiring needs a test.**

The helper tests are good, but nothing asserts what the printer is actually told. I
reverted `plate_id=effective_plate_id` at the start_print call back to `item.plate_id or 1`
and the whole suite stayed green. Neutering the helper does fail its own test, so only that
half is covered.

test_scheduler_nozzle_rack_dispatch_1784.py already has the harness — in-memory DB,
synthetic 3MF, _start_print driven end to end. Please add one there: queue a
plate_2.gcode-only archive with plate_id=None, assert printer_manager.start_print was
called with plate_id=2. That is the test that would have caught this bug.

**3. Feed effective_plate_id to register_expected_print too (print_scheduler.py:6314).**

It still passes item.plate_id, and that value is not just a label — it lands in
_print_plate_ids (main.py:968), gets injected into the usage tracker session at
main.py:3628, and ends up in extract_filament_usage_from_3mf. With plate_id None that sums
every filament in the file rather than the printed plate (threemf_tools.py:910 says so
outright), so a multi-plate archive queued without a plate deducts all of them. Pre-existing
and not the shape you reported, but effective_plate_id is exactly the value that fixes it,
it is one argument, and leaving it makes the two paths disagree about which plate ran. It
changes usage numbers, so please give it its own test.

Two nits, take or leave: the BadZipFile fallback is silent, and an explicit plate the
archive does not hold still wedges the printer the same way. A log line in both spots turns
a mystery HMS error into something searchable. Do not redirect the explicit one — that
would print the wrong plate.

sgiffhorn:
Pushed 2dadc6a0.

**1.** `_effective_plate_id` now takes the explicit branch only on `is not None and > 0`, so 0 and negatives resolve from the archive; docstring updated. Counter-check: dropping the `> 0` reddens exactly the two new cases in test_scheduler_effective_plate_id.py.

**2.** Added to test_scheduler_nozzle_rack_dispatch_1784.py. Two things in that harness had to give first: `_write_3mf` wrote no `.gcode` member at all, so the resolution had nothing to answer from and would have fallen back to 1 either way, and the fixture pinned `plate_id=1`. Both are parameters now, defaulted to the old behaviour, so the nine existing cases are untouched. The new case queues a `Metadata/plate_2.gcode`-only archive with `plate_id=None` on an X1C (keeps the rack path out of it) and asserts `start_print.call_args.kwargs["plate_id"] == 2`. Counter-check: reverting that argument to `item.plate_id or 1` reddens it and nothing else.

**3.** `register_expected_print` gets `effective_plate_id`, with its own case asserting `plate_id=2` reaches it. Counter-check: reverting reddens it.

Nits done: the `except` branch logs path and reason, and an explicit plate the archive has no G-code for is logged with the members it does have, then sent as asked.

Scope question. Three more sites in `_start_print` still pass raw `item.plate_id`: `estimate_queue_source_cost` (5806), `archive.plate_id` (5925), `create_archive` (5983); `_get_filament_requirements` (3091) has the same shape. All of them sit above the resolution point, and the cost one runs before `file_path` exists, so pulling them in means reordering `_start_print`, not changing an argument. Left out here. Say the word if you want them in this PR instead of a follow-up.

Confirming your side note: `bambu_mqtt.py:5962` sets `dispatched_plate_id` and `printer_manager.resolve_plate_id` prefers it over parsing the gcode filename, so the plate-2-only file does stop showing the plate-1 thumbnail.

Suite here: 11084 passed, 2 failed, both test_extract_video_last_frame.py from a broken local ffmpeg (missing libx265 dylib), unrelated to this change. ruff clean.

PR #2954, opened 2026-08-25
Give the AI's own halt a name in the archive (issue #2946)

### LINKED ISSUE #2946: [Bug]: A print halted by the onboard AI is archived with no failure reason — 0300_8003 is missing from _HMS_FAILURE_REASONS while its neighbour 0300_8004 is there
### Product

Bambuddy

### Area

Archives / library / files

### Bug Description

When the printer's own AI monitoring halts a print for spaghetti, the archive is written with
`failure_reason=None`. Bambuddy has the code and it has the sentence; what it does not have is the
one row in `_HMS_FAILURE_REASONS` that connects them.

From my own run, with the print marked failed:

```
21:35:20  [ARCHIVE] HMS errors at failure: [{'code': '0x30008', ...}, {'code': '0x8003', ...}]
21:35:20  [ARCHIVE] HMS errors present but none matched a known failure-reason short code
21:35:20  [ARCHIVE] Archive 2 status updated to failed, failure_reason=None
```

`0x8003` with `attr=50364419` resolves to short code `0300_8003`. The text for it is already in the
tree, twice:

```
backend/app/services/hms_errors.py:45
  "0300_8003": "Spaghetti defects were detected by the AI Print Monitoring. Please check the
                quality of the printed model before continuing your print."
frontend/src/components/HMSErrorModal.tsx:85
  same sentence
```

`_HMS_FAILURE_REASONS` (`main.py:504`) has its immediate neighbour and not it:

```python
    # Filament runout (printer-side & per-AMS-slot)
    "0300_8004": "Filament runout",
```

So a runout is archived as "Filament runout" and an AI-detected pile-up is archived as nothing.
The two are one hex digit apart and arrive through the same path.

**The current fallback behaviour is right and I am not asking to change it.** The comment above the
map (`main.py:492-503`) explains that matching on `module` alone used to archive user cancellations
as "Layer shift", and that the code now matches on the full short code so that "anything not in
this map leaves `failure_reason=None` rather than guessing". `derive_failure_reason`'s docstring
says the same — "don't guess — null is honest". That is the correct design. This is a code with a
documented meaning that the map has not been told about.

**I am not asking for the other code in that log.** The same failure carried `0C00_0008`
(`attr=201327360`, `code=0x30008`), which has no entry in `hms_errors.py` either, so nobody knows
what it means. Adding it would be exactly the guessing the comment rules out.

**There is a second AI-spaghetti sentence you may want to take at the same time.**
`hms_errors.py:577` carries `0C00_8042`, "The AI print monitor has detected a spaghetti defect",
and that is absent from `_HMS_FAILURE_REASONS` too. Unlike `0C00_0008` its meaning is documented,
so adding it would not be a guess — but I have not seen my printer emit it, so I am flagging it
rather than reporting it. Your call whether one row or two.

### Expected Behavior

A print halted by the onboard AI is archived with a failure reason, in the same short-category
style as the existing values. Something like `"Spaghetti / AI detection"` — I have deliberately not
picked the wording, since the existing values ("Layer shift", "Filament runout", "Clogged nozzle")
are yours and this would be the first entry in a new category.

### Steps to Reproduce

1. Set the AI failure detection sensitivity to **high** (raising the spaghetti setting pulls
   pileup up with it). This matters for reproducing it at all — at medium the detector did not
   fire on two consecutive failures here.
2. Let a print fail in a way the onboard AI recognises, and let the printer halt it itself
   (`print_halt: true`, offering `RESUME_PRINTING_DEFECTS` / `STOP_PRINTING`).
3. Stop it **from the printer's screen**. Stopping through Bambuddy sets the user-stop flag, and
   `derive_failure_reason` returns "User cancelled" for a cancelled status before it ever consults
   the map — the null only appears on a genuine `failed`.
4. The archive is `failed` with `failure_reason` null, and the log carries the
   "HMS errors present but none matched a known failure-reason short code" line quoted above.

The map itself can be checked without a printer: `0300_8003` is in `hms_errors.py` and in
`HMSErrorModal.tsx`, and not in `_HMS_FAILURE_REASONS`. Steps 1-4 are what produced the archive
row; the grep is what confirms why.

- Confirmed against `dev` today: `0300_8003` still absent from `_HMS_FAILURE_REASONS`, still
  present in `hms_errors.py:45` and `HMSErrorModal.tsx:85`.
- Confirmed the short-code derivation is not the problem — `_hms_short_code(50364419, 0x8003)`
  gives `0300_8003`, which is exactly the key in the other two tables.
- Checked I was not asking for a fallback change by reading the comment above the map and the
  `derive_failure_reason` docstring first. Both say null-on-no-match is deliberate, and I agree.
- Searched for `0300_8003`, "spaghetti", "HMS failure reason" and "AI print monitoring failure
  reason". Nothing covering this.
- Read #2926 rather than stopping at its title — see below.

### Printer Model

X2D

### Bambuddy Version

1.2.6b1 (`daily` channel)

### Printer Firmware Version

01.02.00.00

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian) -- Ubuntu 24.04

No support package — the relevant lines are the three quoted above, taken from the app log at the
moment of archiving, and the defect is a missing dictionary key rather than anything a bundle would
show. I still have the run, including the timelapse of the detection itself, and can attach either
if it would help.

### Additional Context

**This is not #2926.** That issue asks for `HMS_ERROR_DESCRIPTIONS` — the full sentence catalogue in
`hms_errors.py` — to be carried in the live status response. This one is about
`_HMS_FAILURE_REASONS` in `main.py`, a different table with a different shape (short category
labels) feeding a different consumer (`archive.failure_reason`). `0300_8003` is already in the
table #2926 is about; it is missing from this one.

Worth noting because the two interact: if #2926 lands and this does not, the live view will name
the fault as "Spaghetti defects were detected by the AI Print Monitoring" while the archive of that
same print still shows no reason at all. Fixing this one closes that gap rather than duplicating
the work.

**Observed on one machine.** I only have the X2D to hand for this — the fault needs the onboard AI
to actually fire, which I have not had happen on my other printer. The map itself is a static table
shared by every printer, so I would expect any model whose firmware emits `0300_8003` to hit it,
but I have not watched that happen and am not claiming it.

ojimpo:
## Description

`0300_8003` — the code the onboard AI raises when it halts a print for spaghetti — is missing from `_HMS_FAILURE_REASONS`, so such a print is archived with `failure_reason=None`. Its immediate neighbour `0300_8004` (filament runout) is in the map, and the two arrive through the same path one hex digit apart. The sentence for the code is already in the tree twice, in `hms_errors.py:45` and `HMSErrorModal.tsx:85`. This is a missing key, not a missing meaning.

**Rewritten on top of `5211fd45`.** `_HMS_FAILURE_REASONS` stores canonical keys now, so both rows take `"spaghettiDetached"` and the assertions that named the English text follow. Merged rather than rebased — the branch already carries a merge of `dev`, and rebasing it would have dropped that.

**The reason for reusing an existing option is unchanged, but the sentence explaining it was wrong and is now right.** What made the choice correct was never that `"Spaghetti / Detached"` matched the editor's text character for character; it was that the editor already offers this failure mode. A derived reason that lands on an offered option opens the dropdown on it instead of blank — and survives, because an unrecognised value is cleared on the next save. Naming that option by its key says the same thing without the coincidence in the middle.

**The null-on-no-match fallback is untouched and I am not asking to change it.** The comment above the map explains why matching on `module` alone archived user cancellations as "Layer shift", and `derive_failure_reason`'s docstring finishes the thought — don't guess, null is honest. That design is right; this adds codes whose meaning is already documented to the map that had not been told about them.

## Related Issue

Fixes #2946

#2974, filed from this PR's review, is fixed on `dev` and is what this branch is now written against.

## Documentation

**Pick one**:
- [x] No docs update required — reason: the wiki has no page describing which HMS codes auto-populate `failure_reason`; `failure-analysis.md` documents the *tags* and the failure categories, and `failure-detection.md` is about the Obico ML integration, not the printer's own HMS. The user-visible vocabulary is unchanged because the key is one the editor already offers.

## Type of Change

## Changes Made

- `backend/app/main.py` — two rows in `_HMS_FAILURE_REASONS`: `"0300_8003"` and `"0C00_8042"`, both `"spaghettiDetached"`.
  - A comment says why a module-0x0C row is safe next to the header's warning about module-0x0C: the warning is about matching on the module alone, `0C00_8042` is a full short code with a documented meaning (`hms_errors.py:577`), and the H2D cancel echo is `0C00_001B`, so nothing collides.
  - A second comment records the two exclusions so they do not have to be re-derived: `0C00_C004` says "Possible" and describes a print that is still running, and `0300_800A` is AI monitoring but reports a filament pile-up in the waste chute.
- `backend/tests/unit/test_failure_reason_derivation.py`
  - `test_ai_spaghetti_detection_is_classified` / `test_the_ai_monitors_other_code_is_classified_too` — each code derives `spaghettiDetached`.
  - `test_the_ai_monitors_warnings_are_left_unclassified` — parametrised over the two exclusions; asserts they are absent from the map *and* that `derive_failure_reason` returns `None` for them, so the boundary is pinned rather than described in a comment.
  - `test_every_derived_reason_is_an_option_the_editor_offers` is **removed**, not translated — you are right that the property it checked is a plain Python assertion now, and `test_every_derived_reason_is_a_canonical_key` makes it without a skipif.
  - **Re-aimed as you suggested**, in its own commit: `test_the_backend_vocabulary_matches_the_one_the_frontend_offers` reads the exported `FAILURE_REASON_KEYS` array out of `EditArchiveModal.tsx` and compares it with `_FAILURE_REASON_KEYS`, both directions. `test_every_offered_key_has_english_text` reads the keys — not the labels — out of the `editArchive.failureReasons` block of `en.ts`.

## Screenshots

None — no UI change. The archive's Failure Reason field now shows the same option the editor's dropdown already offers, where it previously showed nothing.

## Testing

- [x] I have tested with my printer model: X2D reported the original archive; the branch is tested here on H2S. The map is a static table shared by every printer.

**Why the re-aimed test is worth its skipif.** The comment above `_FAILURE_REASON_KEYS` asks that it be kept in sync with `EditArchiveModal.tsx`, and asking is all it does. The drift is silent in both directions, and they fail differently: a key the dropdown offers but the backend rejects turns a save into a 400 the modal has no error surface for, and a key the backend accepts but the dropdown omits is a classification the editor discards the next time anyone opens that archive. Neither is visible from Python, because the other end is TypeScript.

**Measured against a mutated tree, one mutation at a time:**

```
backend drops "warping"
  E  AssertionError: backend-only: [], frontend-only: ['warping']
  FAILED test_the_backend_vocabulary_matches_the_one_the_frontend_offers
  1 failed, 22 passed

backend gains "beltSlip"
  E  AssertionError: backend-only: ['beltSlip'], frontend-only: []
  FAILED test_the_backend_vocabulary_matches_the_one_the_frontend_offers
  1 failed, 22 passed

en.ts loses warping's text
  E  AssertionError: offered by the dropdown with no editArchive.failureReasons entry: ['warping']
  FAILED test_every_offered_key_has_english_text
  1 failed, 22 passed
```

And the two rows this PR is actually about, removed from `_HMS_FAILURE_REASONS`:

```
3 failed, 20 passed
  test_ai_spaghetti_detection_is_classified
  test_the_ai_monitors_other_code_is_classified_too
  test_ai_detection_and_its_runout_neighbour_are_distinct
```

**Where it runs.** `Dockerfile.test` copies `backend/`, `pyproject.toml` and the requirements files and nothing else, so `frontend/` is absent inside the test image and the two guards have nothing to read. They skip on `frontend/package.json` being absent — the same signal the launcher-config tests use to tell an image apart from a checkout — and run on every source checkout, which is where `test_backend.sh` runs. In the image: `21 passed, 2 skipped`. With the frontend tree mounted: `23 passed`.

**Suite on this branch merged with `dev` (up to `5211fd45`):**

```
backend/tests/unit    8237 passed, 4 skipped in 673.30s
ruff check            All checks passed!
ruff format --check   840 files already formatted
```

Everything that reads `failure_reason` also run together, with the frontend tree mounted so nothing skipped — including `test_failure_reason_vocabulary_migration.py`, which is new on `dev`:

```
backend/tests/integration/test_failure_reason_vocabulary_migration.py
backend/tests/integration/test_archives_api.py
backend/tests/integration/test_print_queue_api.py
backend/tests/unit/test_print_log.py
backend/tests/unit/test_archive_run_aggregation.py
backend/tests/unit/test_subtask_archive_resume.py
backend/tests/unit/services/test_finance_service_billing.py
  -> 313 passed in 130.70s
```

## Checklist

## Additional Notes

**On the `Stale - ...` question — thank you for answering it with a migration rather than a rule.** Both of the extra findings I raised are closed by `5211fd45`: the third vocabulary is gone from `main.py`, and `noStatusUpdate` is in `_FAILURE_REASON_KEYS` rather than merely tolerated, which is the part I would not have thought to ask for. The 91-vs-1 split on your instance is the observation I could not produce — this installation has a single failed archive and it carries no reason at all.

**On `0C00_8042`:** I still have not seen it fire. It is taken on the documented meaning in `hms_errors.py:577` and your reading of it, not on an observation of mine. `0300_8003` is from the X2D archive that reported this.


maziggy (changes requested):
Reviewed. The reasoning in the description holds up — I checked it rather than taking it
on trust. 0300_8003 is in hms_errors.py:45 and HMSErrorModal.tsx:85 with the AI-monitoring
sentence, the label matches en.ts:5442 character for character, and every other value in
the map resolves 1:1 to a failureReasons key too, so you preserved an invariant rather than
inventing one. Grouping by category instead of code order is the house style. Full backend
suite green here, ruff clean, and removing your row fails exactly your three tests and
nothing else.

Two things, then a note.

**1. Please take 0C00_8042 as well.**

You asked in the issue whether it was one row or two and I never answered — that is on me.
Take both. hms_errors.py:577 is "The AI print monitor has detected a spaghetti defect",
a documented meaning, so it is not the guessing the map header rules out. The module-0x0C
warning in that header is about matching on the module alone; a full short code is a
different thing, and the H2D cancel echo is 0C00_001B, so nothing collides.

Leave 0C00_C004 ("Possible spaghetti failure was detected.", hms_errors.py:580) out —
"possible" reads as a warning rather than a halt. Same for 0300_800A, which is AI
monitoring but is a waste-chute pile-up, not a failed print. Please say so in the comment
above the group so the next person does not have to re-derive it.

**2. test_ai_detection_label_is_the_editors_own_vocabulary does not do what it says.**

Its docstring says it pins the string to editArchive.failureReasons.spaghettiDetached in
en.ts, but it compares a Python literal to a Python literal — change en.ts and it still
passes. Functionally it is test 1 with a different attr. Nothing on the frontend covers it
either; no test references FAILURE_REASON_KEYS or failureReasons, so the coupling the whole
design rests on is untested from both ends.

Please replace it with the real guard: assert every value in _HMS_FAILURE_REASONS appears
in the failureReasons block of en.ts. Six lines or so, all five strings resolve today, and
it covers "Layer shift" and "Filament runout" at the same time.

**3. On your two pre-existing notes — same root cause, and I would rather fix it properly.**

Storing the camelCase keys in the map instead of English labels closes both at once:
EditArchiveModal.tsx:69 early-returns on a known key so the reverse lookup stops depending
on locale, t() actually translates the archive display, and the failure_analysis.py:116
split between derived and user-set reasons collapses. It needs a decision about the English
strings already in the database, so it is not this PR — but every literal we add is one
more row a migration has to handle later. Please open it as a separate issue with what you
found; do not widen this one.

maziggy:
Thanks for filing #2974 - it was accurate end to end, and it is fixed on dev now.

Heads-up that this PR needs a small rebase as a result. _HMS_FAILURE_REASONS stores
the camelCase keys now, so your two rows become "spaghettiDetached", and the four
assertions in the new tests change with them. Both files will conflict.

The Stale - ... question you flagged is answered too: both sentences write a new
noStatusUpdate key, and a startup migration folds 168 historical labels onto the keys
across both columns - every label in all 14 locales resolves to exactly one key, so it
is exact rather than a guess. On my own instance print_log_entries held 91 rows reading
"User cancelled" beside 1 reading "userCancelled", which is the split you described but
could not reproduce.

On test_every_derived_reason_is_an_option_the_editor_offers: the property it checks is
now a plain Python assertion against _FAILURE_REASON_KEYS in api/routes/print_log.py, so
that particular check no longer needs to read en.ts or carry the Docker skipif.

The test is still worth having though, re-aimed. Nothing pins the backend
_FAILURE_REASON_KEYS against the frontend FAILURE_REASON_KEYS - print_log.py only asks
in a comment that the two be kept in sync, and that comment is the only thing standing
between us and the next version of this bug. Your test is exactly the right shape for
it, reading the keys out of en.ts rather than the labels, and the skipif guard still
earns its place. If you want to fold that into this PR, please do.

maziggy (changes requested):
Reviewed. This one is straightforward and I checked the parts that carry the argument rather than the description.

All four codes say what you say they say - hms_errors.py:45, :52, :577 and :580 - and the H2D cancel echo really is 0C00_001B per HMSErrorModal.tsx:942, so the module-0x0C comment is accurate. spaghettiDetached is already in all three vocabularies (print_log.py:238, the FAILURE_REASON_KEYS array, en.ts:5487), so reusing an offered option holds and no locale work follows. Dockerfile.test really is two stages with frontend/ absent from backend-test, and test_launcher_shutdown_config.py:45 uses the identical probe, so the skipif has precedent rather than being invented here.

I re-ran your four mutations against patched copies rather than your numbers, and each one is caught: dropping warping from the TSX array, dropping its text from en.ts, and removing the two new rows all fail the guard you would expect. Non-vacuous.

One thing worth pushing on before merge, and it is about the reasoning rather than the rows.

The boundary between what you mapped and what you excluded is drawn by reading Bambu's English and preserved in a comment. But severity is already in every dict that reaches derive_failure_reason - bambu_mqtt.py:5482 and printer_manager.py:1532 both populate it, and main.py:1798 already filters on e.severity >= 2. derive_failure_reason reads only attr and code. Your new tests hard-code "severity": 2 into every fixture, which reads as though the field is load-bearing when nothing consults it.

That matters because 0300_8003's own text is "Please check the quality of the printed model before continuing your print" - a print that continues. It is the first advisory-shaped code in a map whose other entries are all halts, and derive_failure_reason returns the first matching code in list order. A print that drew a spaghetti advisory, kept going and later failed from a clog now archives as spaghettiDetached if the advisory sits earlier in the HMS list. The same argument you used to exclude 0C00_C004 applies here with less force but not zero force.

I am not asking you to redesign the map. You have the X2D archive that reported this - what is the severity on its 0300_8003 entry, and on a 0C00_C004 if you ever catch one? If the numbers separate them, the exclusion rule can be expressed instead of described, and the fixtures stop implying a dependency that is not there. If they do not, say so and the comment is the right answer.

Minor: _keys_the_frontend_can_translate matches (\w+):\s*' so it only sees single-quoted values, and failureReasons:\s*{(.*?)} stops at the first }. The block is all single quotes with no interpolations today so both hold, but a label with an apostrophe would be double-quoted in TS and drop out of the set, failing the guard with "no en.ts entry" for a key that has one. These guards are meant to outlive us; ['"] costs nothing.

Issue #2955, opened 2026-08-25
[Bug]: The two Inventory stock alert producers have no call site either — nothing in the backend computes a reorder point, so on_stock_reorder_alert and on_stock_break_alert cannot fire once #2945 lets them be saved


ojimpo:
### Product

Bambuddy

### Area

Notifications

### Bug Description

Found while writing the PR for #2945, and it corrects something I believe we both had wrong about that issue.

On #2913 you wrote:

> `on_stock_reorder_alert` and `on_stock_break_alert` have columns, migrations, service methods at `notification_service.py:2188` and `:2213`, templates and UI toggles, but appear nowhere in `schemas/notification.py` or `api/routes/notifications.py`. [...] Those toggles do not persist at all, so that is a **plumbing bug rather than a missing producer**. Different fix, own issue.

The plumbing half is right and is what the #2945 PR fixes. But the producer half is not: `notification_service.on_stock_reorder_alert` and `on_stock_break_alert` are **defined and never called**. So this is the same shape as #2913 *as well as* a plumbing bug — once the toggles persist, they still cannot fire.

Every reference to either name in `backend/` on `dev`:

```
backend/app/models/notification.py:102,103        the two columns
backend/app/core/database.py:3887,3890,3894,3897  the two migrations (sqlite + postgres)
backend/app/services/notification_service.py:2188 async def on_stock_reorder_alert(...)
backend/app/services/notification_service.py:2213 async def on_stock_break_alert(...)
```

That is the complete list. No caller, and no dynamic dispatch either — nothing does `getattr(notification_service, event)`.

The reason there is no caller is more interesting than an oversight, and is why I am not simply writing one. **The forecast that would decide when to fire does not exist in the backend at all.** It is computed in the browser, in `ForecastPanel.tsx`:

```
ForecastPanel.tsx:288   const reorderPointG = ...
ForecastPanel.tsx:296   daysUntilROP  = Math.floor((totalRemainingG - reorderPointG) / dailyRateG)
ForecastPanel.tsx:301   const reorderAlert    = !stockBreakAlert && daysUntilROP !== null && daysUntilROP <= 0
ForecastPanel.tsx:352   alerts = forecasts.filter(f => !f.settings?.alerts_snoozed && (f.stockBreakAlert || f.reorderAlert))
```

[...]

PR #1184, opened 2026-05-01
[Feature] Add Stock forecasting and Logistics view

Keybored02:
## Description

Adds a full Forecast & Reorder Intelligence tab to the Inventory page. The panel analyses historical filament consumption, computes statistical reorder points, projects stock runout dates, and integrates a shopping list workflow — all with permission controls and i18n coverage across 8 locales.

## Related Issue

Fixes #1172 

## Documentation

**Companion docs PRs** (delete lines that don't apply):
- Wiki: maziggy/bambuddy-wiki#22

**Pick one**:
- [x] Docs PR(s) linked above
- [ ] No docs update required — reason: ___

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition or update

## 1. Stock Consumption Forecast Engine

**Daily rate calculation — two-tier:**

- **History-based** (preferred, requires ≥2 print events): iterates consecutive print-event pairs,
  computing `g/day` per interval. Each observation is weighted by `exp(−λ × age_days)` where
  `λ = ln(2)/30`, giving a 30-day half-life so recent prints dominate. Computes weighted mean and
  weighted variance → std dev.
- **Delta-based** (fallback): `total_grams_used / days_since_spool_added` — used when only one
  print event exists.

**Per-SKU forecast object computed for every material/brand/subtype group:**

```
effective_lead_time        = max(global_lead_time, sku_lead_time)  [min: 1 day]
statistical_safety_stock   = Z_95 (1.645) × std_dev × √(effective_lead_time)
safety_margin              = user_margin_days × daily_rate  OR  user_margin_g directly
safety_stock               = statistical_safety_stock + safety_margin
reorder_point              = daily_rate × effective_lead_time + safety_stock
days_remaining             = current_stock_g / daily_rate
days_until_ROP             = (current_stock_g − reorder_point) / daily_rate
```

**Two alert levels:**

| Level | Condition | Meaning |
|---|---|---|
| `reorderAlert` | `days_until_ROP ≤ 0` | Stock has dropped below reorder point — order now |
| `stockBreakAlert` | `days_remaining ≤ effective_lead_time` | Stock will run out before reorder arrives |

---

## 2. Forecast Table UI

[...]

PR #2956, opened 2026-08-25
Let the two stock alert toggles reach the database (issue #2945)

### LINKED ISSUE #2945: [Bug]: The two Inventory stock alert toggles cannot be saved — on_stock_reorder_alert and on_stock_break_alert are missing from the provider schema, so the API drops them
### Product

Bambuddy

### Area

Notifications

### Bug Description

`on_stock_reorder_alert` and `on_stock_break_alert` exist everywhere except the two modules that
would let a user turn them on. They have model columns, service producers, templates, UI toggles
and a frontend test that asserts the toggle sends a PATCH — but they appear nowhere in
`schemas/notification.py` or `api/routes/notifications.py`. Pydantic drops unknown fields by
default, so the PATCH succeeds, nothing is written, and the GET never returns the field.

This is the sibling defect to #2913 and the opposite way round. There, `on_filament_low` is in the
schema and persists, but no code ever calls the producer, so the toggle saves and never fires.
Here the producers exist at `notification_service.py:2188` and `:2213`; what is missing is any way
to switch them on.

Present:

```
backend/app/models/notification.py:102        on_stock_reorder_alert = Column(Boolean, default=False)
backend/app/models/notification.py:103        on_stock_break_alert   = Column(Boolean, default=False)
backend/app/services/notification_service.py:2188   async def on_stock_reorder_alert(...)
backend/app/services/notification_service.py:2213   async def on_stock_break_alert(...)
frontend/src/api/client.ts:2831,2894,2950     declared on the Response, Create and Update types
frontend/src/__tests__/components/NotificationProviderCardStockAlerts.test.tsx
                                              "toggling Reorder Alert sends correct PATCH payload"
```

Absent:

```
backend/app/schemas/notification.py           no match for on_stock
backend/app/api/routes/notifications.py       no match for on_stock
```

The gap dates from #1184 (merged 2026-05-05), which added the feature. It touched
`models/notification.py`, `notification_service.py`, both frontend components and the dedicated
stock-alerts test — and neither `schemas/notification.py` nor `api/routes/notifications.py`. Every
layer it did touch works. The two it missed are the two a toggle has to cross.

`_provider_to_dict` (`routes/notifications.py:32`) serialises 28 `on_*` keys and neither stock key
is among them. (The responses below carry 29. The extra one is `on_ha_sensor_alert`, which the
serialiser also omits and the response model fills from its default — unrelated to these two
fields, and #2827 already adds that line, so I have left it alone.)

The running instance agrees. On 1.2.6b1 with a real, enabled webhook provider:

```
GET /api/v1/notifications/1  ->  200
  id=1  name=teams-bridge  type=webhook  enabled=true
  29 on_* keys, neither stock key present
```

and the generated schema says the same, so this is not a serialisation accident:

| Schema | `on_*` fields | stock keys |
|---|---|---|
| `NotificationProviderCreate` | 29 | none |
| `NotificationProviderUpdate` | 29 | none |
| `NotificationProviderResponse` | 29 | none |

The frontend has a passing test asserting it sends the field. Nothing on the backend is listening
for it.

### Expected Behavior

Both keys are accepted by `NotificationProviderCreate` and `NotificationProviderUpdate`, persisted
to the columns that already exist, and returned by `_provider_to_dict`, so the Inventory Alerts
toggles behave like every other toggle on the card. The producers already in
`notification_service.py` then have a provider to resolve.

### Steps to Reproduce

1. Start Bambuddy 1.2.6b1 with any notification provider configured (webhook is enough — no
   printer required, this is not printer-specific).
2. `GET /api/v1/notifications/{provider_id}`. Count the `on_*` keys: 29, and neither
   `on_stock_reorder_alert` nor `on_stock_break_alert` is among them.
3. `GET /openapi.json` and read `components.schemas.NotificationProviderUpdate.properties`. Same
   29 `on_*` fields, neither stock key.
4. In the source, `grep -rn on_stock backend/app/schemas/notification.py
   backend/app/api/routes/notifications.py` returns nothing, while the same grep over
   `backend/app/models/notification.py` and `backend/app/services/notification_service.py` returns
   the columns and the two producers.

Steps 2 and 3 are what I ran on my own instance. I have deliberately not exercised the write path
there — it is a shared machine in an office and the provider is in real use — so I am not claiming
to have watched a toggle revert in the UI. The schema is the evidence: a field the request model
does not declare cannot be persisted.

- Confirmed against the running 1.2.6b1 instance and against `dev` at the time of writing; the two
  modules have no `on_stock` reference in either.
- Checked this is not the same defect as #2913. It is the mirror image: producer present, schema
  absent, where #2913 is schema present, producer absent. Different fix, and #2913's fix does not
  touch these two fields.
- Searched existing issues for "stock alert notification" and found no report of this.
- Verified the frontend genuinely sends the field rather than only rendering a toggle — the
  dedicated test file pins the PATCH payload for both toggles.

### Printer Model

Not printer-related

### Bambuddy Version

1.2.6b1 (`ghcr.io/maziggy/bambuddy:daily`, digest
`sha256:033f6de1a48a10ef5abd1db2ddabf2b27815254b7a50ce75ea4874552a818bb4`, image built
2026-08-23). Also verified against `dev`.

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian) -- Ubuntu 24.04 on WSL2

No support package. Nothing is logged, because the field is discarded during request validation
before any handler sees it — there is no code path to instrument. The evidence is the schema
comparison above, which reproduces in a clean checkout in seconds and can be read straight off
`/openapi.json` on any running instance. Happy to attach one if you want the surrounding
configuration.

### Additional Context

**This is not a re-report of #2913.** That issue is about `on_filament_low` having no call site.
These two fields have call sites and no way to be enabled. The two fixes touch different modules
and neither subsumes the other. I raised the possibility in #2913's Additional Context and you
asked for it separately, with the schema and route modules already identified — this issue is that
request, with the reproduction taken on my own instance rather than repeated from your comment.

**Confirmed on two machines.** The defect was first noticed while investigating #2913 on an X2D at
home; the observations above are from an H2S in the office, a different printer, a different host,
and a different Bambuddy build. It is not machine-specific or configuration-specific.

I have not checked whether the two producers are themselves reached — the inventory code that
would call them is outside what I looked at. Even if they are wired correctly, the toggles cannot
be turned on today.

ojimpo:
## Description

**This PR carries no production change.** It did when it was opened; #2827 landed the same four lines of wiring while it was in review, so `on_stock_reorder_alert` and `on_stock_break_alert` are on `dev` already and #2945 is fixed there. What is left is the part #2827 did not bring: **a check that fails on the *next* field of this kind rather than on the one already reported.**

The defect class is a column that exists everywhere except one of the modules a toggle has to cross. #1184 added the stock pair's columns, migrations, producers, templates and UI and skipped `schemas/notification.py`, so Pydantic dropped the fields, the PATCH answered 200 and wrote nothing. `on_ha_sensor_alert` was in the other half of the same gap — schema but neither route map — at the same time. No per-field test catches either one, because the field nobody remembered to wire is also the field nobody remembers to test.

So the check is derived from `NotificationProvider.__table__` rather than listed. A toggle added tomorrow is covered the day its column lands.

## Related Issue

Fixes #2945

Related: #2955 — the alerts still cannot fire; see Additional Notes.

## Documentation

**Pick one**:
- [x] No docs update required — reason: tests only, and the wiki's Event Triggers tables have no Inventory section to correct. Adding one now would document the two alerts as working, which they are not yet — see #2955.

## Type of Change

Tests only, on top of the fix that landed in #2827.

## Changes Made

Both hunks answer your review; the previous revision is gone rather than amended.

- `backend/tests/integration/test_notifications_api.py` — `test_create_persists_and_returns_the_toggle` and `test_patch_is_reflected_by_every_read_route` now parametrise over `EVENT_TOGGLE_COLUMNS`, read off `NotificationProvider.__table__`, instead of a hardcoded list of four. Same two tests, 32 columns instead of 4, covering create, both read routes and PATCH behaviourally.
- `backend/tests/unit/test_notification_provider_field_parity.py` — **new**, and now only the two structural checks plus a guard against a vacuous enumeration: every `on_*` column is a field on `NotificationProviderCreate` and on `NotificationProviderUpdate`. The two `inspect.getsource` tests are deleted.

The docstring says why two schemas are the whole of it: the update route applies changes with a generic `model_dump(exclude_unset=True)` + `setattr` loop rather than a third hand-maintained map, so once a field survives the schema it is written.

`test_every_event_column_is_settable_on_create` asserts on `NotificationProviderCreate`, not `NotificationProviderBase`. Same object today, but Create is what the route binds and it survives Create growing fields of its own.

## Screenshots

None — tests only.

## Testing

- [x] I have tested with my printer model: H2S (no printer is involved in this path)

I re-ran your two mutations against the new shape. **Deleting the `on_stock_break_alert` line from `_provider_to_dict` fails in two tests:**

```
FAILED test_notifications_api.py::TestNotificationsAPI::test_create_persists_and_returns_the_toggle[on_stock_break_alert]
FAILED test_notifications_api.py::TestNotificationsAPI::test_patch_is_reflected_by_every_read_route[on_stock_break_alert]
2 failed, 158 passed
```

**Rewriting that same line as `bool(provider.on_stock_break_alert)` now passes**, which was the point:

```
160 passed
```

That is the refactor unblocked. Replacing both maps with a loop over the columns no longer trips anything here.

And with the #2945 defect itself put back on current `dev` — both fields removed from `NotificationProviderBase`, `NotificationProviderUpdate`, `_provider_to_dict` and the create constructor, nothing else touched:

```
FAILED test_notification_provider_field_parity.py::test_every_event_column_is_settable_on_create[on_stock_break_alert]
FAILED test_notification_provider_field_parity.py::test_every_event_column_is_settable_on_create[on_stock_reorder_alert]
FAILED test_notification_provider_field_parity.py::test_every_event_column_is_settable_on_update[on_stock_break_alert]
FAILED test_notification_provider_field_parity.py::test_every_event_column_is_settable_on_update[on_stock_reorder_alert]
FAILED test_notifications_api.py::TestNotificationsAPI::test_create_persists_and_returns_the_toggle[on_stock_break_alert]
FAILED test_notifications_api.py::TestNotificationsAPI::test_create_persists_and_returns_the_toggle[on_stock_reorder_alert]
FAILED test_notifications_api.py::TestNotificationsAPI::test_patch_is_reflected_by_every_read_route[on_stock_break_alert]
FAILED test_notifications_api.py::TestNotificationsAPI::test_patch_is_reflected_by_every_read_route[on_stock_reorder_alert]
FAILED test_notifications_api.py::TestNotificationsAPI::test_a_row_with_null_event_flags_is_still_listable
9 failed, 151 passed
```

The ninth is your own `test_a_row_with_null_event_flags_is_still_listable` from #2827 — it names the columns directly, so removing them takes it down too. The eight above it are the derived ones.

Clean, on this branch with your `dev` merge (`1ac6949a`) underneath it:

```
the two touched test files together      160 passed
backend/tests (whole suite)              11173 passed, 22 skipped
ruff check                               All checks passed!
ruff format --check                      830 files already formatted
```

The whole-suite figure is from the same tree one merge earlier; the three mutation runs above and the 160 are on the current head.

## Checklist

## Additional Notes

Confirming your finding rather than restating mine: the only references to either producer in `backend/app/` are their own definitions, so this PR makes the toggles savable and does not make the alerts fire. #2955 is the one users feel, and it is a design call — the forecast those producers need is computed in the browser in `ForecastPanel.tsx`, not in the backend — so it stayed out of here.


maziggy:

ojimpo:
Rebased onto dev. The conflict was not really a conflict: the fix landed while this was open.

#2827 added `on_stock_reorder_alert` and `on_stock_break_alert` to `NotificationProviderBase`, to `NotificationProviderUpdate`, to `_provider_to_dict` and to the create constructor — the same four places this PR touched — and its integration tests parametrise `test_create_persists_and_returns_the_toggle` and `test_patch_is_reflected_by_every_read_route` over both of them. So the production half of this branch and the three integration tests it carried are both redundant, and both are gone.

What is left is the part #2827 did not bring: the enumeration.

`backend/tests/unit/test_notification_provider_field_parity.py` reads the `on_*` columns off `NotificationProvider.__table__` and asserts each one appears in the Create schema, the Update schema, `_provider_to_dict` and the create constructor. It is derived rather than listed, so a toggle added tomorrow is covered the day its column lands — which is the only form of this check that fails on the *next* one of these rather than on the one already reported.

The previous revision carried an exemption for `on_ha_sensor_alert`, which was in the other half of the same gap — schema but neither map. #2827 closed that too, so the exemption is deleted rather than grown.

Verification, on this branch merged with dev:

```
129 passed
```

And with the #2945 defect put back on top of current dev — the two fields removed from both schemas and both route maps, nothing else touched:

```
FAILED test_every_event_column_is_settable_on_create[on_stock_break_alert]
FAILED test_every_event_column_is_settable_on_create[on_stock_reorder_alert]
FAILED test_every_event_column_is_settable_on_update[on_stock_break_alert]
FAILED test_every_event_column_is_settable_on_update[on_stock_reorder_alert]
FAILED test_every_event_column_is_serialised_in_the_response[on_stock_break_alert]
FAILED test_every_event_column_is_serialised_in_the_response[on_stock_reorder_alert]
FAILED test_every_event_column_is_passed_through_on_create[on_stock_break_alert]
FAILED test_every_event_column_is_passed_through_on_create[on_stock_reorder_alert]
8 failed, 121 passed
```

ruff check and ruff format --check are clean.

If you would rather not carry a test with no production change attached to it, say so and I will close this — #2945 itself is fixed on dev either way, and #2955 is the part that is still open.

maziggy (changes requested):
Thanks @ojimpo - the enumeration is the right instinct and the reasoning behind it is correct:
a hand-maintained per-field map is a standing hazard, and a derived check is the only kind that
fails on the next one rather than the one already reported. Two notes.

1. The two source-scraping tests pin the shape of the maps rather than what they do. I mutation-
   tested both ways. Deleting the on_stock_break_alert line from _provider_to_dict fails the parity
   test, correctly. But rewriting that same line as bool(provider.on_stock_break_alert) - identical
   behaviour - also fails it, while the whole notifications integration suite stays green. So they
   would block the one refactor that ends this bug class for good: replacing both maps with a loop
   over the columns. The docstring calls those maps a standing hazard and then cements them.

   The better version already exists one file over. #2827's test_create_persists_and_returns_the_toggle
   and test_patch_is_reflected_by_every_read_route are parametrised over a hardcoded list of four
   fields - which is the same hazard one layer up. Derive that list from
   NotificationProvider.__table__ instead and you get the same guarantee behaviourally: I ran it,
   64 tests in 14.6s, all green today, it catches the deleted-line mutation in two tests, and it
   passes the behaviour-identical one. That covers create, both read routes and PATCH for all 32
   columns and makes both getsource tests unnecessary.

   Keep the two model_fields tests as they are - those are structural checks of structural facts,
   no source scraping, and they are the half that pins the reported defect. Worth saying in the
   docstring that the update route needs nothing beyond them, because it is a generic
   model_dump(exclude_unset=True) + setattr loop rather than a third map.

2. test_every_event_column_is_settable_on_create asserts against NotificationProviderBase while its
   docstring says the Create schema. NotificationProviderCreate is `pass` today so it is the same
   thing, but asserting on Create is what the route actually binds and it survives Create growing
   fields of its own.

The PR body is stale after the rebase - it still describes the four lines of wiring, the three
integration tests and the "11 tests fail before the change" list, none of which apply now. Your
comment explains it, but the body is what gets read later.

And thank you for the Additional Notes. I checked: the only references to either producer in
backend/app/ are their own definitions, so they are dead code and merging this does not make the
alerts fire. #2955 is the one that matters for users, and you were right to split it.

maziggy (changes requested):
Thanks @ojimpo - all four points from the last round are done, and I checked each against the
diff rather than the body. The getsource pair is gone, both integration tests parametrise off
NotificationProvider.__table__, the parity tests assert on NotificationProviderCreate, and the
docstring's claim about the update route is true: it really is model_dump(exclude_unset=True)
plus setattr, no third map. The body is accurate now. I also checked both hand-maintained maps
against the table - neither is missing a column today.

One finding, and it is in the direction this PR exists to cover.

1. The derived enumeration is vacuous for 9 of the 32 columns. These default to True on the
   model AND on the response schema:

     on_print_complete, on_print_failed, on_print_stopped, on_billing_charge_failed,
     on_ams_drying_suspended, on_plate_not_empty, on_queue_job_waiting,
     on_queue_job_skipped, on_queue_job_failed

   Both tests drive the toggle to True and assert True, so for those nine the default supplies
   the expected answer on its own, from either side of the route. A field dropped by a
   hand-maintained map still reads back True.

   Your own mutation from the last round, moved to a True-default column:

     delete "on_print_complete": provider.on_print_complete from _provider_to_dict
       -> 160 passed, whole notifications suite, mutation survives
     delete "on_stock_break_alert": provider.on_stock_break_alert  (False default)
       -> 2 failed, 158 passed

   Same map, same deletion, opposite outcome, decided entirely by the column default. Your patch
   test's docstring puts it exactly right - "the reads were the broken half" - and for nine
   columns it cannot see that half.

   The fix keeps the derivation and only changes what is asserted: drive each toggle to whatever
   its default is not.

     TOGGLE_TARGET = {
         c.name: not bool(c.default.arg if c.default is not None else False)
         for c in NotificationProvider.__table__.columns
         if c.name.startswith("on_")
     }

   Then target = TOGGLE_TARGET[field] in both tests, seeding the patch factory with not target.
   I wrote it and measured it:

     unmutated                                     160 passed
     read map, on_print_complete    (True default)   2 failed   was surviving
     read map, on_stock_break_alert (False default)  2 failed   still caught
     create constructor, on_print_complete           1 failed   was surviving
     bool(provider.on_print_complete) refactor     160 passed   still tolerated

   That last line is the one that matters for the last round's ask: it catches the omissions
   without pinning the shape of the maps, so the loop-over-columns refactor stays unblocked.

Two nits:

2. startswith("on_") does not filter on type. A future on_* column that is not Boolean turns
   field: True into a 422 and breaks both tests in an unrelated PR - and a spurious failure is
   how a derivation gets edited until it passes. Adding isinstance(c.type, Boolean) keeps it
   honest.

3. The derivation now lives in two files, three once TOGGLE_TARGET joins it. A duplicated
   derivation is a list with extra steps. backend/tests/_fixtures/ is importable from both unit
   and integration if you want one home for it.

ojimpo:
Thanks — you were right, and it is a better finding than the one the PR started with. Nine of the
thirty-two columns were being answered by their own default, so the tests could not see the half
they exist for.

**1. `TOGGLE_TARGET`, as you specified it.**

The derivation stays; only the asserted value changes. Each toggle is driven to whatever its
model-side default is not, so the round trip is the only thing that can produce the answer. Both
tests take `target = TOGGLE_TARGET[field]`, and the patch test seeds the factory with `not target`.

Your numbers reproduce here exactly:

```
unmutated                                       161 passed
read map, on_print_complete    (True default)     2 failed    was surviving
read map, on_stock_break_alert (False default)    2 failed    still caught
create constructor, on_print_complete             1 failed    was surviving
bool(provider.on_print_complete) refactor       161 passed    still tolerated
```

161 rather than your 160 because of one addition — see below. The last line is the one that
mattered from the previous round, and it still holds: the omissions are caught without the shape
of the maps being pinned, so the loop-over-columns refactor stays unblocked.

**2. Filtered on `Boolean`.**

`EVENT_TOGGLE_COLUMNS` is now `startswith("on_") and isinstance(column.type, Boolean)`, with the
reason in a comment: a non-Boolean `on_*` column would turn `{field: True}` into a 422 and break
both tests in an unrelated PR, and a spurious failure is how a derivation gets edited until it
passes.

**3. One home, in `backend/tests/_fixtures/`.**

`backend/tests/_fixtures/notification_toggles.py` carries both derivations, imported by the unit
parity tests and the integration round-trips. A derivation duplicated per file is a list with
extra steps, which is the thing this PR is about.

**One test I added that you did not ask for.**

`test_the_targets_cover_both_directions` asserts `set(TOGGLE_TARGET.values()) == {True, False}`.
`TOGGLE_TARGET` is now what stops the round-trips being vacuous, so it needs the same guard the
enumeration already has: if the default lookup ever collapsed to one constant, every target would
agree with the column default again and nothing would fail. That is the +1 in the count above.

Verified the derivation against the model — 32 `on_*` columns, and the nine with `default=True` are
exactly the nine you listed.

Full backend unit suite on this branch with the `dev` merge under it: **8676 passed, 22 skipped**.
One `PytestUnhandledThreadExceptionWarning` comes with the merge and is not from anything here —
the same suite on this branch before the merge was 8500 passed with no warning. `ruff check` and
`ruff format --check` clean.

maziggy (changes requested):
Reviewed. The claims hold and I checked them against the model rather than the description.

The diff really is tests plus one fixture, fixtures/ is an existing package so the new module follows the convention, and the derivation is exact on today's model: 32 on* columns, all Boolean, 23 defaulting False and 9 True, which is the "nine" your comment names. Your reason for checking only two schemas is right - notifications.py:431 is a generic model_dump(exclude_unset=True) + setattr loop, so the schema is the only gate on the write path and the read map is covered behaviourally by the round-trips.

I also swept every provider column against both schemas looking for a second live instance of #2945. There is none: the only columns absent from Create and Update are last_error, last_error_at and last_success, which are server-owned. And #2955 is confirmed - the two producers appear in backend/app only as their own definitions and the migrations, so this makes the toggles savable and the alerts still cannot fire.

Two things that would make the derivation cover what it promises.

The on_ prefix is a naming convention, but the hazard you describe is per-column: a field that exists everywhere except one of the maps it has to cross. enabled, quiet_hours_enabled, quiet_hours_start and daily_digest_enabled cross the same two maps and nothing guards them. I checked - widening the structural test to every column except id, created_at, updated_at and the three status fields passes unchanged today. Keep TOGGLE_TARGET scoped to the round-trips, where it needs a type-aware target; it is only the presence check that wants to be wider.

Second, model_default reads column.default only. Exact today because all 32 columns declare a Python default, but a column added with server_default and no Python default reports False, the target flips to True, and for a server-default-true column the test sends True and asserts True. That is the vacuity the fixture exists to prevent, restored silently for that one column. Your comment on the Boolean check says a spurious failure is how a derivation gets edited until it passes; this is the quieter twin, a silent pass. An assert that every on* column has a Python default costs a line and is green now.

Last, housekeeping: the description is a commit behind. Changes Made lists two files and the branch has three - _fixtures/notification_toggles.py is new and carries both the derivation and the drive-each-toggle-away-from-its-default reasoning, which is the part a reviewer most needs pointed at. The mutation table above it also predates that commit.

PR #2958, opened 2026-08-25
Drop Python 3.10 support and validate Python 3.14

dj-oyu:
## Summary
- drop Python 3.10 support and require Python 3.11+
- use stdlib enum.StrEnum directly
- make native installers select Python 3.11+
- validate Python 3.14.7 with uv

## Validation
- uv-managed Python 3.14.7 dependency install
- ruff check backend/
- compileall and application import
- 52 related tests passed

Upstream追随済みの専用ブランチです。


dj-oyu:
Closing this PR because the target repository was incorrect. The change will be reopened against dj-oyu/bambuddy.

PR #2990, opened 2026-08-28
File Manager previews for STEP, PDF and spreadsheet files

### LINKED ISSUE #2976: [Feature]: File Manager previews for STEP, PDF and spreadsheet files
### Problem or Use Case

The File Manager is the natural place to keep a whole print job together, but right now only the "printable" formats get a visual preview (STL thumbnails, 3MF/GCode viewer). A real job folder usually holds more than the sliced model:

- the source CAD geometry as STEP (.step/.stp),
- a drawing, assembly instruction or datasheet as PDF,
- a parts / cost / batch list as CSV or XLSX.

For all of those the File Manager shows only a generic file icon. To check whether a folder contains the right revision I have to download the file and open it in another program. On a farm with many jobs and a tablet at the printer, that is a lot of round trips - and it pushes documents that belong to the job into a separate tool.

### Proposed Solution

Extend the File Manager preview to three more file classes:

1. STEP (.step / .stp) - generate a thumbnail on upload and offer a 3D preview in the detail pane (rotate/zoom), ideally reusing the viewer already used for STL/GCode. Batch thumbnail generation for existing files, the same way STL already supports it.
2. PDF (.pdf) - first-page thumbnail in the grid plus an inline page viewer, so drawings and instructions can be read without leaving Bambuddy.
3. Spreadsheets (.csv / .xlsx / .ods) - render the sheet as a read-only table in the preview pane. Sheet tabs when a workbook has several sheets; showing the first N rows with scrolling would already be enough.

Nice to have: the same previews wherever such files are attached to a print/project, not only in the File Manager grid.

### Alternatives Considered

- Mounting the folder as a network share and opening the files locally - works from a desktop, but breaks the "one place for the whole job" idea and is not an option from a tablet or phone at the printer.
- Converting STEP to STL and PDF to PNG up front just to get a thumbnail - an extra manual step, and the original file is no longer the one being previewed.
- Keeping documents in a separate system (NAS / ERP) - then they are no longer next to the print file, which is exactly what the File Manager is good at.

### Feature Category

File Manager

### Priority

Critical for my use case

### Mockups or Examples

Client-side libraries that would fit the existing viewer approach:

- STEP -> occt-import-js (OpenCascade compiled to WASM) rendered with three.js, as used by the Online 3D Viewer project
- PDF -> pdf.js
- XLSX/CSV -> SheetJS (or any lightweight CSV parser for the plain-text case)

### Contribution

### Checklist

Thomansky:
## Description

Makes the document types that ship alongside a job folder viewable in place: STEP gets an interactive 3D preview, PDF an inline page viewer, and CSV/XLSX/ODS a read-only table view — each with a client-rendered grid thumbnail persisted through a new, tightly validated endpoint.

## Related Issue

Fixes #2976

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#80

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- **STEP (.step/.stp)** — interactive 3D preview (rotate/zoom/reset) in the existing three.js model viewer; triangulation via occt-import-js (OpenCascade as WASM) in a dedicated Web Worker, per-part colours preserved
- **PDF** — inline page viewer (page navigation, zoom) built on pdf.js
- **CSV/XLSX/ODS** — read-only table view with one tab per sheet (PapaParse / SheetJS CE); capped at 500 rows / 40 columns with truncation notices; oversized files (20 MB sheets / 50 MB PDF) show a notice; broken files fail soft
- **Thumbnails** — the first successful client render is posted to new `POST /api/v1/library/files/{id}/preview-thumbnail` (PNG-only, 2 MB cap, 512 px cap, PIL re-encode, only the new file types, only fills missing thumbnails, gated by the library update ownership pair)
- `_SCANNABLE_EXTENSIONS` gains `.pdf`, `.csv`, `.xlsx`, `.ods`; grid/list get per-type badges and fallback icons
- **CSP**: `script-src` gains `'wasm-unsafe-eval'`; the STEP worker asset alone is served an eval-permitting policy (occt's embind glue needs `new Function`), pinned by two new tests — the document policy is unchanged
- **Browser baseline**: `build.target: 'safari16'` so pdf.js's class static blocks are lowered to the declared Safari 16.0 floor
- i18n: new keys under `fileManager.preview.*` in all 14 locales, `npm run check:i18n` green
- All libraries load on demand (the project's first `React.lazy` / dynamic imports); main bundle +11 kB (i18n strings only) vs `dev`

## Screenshots

**Before** — generic icons, no preview action:

| Grid | Context menu on a CSV |
|---|---|
| before: grid with generic icons | before: csv context menu |

**After** — per-type previews:

| STEP (interactive 3D) | PDF |
|---|---|
| after: step preview | after: pdf preview |

| XLSX (sheet tabs) | CSV |
|---|---|
| after: xlsx preview | after: csv preview |

**After** — client-rendered thumbnails in the grid:

after: grid thumbnails

## Testing

- Backend: 8 tests pinning the thumbnail endpoint contract (`test_library_preview_thumbnail_api.py`) + 2 CSP tests (`test_security_headers.py`)
- Frontend: component tests for both new modals (real PapaParse/SheetJS parse paths; pdf.js mocked at the module boundary since jsdom cannot rasterise); full suite 3491/3491 green incl. i18n parity
- All four preview classes verified end-to-end against the running app (upload → preview → thumbnail persisted)
- Backend suite green apart from pre-existing Windows-only platform failures that reproduce identically on a clean `dev` checkout

- [x] I have tested with my printer model: no printer-facing behavior changes

## Checklist

## Additional Notes

- Dependencies: `xlsx` (SheetJS CE, Apache-2.0) pinned to the maintained 0.20.3 tarball from cdn.sheetjs.com (the npm-registry package is frozen at 0.18.5 with known CVEs); `pdfjs-dist` (Apache-2.0), `papaparse` (MIT), `occt-import-js` (LGPL-2.1) — all AGPL-3.0-compatible; no new Python dependencies (the endpoint uses the already-present Pillow)
- Per CSP3, a dedicated worker is governed by the policy delivered with its own script response, so eval stays confined to that DOM-less worker context; browsers that still inherit the document CSP into same-origin workers fail soft with the viewer's error message
- Explicitly out of scope per the issue: STEP-to-STL conversion, spreadsheet editing, PDF form rendering
- Bundle detail: `xlsx` 493 kB, `pdf` 428 kB + 1.26 MB worker, STEP worker 59 kB + 7.6 MB WASM (3.1 MB gzip), papaparse 19 kB — each loaded on first use of its preview type only


maziggy:
I don't have any problems with using AI for assistance - but to be honest - I don't want to have AI commits in rhe repo. Please remove. Thanks.

Thomansky:
Done — sorry about that. The trailers were only there for transparency's sake, but I completely understand your point. I've rewritten the commits on all my open PR branches (code and wiki) without them and dropped the footer from the PR descriptions; the content of the commits is unchanged. Won't happen again.

maziggy (changes requested):
Thanks, this is careful work - the worker-confined eval, the code splitting and
the CLIENT_THUMBNAIL_TYPES allowlist that cannot overwrite a server thumbnail
are all the right calls. Verified green here: build, eslint, 3673 vitest, i18n
parity, and the two backend test files. Main chunk really is +11 kB, and
build.target costs 0 bytes - I measured both builds.

Two I would not merge without:

1. pdf.worker.min.mjs is imported ?url, so build.target never transforms it and
   it still ships `class WorkerMessageHandler{static{...}}` (Safari 16.4+). The
   target does fix the bundled pdf-*.js chunk - with it removed, that one
   carries `static {` too. Separately, check-browser-baseline.mjs:65 filters on
   .js, so the .mjs asset is never scanned and the build printed a green tick
   over it. Please widen that filter; it is the bit that will bite later.
   Effect today is only a broken PDF preview on iOS 16.0-16.3, not a blank app.

2. PdfPreviewModal: loadingTask is assigned after `await fetch` and `await
   import('pdfjs-dist')`, so closing the modal in that window leaves cleanup
   with null and the resolved task is never destroyed - a leaked pdf.js worker
   per open/close. `if (cancelled) { loadingTask.destroy(); return; }`.

Also worth fixing:

3. getDocument() gets no cMapUrl / standardFontDataUrl / wasmUrl, and
   pdfjs-dist's cmaps/, standard_fonts/ and wasm/ are not copied into static/.
   CJK PDFs lose text, JPEG2000 images and ICC colour do not decode.
4. Because of 3, nothing under the document CSP compiles WebAssembly today -
   the main chunk has zero WebAssembly references and the STEP worker has its
   own branch with 'unsafe-eval'. So 'wasm-unsafe-eval' on the document policy
   is currently unused and the comment credits the wrong consumer. (The STEP
   worker branch itself is justified - the emscripten newFunc(Function, ...)
   invoker factory is in the built worker, I checked.)
5. preview-thumbnail: PIL's DecompressionBombError subclasses Exception only,
   so it escapes the except tuple - a 68-byte PNG with an IHDR claiming
   20000x20000 returns 500. Worse, 12000x7000 is under PIL's limit and decodes
   for real: a few-KB POST allocating ~340 MB. Please check img.size before
   load(). And move img.save() out of that try - a full disk should not come
   back as 400 "Invalid thumbnail image".
6. Tests: no 403 case for the ownership branch (nor the created_by_id is None
   path), and onSnapshot is untested on both modals, so the only caller of the
   new endpoint has no coverage. Stubbing toBlob would cover the sheet path.

Minor, your call: two concurrent uploads both write a PNG and orphan one; the
CSV path parses in full then slices, unlike the XLSX comment right below it;
the PDF canvas at zoom 4 x dpr 2 exceeds iOS Safari's canvas area cap and comes
back blank. occt-import-js being LGPL-2.1 is fine to use, but it is the first
copyleft library in the frontend - I want to look at that separately.

Thomansky:
Thank you — the `?url` point in particular is one I would not have found
without you measuring the emitted asset rather than the chunk list. All six
points are addressed, in four commits.

**1. Safari baseline / worker asset.** Both halves confirmed. The scanner
filtered `f.endsWith('.js')` on a flat `readdirSync` of `static/assets`, so
`pdf.worker.min-<hash>.mjs` was never opened; it now walks the tree and covers
`.js`, `.mjs` and `.cjs`.

For the worker itself I checked `legacy/` first, hoping to just swap the
import — but `pdfjs-dist/legacy/build/pdf.worker.min.mjs` carries
`class WorkerMessageHandler{static{…}}` too, and so does `legacy/build/pdf.min.mjs`.
So there is a `generateBundle` step in `vite.config.ts` that runs emitted
script *assets* through `transformWithOxc` at the configured `build.target`
and re-minifies them. `transformWithEsbuild` is gone in Vite 8 unless esbuild
is installed separately, so this uses the oxc path the rest of the build
already uses. Result: 1 262 199 → 1 171 579 bytes, i.e. 90 kB *smaller* than
what pdfjs-dist publishes, and zero `static {` in the emitted asset. I also
verified the scanner fails on a regression by appending a `static {}` block to
the built worker — the build stops and prints the offending snippet.

Emitted assets under `assets/pdfjs/` are excluded from that transform: they
are pdf.js's own wasm fallbacks, `openjpeg_nowasm_fallback.js` is asm.js-shaped
emscripten output, and they already meet the baseline (checked: no `static {`,
no lookbehind). The scanner still reads them, so a regression there fails the
build rather than passing quietly.

**2. Leaked pdf.js worker.** Fixed exactly as you described — `if (cancelled)
{ loadingTask.destroy(); return; }` right after the assignment, where the
check and the assignment are one synchronous step so nothing can slip between
them. The `if (cancelled)` after `await loadingTask.promise` stays, and its
comment now says why it does *not* destroy (cleanup ran with the task
assigned, so it already did). Pinned by a test that holds the `fetch` promise
open, unmounts, then resolves it and asserts `destroy` was called once.

I looked for the same shape elsewhere: `ModelViewer`'s STEP worker is a
module-level singleton with its own `onerror` teardown, not created per modal,
so there is nothing to leak there. `SpreadsheetPreviewModal` holds no
destroyable resource.

**3. pdf.js resource URLs.** `cmaps/`, `iccs/`, `standard_fonts/` and `wasm/`
are published to `static/assets/pdfjs/` by a second small plugin, so the
existing `/assets` mount serves them and no new backend route is needed. The
dev server answers the same paths from a map built once at startup, which
doubles as the allowlist — a request cannot name a file outside the package
directory. `getDocument()` now passes `cMapUrl`, `iccUrl`,
`standardFontDataUrl` and `wasmUrl` (`cMapPacked` already defaults to `true`).

Cost: 200 files, 3.8 MB in `static/` — 1.5 MB cmaps, 804 kB standard fonts,
1.6 MB wasm, 20 kB iccs. All of it is fetched on demand by the worker, none of
it is in any chunk. I kept the `LICENSE_*` files that ship in `wasm/` since
they cover the binaries next to them. Say the word if you would rather drop
`quickjs-eval.*` (469 kB, only used with `enableScripting`) — I left it in
because a partial copy of a directory is the kind of thing that bites two
releases later.

**4. CSP.** You are right that nothing under the document policy compiles
wasm — but dropping `wasm-unsafe-eval` outright would have broken 3, because
until now the pdf.js worker asset fell through to the document policy and that
is what was permitting its compile. So: the pdf worker gets its own branch,
matched on `^/assets/pdf\.worker\.min-[\w-]+\.mjs$`, with
`script-src 'self' 'wasm-unsafe-eval'` and no JS eval — it needs none — and
the document policy goes back to `'self'` plus the nonce. The STEP worker
branch is unchanged.

One caveat worth naming: if the pdf.js worker ever fails to load, pdf.js falls
back to running it on the main thread, where the document policy now blocks
the wasm compile. That path loses JPEG2000/JBIG2 and ICC decoding rather than
the preview, and it is already the degraded path, so I did not widen the
document policy for it.

The tests pin all three shapes (eval-free document, wasm-only pdf worker,
eval-allowed STEP worker) on whole tokens, and check that the lookalike
`pdf.worker.min-<hash>.js` chunk — Vite emits one next to the asset — does
*not* pick up the worker policy.

**5. preview-thumbnail hardening.** Confirmed both: `Image.open()` itself
raises `DecompressionBombError` for 20000×20000 (it calls
`_decompression_bomb_check` on the header), and that class derives straight
from `Exception`. The declared size is now checked against a 2048px edge cap
while `Image.open()` has only read the header, so 12000×7000 is a 400 before a
single pixel is allocated; the bomb error is named in the guard for the sizes
PIL refuses first. `img.save()` moved out of the guard and a failure there is
a 500 with its own log line. Tests for both, plus one that stubs `save` to
raise `ENOSPC` and asserts the response is not "Invalid thumbnail image".

**6. Test gaps.** Added: operator posting for another user's file → 403;
operator posting for a `created_by_id is NULL` file → 403; operator for their
own file → 200; admin for the ownerless one → 200. These reuse
`TestOwnershipPermissionsSetup`, the way `test_github_restore_api.py` does.
`onSnapshot` is covered on both modals with `getContext`/`toBlob` stubbed —
the PDF one asserts a PNG blob, that it is sent once and not again when you
page away and back, and that nothing is sent when the canvas yields no blob;
the sheet one asserts the mini table is actually drawn, that an empty leading
sheet is skipped in favour of the first sheet with rows, and that a file with
no rows sends nothing.

**Minors.** All three code ones are in. Concurrent uploads: the row is claimed
with a conditional `UPDATE … WHERE thumbnail_path IS NULL` and the loser
unlinks its PNG, so nothing is orphaned. I did not add a test for it — racing
two ASGI requests against this suite's shared in-memory SQLite connection
needs enough contortion that the test would be testing the contortion. CSV:
rows are counted and capped in a `step` callback now, so the file is never
materialised as rows the preview throws away; the row/column totals in the
truncation notice stay exact. Canvas cap: the raster scale is clamped to
16 777 216 px of canvas area, so past the cap a zoom stops gaining sharpness
instead of returning a blank page. The CSS size is untouched, so the page
still grows. Where the clamp bites depends on the fitted width: an A4 page on
an iPad-width panel at dpr 2 reaches it around zoom 1.7, on a phone-width one
not before zoom 4 at all.

occt-import-js / LGPL-2.1: understood, nothing changed there. Happy to swap it
or vendor the wasm differently if you land somewhere else on it.

Verified here: `ruff check` clean, `tsc --noEmit` clean on both projects,
eslint clean, `npm run build` green including the widened baseline check, i18n
parity green, 273 frontend files / 3744 tests green, and 149 backend tests
across the preview-thumbnail, PDF-thumbnail, security-header, ownership and
pdf_thumbnail files.

---

One thing I should flag rather than let you discover it in the diff: while your review was open I also pushed a few things onto this branch that go beyond it, because my own farm was tripping over them — PDF thumbnails rendered server-side on upload (pypdfium2) so a PDF card is not a blank icon until someone opens it, one shared preview window that actually uses the screen (the old one was a small box on a wide display), wheel/pinch zoom and a fullscreen toggle incl. double-click, opening a preview by double-click or from the selection toolbar instead of only the kebab, and previews for image files.

If you would rather review this PR as the thing you already read, say so and I will split those commits onto a follow-up branch and rewind this one to the review fixes — no hard feelings either way, it is your queue.

Thomansky:
Brought up to date with `dev`, which also means today's Swedish locale: the keys this branch adds are written in `sv.ts` and `check:i18n` is green at 15 locales. Nothing else changed since the six points above were answered.

Taking it out of draft.

Issue #1754, opened 2026-06-15
[Feature]: Custom fields for filament inventory


RB14060:
### Problem or Use Case

Right now I use Spoolman to track my inventory and I'm able to add custom fields for whatever I want, in this case I'm using them to track the last time I dried the filament. It would be nice to be able to add the same kind of fields in Bambuddy.

### Proposed Solution

Add the ability to create custom fields with a user defined title and a selection of options for field type, like text, date, date+time, etc.

### Alternatives Considered

Currently using Spoolman but would like to switch to Bambuddy's built in filament manager so it's one less thing I have to run and maintain.

### Feature Category

Spool Inventory

### Priority

Would improve my workflow

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Before we commit to building this, I'd like to gauge community interest. If you'd find this feature useful, please give this issue a thumbs up (👍) reaction so we can prioritize accordingly.

PR #2994, opened 2026-08-29
Material number as a first-class spool field

### LINKED ISSUE #2870: [Feature]: Material / article number as a first-class field on spool inventory (sortable, filterable, usable in statistics)
### Problem or Use Case

We run Bambuddy commercially on a 9-printer Bambu Lab farm (H2S / H2D / X1E).

Internally, every material we buy has its own number - 15 = Bambu Lab PLA Basic, 16 = Bambu Lab ABS-GF, and so on. That number, not the brand + material + colour combination, is how our costing and ERP system identifies a material. Every gram of filament we consume has to be booked against it.

Bambuddy's inventory currently has nowhere to put that number:

- **Category** would technically work, but it is clearly meant for something else - the placeholder itself suggests "Production, Prototype, Customer A". Using it for the material number means giving up the category function entirely.
- **Note** is free text, but it is not sortable, not filterable, and not something a cost report can be grouped by.

So today the number either lives nowhere, or it has to be squeezed into a field that already has a different job.

### Proposed Solution

A dedicated, optional free-text field on the spool / inventory record - "Material number" / "Article number", ~64 characters, no validation and no uniqueness requirement (every shop numbers differently):

1. Editable in the **Edit spool** dialog, next to Category / Cost per kg.
2. Shown as an optional column in the spool inventory list, sortable, and included in the inventory search/filter.
3. Usable as a grouping key in **Statistics**, so filament consumption and cost can be reported *per material number* instead of per brand + material + colour.
4. Included in CSV/Excel export and exposed in the REST API (`/api/v1/inventory/spools`), so it can be pulled into an ERP system.
5. Inherited automatically when a new spool of the same material / sub-type / brand / colour is created or auto-matched via RFID - otherwise every newly detected spool has to be numbered by hand.

Point 3 is the one that actually matters: the number is only worth having if consumption can be reported against it.

### Alternatives Considered

- **Category field** - works as a container, but it is already the field for production / customer grouping. One of the two use cases has to lose.
- **Note field** - free text, but not sortable, filterable or groupable.
- **#1754 (Custom fields for filament inventory)** - a generic custom-field system would cover this *if* those fields are sortable, filterable and usable in statistics and export. If #1754 is implemented that way, this request is a subset of it and can be closed as a duplicate. If custom fields end up being display-only, a dedicated article-number field would still be needed.
- **Spoolman** - Spoolman's filament model already has exactly this field (`article_number`, `String(64)`, filterable, plus an endpoint listing all distinct article numbers). Running Spoolman alongside Bambuddy just for one field is a lot of overhead - and if Bambuddy had the field, the Spoolman sync could map it 1:1 to `article_number`.
- **Encoding the number in the slicer preset name** - Bambu Studio has no custom field either, so the number would have to be smuggled into the preset name and parsed back out of a display string. Fragile and not machine-readable.

### Feature Category

Spool Inventory

### Priority

Would improve my workflow

### Mockups or Examples

In the **Edit spool** dialog the field would sit between "Cost per kg" and "Category":

```
Cost per kg        [ 32.90 EUR  ]
Material number    [ 15         ]   <- new
Category           [ Production ]
Note               [ ...        ]
```

Reference implementation: Spoolman stores this as `article_number` on the filament object (max 64 chars, filterable, with its own endpoint returning all distinct article numbers).

### Contribution

### Checklist

Thomansky:
﻿## Description

Adds the internal material number as a first-class spool field, as discussed in the issue: an optional free-text purchasing identifier shared by all spools of the same product, with grouping in statistics as the core deliverable.

## Related Issue

Fixes #2870

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#81

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- `Spool.material_number` (nullable VARCHAR(64), no uniqueness) with an idempotent `_safe_execute` migration in `run_migrations()`; identical DDL on SQLite and PostgreSQL
- Schemas: field on `SpoolBase`/`SpoolUpdate`, included in all spool responses
- Spool dialog: input in the Additional Details section with autocomplete from numbers already in use; hidden in Spoolman mode (see below)
- Inventory list: optional sortable "Material No." column, filter chip (specific number or "no number"), free-text search matches it, bulk edit can assign it
- Inheritance: a new spool of an already-numbered product — NULL-safe (material, subtype, brand, color_name) match, newest donor wins, explicit value always wins — arrives numbered; covers manual create, bulk create, the API and the RFID auto-add
- Statistics: `GET /api/v1/inventory/stats/material-numbers` (active spool count + remaining weight; consumed grams + cost from the recorded usage history, archived spools included) plus a "By Material Number" dashboard widget
- CSV export/import round-trip via a new `material_number` column
- Spoolman mode: read-only mapping from the filament-level `article_number`; the number is maintained in Spoolman itself there
- i18n: all strings in `en.ts` and the 13 other locales, `npm run check:i18n` green

## Screenshots

| Before | After |
| --- | --- |
| before: the Additional section of the spool dialog without the field | after: Material No. input with autocomplete in the Additional section |
| before: end of the statistics dashboard, no material-number widget | after: By Material Number statistics widget |

## Testing

- Backend: migration add + idempotency, CRUD, inheritance (incl. archived-donor and explicit-wins cases), statistics aggregation, CSV round-trip — 13 tests
- Frontend: form field and statistics widget tests; full frontend suite and `tsc -b` green
- Branch rebased onto current `dev` before opening this PR; targeted suites re-run green after the rebase
- Full backend suite green on Linux-relevant tests; the only failures on my Windows dev box are the pre-existing platform-specific ones (chmod/procfs/systemd) that fail identically on a clean `dev` checkout

- [x] I have tested with my printer model: no printer-facing behavior changes; the RFID auto-add path is covered by the inheritance tests

## Checklist

## Additional Notes

No generic custom-field system, as noted in the issue — if #1754 lands first, this field becomes redundant.

### DIFF
--- a/backend/app/api/routes/_spoolman_helpers.py
+++ b/backend/app/api/routes/_spoolman_helpers.py
@@ -54,6 +54,10 @@ class MappedSpoolFields(TypedDict):
     created_at: str | None  # None when Spoolman spool has no registered timestamp
     updated_at: str | None
     cost_per_kg: float | None
+    # Spoolman's native filament.article_number, surfaced as the internal
+    # material number (#2870). Read-only in Spoolman mode — the number is
+    # filament-level there and maintained in Spoolman itself.
+    material_number: str | None
     storage_location: str | None
     location_id: int | None
     k_profiles: list[Any]
@@ -340,6 +344,9 @@ def _map_spoolman_spool(spool: dict) -> MappedSpoolFields:
         # Spoolman has no updated_at field; use registered timestamp as best available proxy
         "updated_at": created_at,
         "cost_per_kg": _safe_optional_float(spool.get("price")),
+        # Spoolman's filament.article_number maps 1:1 onto the internal
+        # material number (#2870): both identify the purchasable product.
+        "material_number": (filament.get("article_number") or None),
         "storage_location": spool.get("location") or None,
         "location_id": None,
         "k_profiles": [],
--- a/backend/app/api/routes/inventory.py
+++ b/backend/app/api/routes/inventory.py
@@ -31,6 +31,7 @@ from backend.app.models.spool_k_profile import SpoolKProfile
 from backend.app.models.user import User
 from backend.app.schemas.location import LocationCreate, LocationResponse, LocationUpdate
 from backend.app.schemas.spool import (
+    MaterialNumberStats,
     SpoolAssignmentCreate,
     SpoolAssignmentResponse,
     SpoolBulkCreate,
@@ -55,6 +56,7 @@ from backend.app.services.location_service import (
     prepare_internal_spool_payload,
     rename_location as rename_location_record,
 )
+from backend.app.services.material_number import apply_material_number_inheritance
 from backend.app.services.slicer_filament_resolver import resolve_slicer_filament
 from backend.app.services.slot_nozzle import resolve_slot_nozzle
 from backend.app.services.spool_csv import (
@@ -1340,6 +1342,8 @@ async def create_spool(
         payload = await prepare_internal_spool_payload(db, spool_data.model_dump(), set(spool_data.model_fields_set))
     except ValueError as exc:
         raise HTTPException(status_code=400, detail=str(exc)) from exc
+    # A new spool of an already-numbered product inherits its material number (#2870).
+    payload = await apply_material_number_inheritance(db, payload)
     spool = Spool(**payload)
     db.add(spool)
     await db.commit()
@@ -1362,6 +1366,8 @@ async def bulk_create_spools(
         payload = await prepare_internal_spool_payload(db, data.spool.model_dump(), fields_set)
     except ValueError as exc:
         raise HTTPException(status_code=400, detail=str(exc)) from exc
+    # A new spool of an already-numbered product inherits its material number (#2870).
+    payload = await apply_material_number_inheritance(db, payload)
     for _ in range(data.quantity):
         spool = Spool(**payload)
         db.add(spool)
@@ -2177,6 +2183,67 @@ async def get_spool_usage_history(
         .limit(limit)
     )
     return list(result.scalars().all())
+
+
+@router.get("/stats/material-numbers", response_model=list[MaterialNumberStats])
+async def get_material_number_stats(
+    db: AsyncSession = Depends(get_db),
+    _: User | None = RequirePermissionIfAuthEnabled(Permission.INVENTORY_READ),
+):
+    """Aggregate the inventory by material number (#2870).
+
+    The material number is the internal purchasing identifier shared by all
+    spools of a product, so this is the grouping the business actually costs
+    by — unlike brand+material+colour. Two queries: active-spool counts and
+    remaining weight from the spool table, consumption and cost from the
+    recorded usage history (archived spools included — their consumption
+    happened). Sorted by consumption, heaviest first.
+    """
+    from backend.app.models.spool_usage_history import SpoolUsageHistory
+
+    has_number = Spool.material_number.is_not(None) & (Spool.material_number != "")
+
+    inventory_rows = await db.execute(
+        select(
+            Spool.material_number,
+            func.count(Spool.id),
+            func.sum(Spool.label_weight - Spool.weight_used),
+        )
+        .where(has_number, Spool.archived_at.is_(None))
+        .group_by(Spool.material_number)
+    )
+
+    usage_rows = await db.execute(
+        select(
+            Spool.material_number,
+            func.sum(SpoolUsageHistory.weight_used),
+            func.sum(SpoolUsageHistory.cost),
+        )
+        .join(Spool, SpoolUsageHistory.spool_id == Spool.id)
+        .where(has_number)
+        .group_by(Spool.material_number)
+    )
+
+    stats: dict[str, MaterialNumberStats] = {}
+    for number, count, remaining in inventory_rows.all():
+        stats[number] = MaterialNumberStats(
+            material_number=number,
+            spool_count=count,
+            remaining_g=max(0.0, float(remaining or 0)),
+            consumed_g=0.0,
+            cost=0.0,
+        )
+    for number, consumed, cost in usage_rows.all():
+        entry = stats.get(number)
+        if entry is None:
+            entry = MaterialNumberStats(
+                material_number=number, spool_count=0, remaining_g=0.0, consumed_g=0.0, cost=0.0
+            )
+            stats[number] = entry
+        entry.consumed_g = float(consumed or 0)
+        entry.cost = float(cost or 0)
+
+    return sorted(stats.values(), key=lambda s: s.consumed_g, reverse=True)
 
 
 @router.get("/usage", response_model=list[SpoolUsageHistoryResponse])
--- a/backend/app/core/database.py
+++ b/backend/app/core/database.py
@@ -4865,6 +4865,12 @@ async def run_migrations(conn):
     # on fresh installs only — this covers databases whose table predates it.
     await _migrate_location_ha_sensor_unique_binding(conn)
 
+    # Migration: Add material_number to spool (#2870). Nullable free text —
+    # the internal purchasing/article number a business costs by, shared by
+    # all spools of the same product. VARCHAR(64) is spelled identically on
+    # SQLite and Postgres.
+    await _safe_execute(conn, "ALTER TABLE spool ADD COLUMN material_number VARCHAR(64)")
+
     # Migration: repair the tare of spools the RFID auto-add gave the wrong
     # Bambu spool row (#2909). Runs last so the spool catalogue it reads is
     # whatever this database actually holds.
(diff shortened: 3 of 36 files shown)


Thomansky:
No pressure at all — just flagging that this one is still mergeable and rebased on current `dev`, in case it dropped off the list.

If there is anything that would make it quicker to look at — splitting it up, dropping a part you are unsure about, or me writing a short "what to check" list — say the word and I will do that instead of adding anything new to it. I know the review queue is the scarce resource here, not the code.

maziggy (changes requested):
Thanks for this, and for the patience while it sat in the queue. Reviewed in full
against current dev. The code is careful, the comments explain why rather than what,
and it is additive throughout -- nothing here changes existing behaviour. Happy to
merge once the points below are addressed.

Verified on the branch: ruff clean, backend 12236 passed, vitest 3683 passed, eslint
and build clean, i18n parity green across all 14 locales. I also dry-ran the migration
against a real PostgreSQL instance inside a rolled-back transaction: the column applies
as character varying(64) nullable, existing rows read back NULL, a second ALTER raises
42701 which _safe_execute swallows through its savepoint, and the transaction survives.
Both aggregate queries run unchanged on Postgres, and spool.updated_at is NOT NULL so
the ORDER BY updated_at DESC donor pick has no NULLS FIRST/LAST divergence between the
dialects. That part is solid.

Please fix:

1. The statistics widget ignores the dashboard timeframe. StatsPage.tsx:1184 puts the
   widget in the same grid as archiveStats, FailureAnalysis, PrintActivity and
   FilamentTrends, all of which take effectiveDateRange. The endpoint has no date
   parameter, so "By Material Number" shows lifetime totals next to widgets showing
   "Last 30 days". Point 3 of the issue is a per-period number. The inventory half
   (spool_count, remaining_g) is rightly point-in-time; only the usage-history half
   needs dateFrom/dateTo, plus the range in the query key.

2. MaterialNumberStats.tsx:38 reports an API failure as "no data". isError, empty data
   and a genuinely empty result all render "No material numbers assigned yet. Add them
   to spools in the inventory...". A 403 from a missing INVENTORY_READ, a 500 or a
   dropped connection then tell the user they have not done something they have. The
   third test currently pins this as intended -- please give the error case its own
   message instead.

3. Nothing normalises the value server-side. SpoolFormModal trims, but bulk edit, CSV
   import (spool_csv.py:363 passes the cell through raw) and direct API calls do not.
   "15" and "15 " become two groups in the stats endpoint and two chip entries, and the
   chip's exact match at InventoryPage.tsx:1290 never matches the padded one because the
   option list is built from trimmed values. One field_validator on SpoolBase/SpoolUpdate
   that strips and maps "" to None fixes every path at once, and makes the != "" guards
   in both queries redundant.

4. Spoolman mode: bulk edit offers a field that cannot be written. The spool dialog
   correctly hides Material No. there and you have a test for it, but
   BulkEditSpoolsModal gets no spoolmanMode prop, so FIELDS still lists it.
   SpoolmanInventoryUpdate has no material_number, so the payload dumps to {} and the
   route answers 400 "update must include at least one field". category and
   low_stock_threshold_pct have the same defect today, so this is pre-existing rather
   than something you introduced -- but with a third field joining them it is worth
   fixing generically by passing spoolmanMode and filtering FIELDS.

5. Spoolman mode: the statistics widget is permanently empty. The endpoint reads the
   internal spool table, which is empty in that mode, while the inventory list does show
   Material No. mapped from filament.article_number. The user sees numbers in inventory
   and "No material numbers assigned yet" in statistics. Either hide the widget in
   Spoolman mode or aggregate the mapped spools.

Smaller:

- models/spool.py:64 says new spools inherit "see prepare_internal_spool_payload".
  That function does not do this -- inheritance lives in services/material_number.py
  and is called from the routes.
- The PR description says the RFID auto-add path is covered by the inheritance tests.
  It is not: the tests exercise POST /spools and /spools/bulk only.
  spool_tag_matcher.py:233 is the one path that hardcodes brand="Bambu Lab" in the
  lookup, and the one nobody is watching when it fires. Worth a test.
- CSV import is the fifth create path and does not inherit (inventory.py:1264 builds
  Spool(**row.spool) directly). Probably right, since the file is authoritative -- but
  please say so in a comment so the next reader does not read it as an oversight.
- No test for the Spoolman article_number mapping. One assert in
  backend/tests/unit/test_spoolman_inventory_helpers.py, which already covers
  _map_spoolman_spool.
- Backup and restore drop the number. github_backup._collect_spools and github_restore
  are both explicit field whitelists and material_number is in neither -- nor are
  category, storage_location, location_id or low_stock_threshold_pct today. Pre-existing,
  but this field is the business's own data by definition, so it is worth adding to both.
- '__none__' as the filter sentinel collides with a literal material number __none__.
  Theoretical, but free text is free text.
- AdditionalSection.tsx:326: the div inside the new {!spoolmanMode && ( is not indented
  under it.
- The widget's query key is not in the inventory_changed invalidation set. staleTime is
  60s so it self-heals; take it or leave it.

One design note, not a change request: a blank number on a new spool of an already
numbered product always inherits, so "deliberately unnumbered" cannot be expressed at
creation -- you clear it afterwards. That seems right to me, it just deserves a line in
the wiki page.

Thomansky:
Thanks for going through this so carefully. Reproducing each point first was worth it — four came up exactly as you described, and one turned out to be half right.

**1. The widget ignored the dashboard timeframe.** Right, and it was as visible as you said: lifetime totals sitting next to cards headed "Last 30 days". The endpoint takes `date_from`/`date_to` now and applies them to the usage half only — `spool_count` and `remaining_g` stay whole, because "how much do I hold" has no date range. The range is part of the query key, so the cache can't answer a 90-day dashboard with the 30-day numbers. One thing that only became visible once a narrow range makes most rows zero: ties now sort by number, so the order is stable instead of falling back to whatever the database returns. `TestMaterialNumberStatsTimeframe` covers `date_from`, `date_to`, no range, stock staying unaffected, and the tie-break; `StatsPageMaterialNumbers.test.tsx` asserts the request really carries the dashboard's range and carries none on "all time".

**2. An API failure read as "no data".** Agreed, and you were right that the third test was pinning it rather than catching it. `isError` has its own message now (`stats.materialNumbers.loadFailed`, translated into all 14 locales), and that test is rewritten to assert the failure message appears and the "assigned yet" hint does not.

**3. Nothing normalised the value server-side.** Fixed the way you suggested: one `field_validator` on `SpoolBase` and `SpoolUpdate` that strips and maps blank to NULL. It covers the form, bulk edit and direct API calls in one place, and it let me drop the `!= ""` guards in both the aggregate and the inheritance lookup. One correction — CSV import was already safe. `spool_csv.cell()` strips at line 318, before the passthrough loop at 363, so a padded cell never reached the model; I pinned that with a test anyway so it stays true. `TestMaterialNumberNormalisation` covers create, patch, bulk edit, CSV, and that a padded duplicate no longer becomes a second group. Spoolman mode doesn't pass through the validator at all, so I trimmed `filament.article_number` at the mapper too — same defect, same chip mismatch, just the one path the validator can't reach.

**4. Bulk edit offered a field that cannot be written.** Reproduced exactly: the payload dumps to `{}` and the route answers 400, for `category` and `low_stock_threshold_pct` just as much as for the new field. Done generically, as you asked — a `FieldSpec` marks itself `internalOnly`, the modal takes `spoolmanMode` and filters the list once, so the three disappear together and anything added later declares itself the same way. That closes the two older instances with it. `BulkEditSpoolsModal.test.tsx` pins both modes.

**5. The widget was permanently empty in Spoolman mode.** Hidden rather than aggregated. The consumption and cost columns come from the internal usage history, and there is none for Spoolman spools, so aggregating the mapped list would have produced a half-empty table rather than an answer — worse than not offering the card. The test asserts the card is gone and the endpoint is never called.

**Smaller:** the model comment points at `services/material_number.py` now instead of `prepare_internal_spool_payload`. The RFID auto-add has tests — one that it inherits, one that it doesn't inherit across products. The CSV path's deliberate non-inheritance is stated in a comment. The `article_number` mapping has its assert, blank and padded included. The `__none__` sentinel is now longer than the column's 64-character cap, so nothing a spool can hold collides with it; I left the category and storage-location chips on their readable `__none__` — same theoretical exposure, but not this branch's business. Indentation fixed, and the query key joined the `inventory_changed` set.

**Backup and restore:** added `material_number`, `category`, `low_stock_threshold_pct` and `storage_location` to both whitelists, which closes the older omissions too. Restore applies them only when the file actually carries them, so an older backup can't wipe what the live row holds. I left `location_id` out: the locations table isn't in the backup, so the ID would point at whatever happens to own it on the target. Say the word and I'll add location backup alongside it.

The design note is a wiki line rather than a code change, so the behaviour is unchanged.
---

While re-reading the branch against your points I found four more and fixed them in the same pass, so the diff you get back is a little wider than the list above:

- `remaining_g` clamped the **group** total rather than each spool, so one over-consumed spool subtracted from its siblings — two spools on number "15" reported 800 g where the inventory page shows 1000 g. Per-spool clamping is the house convention in six other places; now it is the convention here too.
- SpoolBuddy's new-spool form offered a Material No. input it never submitted. Typing a number there silently lost it.
- The Spoolman-mode hide was not gated on the settings query resolving, so the widget flashed up for a moment before disappearing.
- One test named a tie-break it did not actually exercise.

PR #2970, opened 2026-08-27
Easy management of printer locations (groups) (issue #2962)

SVAN19:
## Description

This PR adds comprehensive group management for printer locations. Users can now organize printers into named groups with custom icons and colors, sort groups, perform bulk operations (select/delete/move), and customize the visual appearance of each group.

## Related Issue
[LINK](https://github.com/maziggy/bambuddy/issues/2962)

## Documentation
[PrinterLocationsPage-UserGuide-EN.md](https://github.com/user-attachments/files/31504843/PrinterLocationsPage-UserGuide-EN.md)

**Pick one**:
- [x] Docs PR(s) linked above
- [ ] No docs update required — reason: ___

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

The “Printer Groups” (Printer Locations) page is designed to organize printers by locations and groups. Main features:

Group management:

Creating groups with arbitrary names
Editing a group — changing the name, icon, and color
Deleting a single group (all printers are moved to “No Group”)
Bulk deletion of multiple groups via selection mode
Visual design:

Sorting and filtering:

4 sorting modes: by name A→Z, Z→A, by the number of printers ↑ and ↓
“Without group” always remains regardless of sorting
Hiding/showing empty groups (without printers)
Searching for groups by name
Printer management:
Expanding a group to view the list of printers with their status (online/printing/offline)
Moving an individual printer to another group
Removing a printer from a group (moving it to “Without group”)
Bulk movement of selected printers between groups
Mass deletion of selected groups

Bulk operations:

Group selection mode — the “Select”/“Done” button enables checkboxes on the group cards; the expand arrows are replaced with checkboxes; group expansion is disabled in this mode.
Bulk deletion panel at the bottom of the screen with confirmation and a count of the actual number of affected printers.
Bulk movement panel for selected printers.

## Screenshots

**Screenshots 1** 
<details>
**Entry point**
Image
**Appearance**
Image

</details>
**Screenshots 2**
<details>
**Add new groupp**

Image
**Edit group**

Image
Mass movement of printers by groups

Image

</details>
**Screenshots 3**
<details>
**Mass deletion of groups**
Image
**Moving only 1 printer to the group.**
Image

</details>
**Screenshots 4**
[...]

Issue #2962, opened 2026-08-26
[Feature]: Easy management of printer locations (groups)


SVAN19:
### Problem or Use Case

The current implementation of the control of printer groups does not allow this to be done quickly. Any typo in the name of the group (location) of the printer leads to the creation of a new one. In fact, changing the printer location is the wait for a connection check. It's been a long time.

### Proposed Solution

A simple solution is to manage printer groups (locations) separately. Nothing will change, but a new interface for managing printer locations will be added, allowing you to quickly:
- Create a new location
- Delete unnecessary locations
- Assign printers to locations
- Visually see which printer is in which location

### Feature Category

Other

### Priority

Critical for my use case

### Mockups or Examples

<img width="1908" height="935" alt="Image" src="https://github.com/user-attachments/assets/6ae3cdc4-1634-4b12-91b0-13fb0dfff1f3" />

<img width="1901" height="941" alt="Image" src="https://github.com/user-attachments/assets/9dcda0d4-07db-439e-8f39-89decb5ced30" />

### Contribution

- [x] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
I would prefer a small "Manage Locations" button on the printer's page instead of adding a new sidebar item.

Please note our rules from CONTRIBUTING.md. Thanks!

SVAN19:
I hear you. That's a good option, too. But it won't allow you to create groups (locations) with zero numbers. Example:
a simple move of printers from one rack to a new rack or to a new room. In my implementation, I can create future locations or groups. A new future rack or room. If you don't want to see this as a separate navigation menu item, you can make it a separate option in the settings.
Regarding the implementation on the printer page, I hear you. I'll try to implement it as soon as possible.

SVAN19:
<img width="1908" height="951" alt="Image" src="https://github.com/user-attachments/assets/90572406-063e-450e-acfd-40d50139854b" />

Is it okay if this is located here?

maziggy:
Yes, thanks!

SVAN19:
Hi, [maziggy](https://github.com/maziggy). I’ve finished working on the functionality related to printer groups.

**Screenshots 1** 
<details>
**Entry point**
<img width="1832" height="298" alt="Image" src="https://github.com/user-attachments/assets/e644207b-b4a3-486f-a343-128908f49b95" />
**Appearance**
[...]

PR #2995, opened 2026-08-29
Easy management of printer locations (groups) (issue #2962)- #2970

SVAN19:
## Description

This PR adds comprehensive group management for printer locations. Users can now organize printers into named groups with custom icons and colors, sort groups, perform bulk operations (select/delete/move), and customize the visual appearance of each group.

## Related Issue
[LINK](https://github.com/maziggy/bambuddy/issues/2962)

## Documentation
[PrinterLocationsPage-UserGuide-EN.md](https://github.com/user-attachments/files/31504843/PrinterLocationsPage-UserGuide-EN.md)

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

The “Printer Groups” (Printer Locations) page is designed to organize printers by locations and groups. Main features:

Group management:

Creating groups with arbitrary names
Editing a group — changing the name, icon, and color
Deleting a single group (all printers are moved to “No Group”)
Bulk deletion of multiple groups via selection mode
Visual design:

Sorting and filtering:

4 sorting modes: by name A→Z, Z→A, by the number of printers ↑ and ↓
“Without group” always remains regardless of sorting
Hiding/showing empty groups (without printers)
Searching for groups by name
Printer management:
Expanding a group to view the list of printers with their status (online/printing/offline)
Moving an individual printer to another group
Removing a printer from a group (moving it to “Without group”)
Bulk movement of selected printers between groups
Mass deletion of selected groups

Bulk operations:

Group selection mode — the “Select”/“Done” button enables checkboxes on the group cards; the expand arrows are replaced with checkboxes; group expansion is disabled in this mode.
Bulk deletion panel at the bottom of the screen with confirmation and a count of the actual number of affected printers.
Bulk movement panel for selected printers.

## Screenshots

**Screenshots 1** 
<details>
**Entry point**
Image
**Appearance**
Image

</details>
**Screenshots 2**
<details>
**Add new groupp**

Image
**Edit group**

Image
Mass movement of printers by groups

Image

</details>
**Screenshots 3**
<details>
**Mass deletion of groups**
Image
**Moving only 1 printer to the group.**
Image

</details>
**Screenshots 4**
<details>
Functional panel: sorting by letter, by the number of printers; showing/hiding empty groups.
Image

</details>

## Testing

## Checklist

## Additional Notes

[PrinterLocationsPage-Overview.md](https://github.com/user-attachments/files/31505154/PrinterLocationsPage-Overview.md)


maziggy (changes requested):
Thanks - and you were right to close #2970 and start again. For the record of what went wrong there: the final merge pulled a stale pre-rebase copy of your branch back in, and that resurrected old file contents on top of dev. Relative to the merge base it rolled the app version back from 1.2.6b1 to 1.2.5.3, removed a Bambu Cloud sign-in guard, dropped four CHANGELOG entries and two sponsors from BACKERS.md, and left static/index.html pointing at a bundle filename that was not in the tree, so a deploy from it would have served a blank page. None of that is in #2995. Two habits will keep it from recurring: never merge one copy of your branch into another (rebase onto dev instead), and leave static/ out entirely - it is build output and we regenerate it.

On #2995 itself, the page is a lot of careful work and the Russian translation is genuinely good. Three things block it.

It deletes a key that is still used. printers.dropToQueue is gone from en.ts and ru.ts, replaced by a new cannotPrint that nothing references. PrintersPage.tsx:3857 still calls t('printers.dropToQueue'), so English and Russian users dragging a file onto a busy printer will see the raw key text on screen. I think this came in by accident from another branch.

Parity. I applied your two locale hunks to the other twelve locales and counted: 1020 missing keys and 13 extra. npm run test:run runs check:i18n and will fail. Every new key has to exist in all fourteen locales - English text as a placeholder is fine for the ones you do not speak.

English pluralisation. You used the Russian plural categories (_few, _many) in en.ts, but English only has one and other. I ran it through our own i18next: printer with count 5 renders "5 printer", and group with count 7 renders the bare word "group" with no number at all, because group_one exists but group_other does not. Those are the two most visible labels on the page. English needs _one and _other only; keep _few/_many in ru.ts where they are correct.

Beyond that, four things I would want before merging:

- Bulk delete, rename and bulk move all use Promise.all over one PATCH per printer. If the sixth of ten fails, the group is left split under two names and the toast only says it failed. Promise.allSettled with an "n of m" message would be honest about what happened.
- The mutations read the cached printer list rather than fetching, so a move made by another user in the meantime gets overwritten.
- Printer.location is varchar(100) with no validation on either side, and the name inputs have no maxLength. On PostgreSQL a longer name is a 500 in the middle of a bulk write.
- There are no tests for 1374 lines of page plus the three utils.

One design question, and it is the important one. Groups with no printers exist only in the browser's localStorage, and so do the icons and colours. That is exactly the capability you argued for - creating a future rack or room before the printers move in - but it means the group is invisible on your phone, invisible to every other user, and gone if site data is cleared. For a single-user install it works; for a shared one it will look like data loss. If we keep that capability, locations should become a real table on the backend, which would also make rename one write instead of one per printer. That is a bigger change than this PR, so tell me how you want to handle it before you rework anything else.

Last thing: I asked for a button on the printers page rather than a new sidebar item, and this adds both. Which one did the screenshot I approved show?

PR #2996, opened 2026-08-29
Suppliers as a managed list with per-spool assignments

### LINKED ISSUE #2988: [Feature]: Suppliers as a managed list, with a material assignable to several suppliers
### Problem or Use Case

We run Bambuddy commercially on a 9-printer Bambu Lab farm (H2S / H2D / X1E), and the inventory is also what purchasing looks at.

A spool record today knows *who made* the filament (`brand` — "Bambu Lab", "Polymaker", "Extrudr"), but not *where we buy it*. In practice those are two different things:

- the same product is available from several suppliers at different prices, lead times and minimum quantities — we buy Bambu Lab filament both directly and through resellers,
- one supplier ships many brands, so "who do I order this from" cannot be derived from the brand,
- for some materials there is exactly one distributor, and today that fact lives only in someone's head.

There is nowhere in the spool record to put this:

- `note` is free text, but it is not sortable, not filterable and nothing a report can be grouped by,
- `category` would technically work, but it is already the production / customer grouping field — the same conflict described in #2870, and one of the two use cases would have to lose,
- writing it into `brand` ("Bambu Lab (material4print)") corrupts the field that RFID auto-match, the spool catalog and the brand grouping in statistics rely on.

A plain free-text field would not really solve it either: typed by hand on every roll it drifts immediately ("material4print", "Material4Print", "m4p"), and the interesting case — *this material is available from these three suppliers* — cannot be expressed in a single text field at all.

### Proposed Solution

Model suppliers once, then pick them from a dropdown on the material.

**1. A managed supplier list**

A small master list (Settings → Inventory → Suppliers, or next to the spool catalog):

- name (required), website / shop URL, own customer number, note,
- normal create / edit / delete, with the number of spools using a supplier shown so deleting is a conscious act (block or soft-delete rather than silently orphaning references).

**2. Assign suppliers on the material / spool**

- In the Edit spool dialog, a **multi-select dropdown "Suppliers"**: a material can be sourced from several suppliers, and you tick the ones that apply — shown as chips, no typing, no spelling drift.
- "+ New supplier" inline in that dropdown, so adding one does not mean leaving the dialog.
- Optional per assignment, and the part that makes it useful for purchasing: **supplier article number** (their number — deliberately not the internal material number from #2870) and **price per kg at that supplier**, so the same product can be compared across suppliers.
- Optional: mark which supplier this particular spool was actually bought from, so cost per spool stays exact while the other entries remain as alternative sources.

**3. Use it everywhere the inventory is already used**

- Optional, sortable column in the inventory list plus a filter "show everything sourced from supplier X".
- Grouping key in Statistics, so consumption and cost can be reported per supplier, not only per brand.
- Included in the CSV/Excel export and exposed in the REST API (`/api/v1/inventory/spools`, plus a `/api/v1/suppliers` endpoint for the list itself), so purchasing / ERP can pull it.
- Carried over when a new spool of the same brand / material / subtype / colour is created or auto-matched via RFID — otherwise every newly scanned roll has to be sourced by hand again (same argument as #2870). If #2936 (linked spools) lands, the supplier assignment clearly belongs to the synced master data, not to the per-spool data.

Scope note: even the minimal version — supplier list plus multi-select on the spool — would already solve the main problem. Per-supplier article number and price can follow later.

### Alternatives Considered

- **A plain free-text supplier field** — better than nothing, but it drifts in spelling, offers no dropdown, and cannot express "available from several suppliers".
- **Putting the supplier into `brand`** — breaks RFID auto-match, spool catalog lookups and brand grouping in statistics. That field has a job already.
- **`note`** — free text, not sortable, filterable or groupable, so it cannot drive a reorder list or a cost report.
- **`category`** — already the production / customer grouping field (#2870 hits the same wall).
- **#1754 (Custom fields for filament inventory)** — a generic custom-field system would cover the single-supplier case *if* those fields are sortable, filterable, exportable and usable in statistics. It would not cover several suppliers per material with their own article number and price, which is the actual point here.
- **Spoolman** — Spoolman's `vendor` on the filament object models the *manufacturer*, not the reseller, and it is a single reference, not a list. Running Spoolman alongside Bambuddy for this alone is a lot of overhead.
- **Keeping sourcing in the ERP / a spreadsheet only** — what we do today. It works for ordering, but the spool in Bambuddy still cannot tell you where it came from, and the two lists drift apart.

### Feature Category

Spool Inventory

### Priority

Would improve my workflow

### Mockups or Examples

Edit spool dialog:

```
Brand        [ Bambu Lab                         ]
Suppliers    [ material4print x ] [ Bambu Lab x ] [ + ]     <- new, multi-select
                 |
                 +-- material4print      art. no. [ PLA-BL-10101 ]   32.90 EUR/kg   [x] bought here
                 +-- Bambu Lab (direct)  art. no. [ 10101        ]   34.50 EUR/kg
Cost per kg  [ 32.90 EUR ]
Category     [ Production ]
```

Settings → Inventory → Suppliers:

```
Name              Website                      Customer no.   Spools
material4print    https://material4print.de     12345          47     [edit] [delete]
CR-3D             https://cr-3d.de              -               8     [edit] [delete]
Extrudr           https://extrudr.com           -               5     [edit] [delete]
```

Related: #2870 (internal material number — ours; the supplier article number is theirs), #2936 (linked spools — the supplier assignment would be synced master data), #1754 (generic custom fields).

### Contribution

### Checklist

Thomansky:
﻿## Description

Suppliers as a managed master list with per-spool assignments, in the shape agreed in the issue thread: inventory master data following the Locations pattern, no new permissions, a clearly-named quoted price per assignment, full Spoolman parity, and a fully specified CSV round-trip.

## Related Issue

Fixes #2988

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#82

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- **Data model**: `suppliers` master table plus `spool_suppliers` (surrogate id + UNIQUE(spool_id, supplier_id), per-assignment `supplier_article_number`, `quoted_price_per_kg`, `is_purchase_source`) and the `spoolman_spool_suppliers` twin keyed by the remote spool id with no FK (`SpoolmanKProfile` precedent). Idempotent create-table migration for SQLite and PostgreSQL.
- **Routes**: `/api/v1/inventory/suppliers` CRUD in `inventory.py` next to the location routes, gated by `INVENTORY_READ`/`INVENTORY_UPDATE` — no new permissions. Deleting a referenced supplier returns 409 exactly like `delete_location`; the reference count covers both inventories. Every mutation broadcasts `{"type": "inventory_changed"}`.
- **Assignments**: `PUT /inventory/spools/{id}/suppliers` (replace-all, k-profiles pattern, at most one purchase source) and the mirrored `PUT /spoolman/inventory/spools/{id}/suppliers`; assignments embedded in the spool responses of both inventories.
- **Quoted price**: the per-assignment price is `quoted_price_per_kg` — a quoted price for comparing sources, never a cost basis. `spool.cost_per_kg` is never written from assignments, and supplier statistics aggregate the recorded usage history only.
- **UI**: a **Suppliers** button on the Inventory toolbar (next to Locations) opening a `SuppliersModal`, in both inventory modes; a multi-select with chips + inline "+ New supplier" in the spool dialog, same section in Spoolman mode saving to the twin endpoint; sortable Suppliers column, filter chip, free-text search match.
- **Inheritance**: a new spool of an already-assigned product (NULL-safe brand/material/subtype/colour match) arrives with the source list; the *bought here* marker is never inherited. Covers manual create, bulk create, the API and the RFID auto-add.
- **Statistics**: `GET /inventory/stats/suppliers` + a "By Supplier" dashboard widget, grouped by purchase source.
- **CSV**: two columns — `suppliers` (`;`-joined names) and `purchase_supplier`. Import matches names trimmed and case-insensitively against the existing list and **creates nothing**: an unknown name surfaces as a dry-run warning, the row imports, the unmatched assignment is dropped. Article number and quoted price deliberately stay out of the CSV.
- **i18n**: all strings in `en.ts` and the 13 other locales, `npm run check:i18n` green.

## Screenshots

| Before | After |
| --- | --- |
| before: inventory toolbar without the Suppliers button | after: Suppliers button next to Locations |
| *(no equivalent — modal is new)* | after: the Suppliers master-list modal |
| before: Color & Cost tab without the section | after: Suppliers assignment in the spool dialog |

## Testing

- Backend: 19 tests — supplier CRUD incl. the delete guard (both inventories), replace-all assignments with validation, Spoolman-twin endpoints incl. embedding, inheritance (incl. explicit-wins and different-product cases), statistics grouping, CSV export/import/round-trip, migration add + idempotency.
- Frontend: SuppliersModal, SupplierSection and SpoolFormModal tests; full suite and `tsc -b` green, i18n parity across 14 locales.
- Branch rebased onto current `dev`; broad spool/inventory/permission backend subset green (the only failures on my Windows dev box are the pre-existing platform-specific ones that fail identically on a clean `dev` checkout).

- [ ] I have tested with my printer model: no printer-facing behavior changes; the RFID auto-add path is covered by the inheritance tests

## Checklist

## Additional Notes

- `supplier.name` is what the CSV matches and the delete guard reports; renaming a supplier re-points future CSV imports that used the old name (same exposure `storage_location` already has), as flagged in the issue thread.
- Out of scope, per the issue: ordering, live shop prices, supplier ratings.


maziggy:
I don't have any problems with using AI for assistance - but to be honest - I don't want to have AI commits in rhe repo. Please remove. Thanks.

Thomansky:
Done — sorry about that. The trailers were only there for transparency's sake, but I completely understand your point. I've rewritten the commits on all my open PR branches (code and wiki) without them and dropped the footer from the PR descriptions; the content of the commits is unchanged. Won't happen again.

Thomansky:
No pressure at all — just flagging that this one is still mergeable and rebased on current `dev`, in case it dropped off the list.

If there is anything that would make it quicker to look at — splitting it up, dropping a part you are unsure about, or me writing a short "what to check" list — say the word and I will do that instead of adding anything new to it. I know the review queue is the scarce resource here, not the code.

maziggy (changes requested):
Reviewed in full against current dev. I checked the code against each of the seven
points we agreed in the issue thread, and all seven are honoured: the Locations
pattern for the modal and the routes, the 409 delete guard counting both inventories,
full scope in one PR, quoted_price_per_kg kept out of cost statistics with
spool.cost_per_kg never written from an assignment, the Spoolman twin on the
SpoolmanKProfile precedent with mirrored routes and the same UI in both modes, no new
permissions plus the inventory_changed broadcast, and the CSV done properly rather than
half-done. Thank you for the rework -- that is a lot of ground.

Verified on the branch: ruff clean, backend 12242 passed, vitest 3685 passed, eslint
and build clean, i18n parity green across all 14 locales. I also dry-ran the migration
against a real PostgreSQL instance inside a rolled-back transaction: all three tables
and their indexes create, a second run is a silent no-op, and the foreign key backstops
the delete guard at the database level.

Six things to fix before this can go in.

1. /spools/from-slot returns an empty supplier list for a spool that just inherited
   one. I reproduced the sequence against the real models: create_spool_from_tray sets
   spool.supplier_links = [] before the flush, apply_supplier_inheritance then adds rows
   with db.add(SpoolSupplier(...)) rather than through the relationship, and the route's
   closing select(Spool).where(id == ...) returns the same identity-mapped instance --
   SQLAlchemy does not overwrite an already-loaded collection. Result: one row in the
   table, [] in the response, and a fresh session sees the row. So "+ Add to inventory"
   shows no suppliers until the next fetch. Append through spool.supplier_links, or use
   populate_existing on that query. Related: the PR checkbox says the RFID auto-add path
   is covered by the inheritance tests -- it is not, those go through POST /spools, which
   re-queries and therefore cannot catch this.

2. Supplier names are not unique, and the whole feature keys on them. There is no unique
   constraint on suppliers.name and no duplicate check in create_supplier or
   update_supplier -- I inserted two rows named "Extrudr" on Postgres to confirm.
   create_location, the route this mirrors, returns 409 DUPLICATE_LOCATION_NAME and
   catches the IntegrityError behind it. Your own plan said "name (unique, 100)", and
   your note about renaming re-pointing CSV imports assumed it. With duplicates allowed,
   _load_supplier_map resolves an imported name to whichever row it happened to build
   last, which is silently the wrong supplier.

3. lazy="selectin" on Spool.supplier_links changes loading for the whole application.
   It is the only relationship on Spool with a non-default loader -- every other one is
   loaded explicitly with selectinload() at the query site. There are around fifty
   select(Spool) call sites across the usage tracker, AMS sync, labels, backup and the
   printer routes, and each now pays two extra round trips, since SpoolSupplier.supplier
   is selectin as well. Please use selectinload(Spool.supplier_links) at the few places
   that actually embed them.

4. Deleting a Spoolman spool orphans the twin rows, and the orphan blocks the supplier
   permanently. The Spoolman delete_spool route cleans up nothing local. That is
   pre-existing for SpoolmanKProfile and SpoolmanFilamentPreset, but those leaks are
   inert -- a leaked spoolman_spool_suppliers row keeps _supplier_reference_counts
   non-zero, so the supplier answers 409 forever and the user has no way to find the
   phantom reference. Delete the twin rows in that route.

5. The supplier list is not in react-query, so the broadcast is not reaching anything.
   SuppliersModal and SupplierSection each keep a private useState copy fetched on mount,
   and useWebSocket invalidates inventory-spools, spoolman-inventory-spools and
   inventory-locations -- there is no supplier key anywhere. A supplier created inline in
   the spool dialog therefore does not appear in the modal, and a second tab or user never
   sees it. LocationsModal, the pattern this follows elsewhere, uses
   inventoryLocationsQueryKey for exactly this.

6. The statistics widget repeats both problems from #2994: it ignores the dashboard
   timeframe that every other widget on that page honours, and isError renders the
   "no suppliers yet" text, so a 403 or a 500 reads as "you have not assigned any".
   SupplierStats.tsx is also a near-copy of MaterialNumberStats.tsx -- same formatGrams,
   same table, same three states. If both PRs land, please make that one component.

Smaller:

- SettingsPage.tsx: the only line this PR touches in that file changes a comment to read
  "Spool Catalog + Color Catalog + Suppliers", but nothing supplier-related renders there.
  Leftover from the pre-agreement design.
- Same leftover in i18n: the toolbar button and the whole modal use settings.suppliers.*,
  while the dialog and inventory strings use inventory.suppliers.*, for a feature that is
  deliberately not in Settings.
- In SpoolFormModal, saveSupplierLinks was inserted between savePrinterProfiles' docstring
  comment and the function, so that comment now describes the wrong function and
  savePrinterProfiles has none.
- Bulk create calls apply_supplier_inheritance once per copy, three queries each, for an
  answer the comment itself says is identical for all copies. Resolve the templates once.
- SupplierUpdate.name is str | None, so an explicit {"name": null} passes validation and
  then hits NOT NULL -- a 500 where it should be a 422.
- A supplier name containing a semicolon breaks the CSV round-trip silently: it splits
  into unknown names and the assignments are dropped. Reject the separator in the name,
  or escape it.
- The unknown-supplier warning is emitted per row, so a 500-row CSV against an empty
  supplier list produces 500+ warning lines in the dry run. Dedupe per name.
- SupplierSection swallows a failed getSuppliers() into console.error and renders an
  empty list, which invites the user to create a duplicate.
- Missing tests: bulk-create inheritance, a spool delete cascading its links (and the
  supplier becoming deletable again), and the Spoolman twin rows after the remote spool
  is gone.

Thomansky:
PR REPLY (owner's voice) ===

Thanks for going through it that carefully — the Postgres dry-run and the reproduction on point 1 in particular saved me a lot of guessing. All six are fixed, plus everything under "Smaller". Four commits on the branch.

**1. /spools/from-slot returning an empty supplier list.** You had it exactly right, and I reproduced it before changing anything: the row is written, a fresh session sees it, the response says `[]`. `create_spool_from_tray` sets `spool.supplier_links = []` before the flush so the cascade can't lazy-load it, and the closing `select` hands back that same identity-mapped instance. That query now runs with `populate_existing`. You were also right that the path had no coverage — the inheritance tests go through `POST /spools`, which builds a fresh instance and can't reach it. `TestFromSlotInheritance::test_from_slot_response_carries_inherited_suppliers` drives the slot endpoint and asserts both the response and the row; on the old code it fails with `[] != ['Supplier A']`.

**2. Supplier names not unique.** Fixed, and case-insensitively — the CSV map is keyed on the lower-cased name, so `extrudr` would have been just as ambiguous as a second `Extrudr`. `suppliers` now carries a unique index on `lower(name)`, declared on the model so `create_all` covers fresh installs and spelled out in `_migrate_create_supplier_tables` for upgrades, and `create_supplier`/`update_supplier` check first and catch the `IntegrityError` behind it the way `create_location` does. Pinned by `TestSupplierNameUniqueness` (duplicate, case variant, rename collision, trimming) and by `test_migration_enforces_case_insensitive_unique_names` on an upgraded database.

**3. lazy="selectin" on Spool.supplier_links.** Gone, along with the same on `SpoolSupplier.supplier` and `SpoolmanSpoolSupplier.supplier`. The routes that actually embed the links load them through a small `spool_response_loads()` helper, next to the `selectinload(Spool.k_profiles)` they already carried — every route answering with a `SpoolResponse`, the two assignment routes that nest one, and the CSV export. Everything else that selects a `Spool` pays nothing now. `test_supplier_relationships_use_the_default_loader` pins that none of the three picks up a non-default loader again.

**4. Orphaned Spoolman twin rows.** One helper now purges `SpoolmanKProfile`, `SpoolmanFilamentPreset` and `SpoolmanSpoolSupplier` on both the single and the bulk delete, so the two older leaks go with it. Covered by `test_deleting_a_spoolman_spool_drops_the_twin_rows` and `test_bulk_delete_drops_the_twin_rows_too`; both fail without the purge, and both assert the supplier becomes deletable again afterwards.

**5. Supplier list not in react-query.** `SuppliersModal` and `SupplierSection` both read it through `inventorySuppliersQueryKey` now, `useWebSocket` invalidates that key on `inventory_changed`, and the write paths invalidate instead of patching a private copy. The modal test pins that a supplier created elsewhere shows up once the key is invalidated and that the create path re-reads rather than pushing into local state; the `useWebSocket` test pins the key itself.

**6. Statistics widget.** `GET /inventory/stats/suppliers` takes `date_from`/`date_to`, scoping the usage half only — stock stays point-in-time, as you said — and the range is part of the query key. `isError` has its own message instead of the empty-state text. `SupplierStats.test.tsx` covers the timeframe reaching the endpoint, the refetch when it changes, and error vs genuinely empty. On folding it together with `MaterialNumberStats`: agreed, but I'd rather do that once #2994 has landed than guess at its final shape from here.

**Smaller.** All done. The `SettingsPage.tsx` comment is back to what it was; the master-list strings moved from `settings.suppliers.*` to `inventory.suppliers.*` so the whole feature sits in one namespace; the `savePrinterProfiles` comment is back on `savePrinterProfiles`; the bulk create resolves the inheritance templates once for the batch instead of three queries per copy; `{"name": null}` is a 422 instead of a 500; a `;` in the name is refused rather than silently breaking the CSV round-trip; the unknown-supplier warning is emitted once per name; and a failed `getSuppliers()` says so in both places instead of rendering an empty list. The three missing tests are there too — bulk-create inheritance, a spool delete freeing its supplier, and the twin rows after the remote spool is gone.

And you were right about the PR checkbox claiming the RFID auto-add path was covered by the inheritance tests — it wasn't. I'll un-tick it; the new slot test covers it now.

=== KEY FILES ===
C:/Users/thoma/Documents/Coding/Bambubaddy/wt-sup/backend/app/api/routes/inventory.py (spool_response_loads helper, DUPLICATE_SUPPLIER_NAME + _supplier_by_name, populate_existing on from-slot, date window on /stats/suppliers)
C:/Users/thoma/Documents/Coding/Bambubaddy/wt-sup/backend/app/api/routes/spoolman_inventory.py (_purge_local_rows_for_spool, explicit supplier loaders)
C:/Users/thoma/Documents/Coding/Bambubaddy/wt-sup/backend/app/models/supplier.py (unique index on lower(name), loaders removed)
C:/Users/thoma/Documents/Coding/Bambubaddy/wt-sup/backend/app/models/spool.py (lazy="selectin" removed)
C:/Users/thoma/Documents/Coding/Bambubaddy/wt-sup/backend/app/schemas/supplier.py (validate_supplier_name)
C:/Users/thoma/Documents/Coding/Bambubaddy/wt-sup/backend/app/services/supplier_links.py (apply_supplier_inheritance_to_batch)
C:/Users/thoma/Documents/Coding/Bambubaddy/wt-sup/backend/app/services/spool_csv.py (warning dedupe)
C:/Users/thoma/Documents/Coding/Bambubaddy/wt-sup/backend/app/core/database.py (unique-index migration)
C:/Users/thoma/Documents/Coding/Bambubaddy/wt-sup/frontend/src/utils/inventoryQueries.ts (inventorySuppliersQueryKey)
C:/Users/thoma/Documents/Coding/Bambubaddy/wt-sup/frontend/src/components/SuppliersModal.tsx, .../spool-form/SupplierSection.tsx, .../SupplierStats.tsx
C:/Users/thoma/Documents/Coding/Bambubaddy/wt-sup/frontend/src/__tests__/components/SupplierStats.test.tsx (new)
---

One more, and it is the reason I am glad you inserted those two "Extrudr" rows rather than only describing the problem: **the unique-index migration aborted startup on a database that already holds duplicates.** `_safe_execute` catches `OperationalError` and `ProgrammingError`, and a unique-index build on colliding rows raises `IntegrityError`, which propagates straight out of `run_migrations`. Any install that had created two suppliers with the same name — which yours now has — would have failed to start after the upgrade. Colliding rows are merged before the index is built: oldest row wins, takes over both assignment tables, fills its own empty fields from the rows it absorbs.

Three others from the same re-read: the case-insensitive uniqueness was ASCII-only on SQLite, so the CSV ambiguity you describe survived for non-ASCII names; copy mode displayed supplier assignments it never saved; and the twin-row purge only ran when Bambuddy itself performed the delete, so a 409 could still be permanent if the row went away another way. One test could pass vacuously and now cannot.

PR #2998, opened 2026-08-29
Improve spool ID search and AMS assignment (issue #2978)

### LINKED ISSUE #2978: [Feature]: Improve spool ID search and allow direct printer/AMS assignment from the spool view
### Problem or Use Case

Problem and motivation

Larger print farms often have several visually identical filament spools. The sequential Bambuddy spool ID and its QR code are therefore the only unambiguous link to the physical spool.

In the existing workflow, the spool ID was not visible in the assignment list and could not be searched exactly. Searching for “3”, for example, could also return spools 13 or 30. In addition, a spool opened through its QR code could not be assigned directly to a printer or AMS slot. The assignment had to be started from the printer instead.

Useful improvements:

Display the spool ID as #ID
Support exact searches such as #3 while preserving the existing general search
Assign a spool directly from its edit and QR-code view
Select the printer, AMS or external slot, and spool colour in one dialog
Require explicit confirmation instead of assigning immediately after a slot click
Provide a responsive and more accessible dialog with translations
Automatically follow Bambuddy’s configured light or dark theme

This creates a shorter and less error-prone workflow from the physical spool’s QR code to the correct AMS slot, especially for users managing many identical spools.

### Proposed Solution

Test installer for v1.2.6b1 beta

I can also provide a self-contained patch installer for the original Bambuddy v1.2.6b1 beta build (commit 14f4651) if the maintainers or other users would like to test the proposed workflow.

The installer verifies the exact source version, applies the complete patch, rebuilds the frontend, creates a backup of the current installation, and automatically rolls back the changes if the build or service restart fails.

It is intentionally version-specific and intended only for testing on the original v1.2.6b1 beta, not as a general installer for current or future Bambuddy releases.

[bambuddy-v1.2.6b1-spool-enhancements-complete.patch](https://github.com/user-attachments/files/31527089/bambuddy-v1.2.6b1-spool-enhancements-complete.patch)
[install-bambuddy-v1.2.6b1-spool-enhancements-complete.sh](https://github.com/user-attachments/files/31527088/install-bambuddy-v1.2.6b1-spool-enhancements-complete.sh)

### Alternatives Considered

_No response_

### Feature Category

Spool Inventory

### Priority

Nice to have

### Mockups or Examples

_No response_

### Contribution

### Checklist

pd81:
## Description

Improves spool identification and shortens the workflow from a physical spool's QR code to its printer or AMS slot.

## Related Issue

Fixes #2978

## Documentation

**Companion docs PRs**:
- Wiki: https://github.com/maziggy/bambuddy-wiki/pull/83

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Breaking change
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition or update

## Changes Made

- Support exact spool ID searches such as `#3` while preserving partial numeric searches.
- Display `#ID` in local-inventory and Spoolman assignment results.
- Allow assignment directly from the spool editor and QR-code view.
- Add printer, AMS/external-slot, and optional colour selection to one modal.
- Require explicit confirmation before changing the assignment.
- Respect Bambuddy's configured light/dark theme and accent colours.
- Improve responsive layout, keyboard operation, and modal accessibility.
- Correctly identify `Ext-L` and `Ext-R` on dual-nozzle printers.

## Screenshots

Screenshots are attached for:
- Exact `#ID` search
- Assignment modal in light mode
- Assignment modal in dark mode / narrow layout

## Testing

Additional checks completed:

- 76 affected frontend tests passed
- TypeScript compilation passed
- ESLint passed
- Production build passed
- Safari 16 compatibility baseline passed
- Translation parity passed for all 14 locales
- Light/dark and 390×844 responsive modal rendering verified
- Strict companion Wiki build passed

## Checklist

## Additional Notes

The implementation ports the tested v1.2.6b1 beta workflow to the current `dev` branch and includes an additional fix for left/right external-slot labeling on dual-nozzle printers.

Bambuddy_Assign_Spool_3
Bambuddy_Assign_Spool_2
Bambuddy_Assign_Spool_1
Bambuddy_ID_Search


maziggy:

maziggy (changes requested):
Thanks for this -- reviewed in full against current dev.

The #ID search is clean work: it trims, it keeps plain numeric searches as partial
matches, the comment says why, and it is tested three ways. The Ext-L/Ext-R fix is
correct and I checked it against the surrounding code: singleSlots builds external
entries as {amsId: 255, trayId: 0 or 1}, but getSlotLocationLabel only special-cased
amsId === 254, so on a dual-nozzle printer both external slots were labelled "Ext-R"
in the confirm dialog and the footer. Your mapping matches both singleSlots and
getTrayForSlot and keeps 254 for older assignment records. The theming claim holds too
-- the dark: variants match the app's darkMode: 'class' plus ThemeContext, and
--border-color and bambu-gray-light are real theme tokens. Resetting
cachedAmsExtruderMap when the printer changes is a genuine fix the new picker needed.

Verified on the branch: i18n parity green across all 14 locales, eslint clean, tsc -b
clean, vitest 3686 passed, production build clean. All 29 t() keys the diff uses resolve
in en.ts, and nothing was hardcoded.

Four things to fix.

1. Assigning a spool silently rewrites its colour record. I ran this against your own
   test harness -- three spools, assigned without ever touching the colour panel:

     color_name: null,    rgba 00AE42FF  ->  updateSpool(42, {rgba:'00AE42FF', color_name:'Bambu Green'})
     color_name: 'A1-B2', rgba FF00AAFF  ->  updateSpool(42, {rgba:'FF00AAFF', color_name: null})
     color_name: 'Firetruck'             ->  no write (control, correct)

   selectedColorName is seeded from resolveSpoolColorName(), which synthesizes a name
   from the colour catalog when none is stored and returns null for a Bambu code such as
   A1-B2. The "did the user change it" guard then compares that display value against the
   raw spool.color_name ?? ''. So the second case erases the stored code, and the first
   turns a display-time synthesis into stored data. Your tests use color_name: 'Red',
   which is the one shape that does not trigger it. Seed the field from
   spool.color_name ?? '' and show the resolved name as a placeholder, or track a touched
   flag -- SpoolFormModal already does exactly that for weightTouched and
   locationIdTouched.

2. The colour write is not rolled back when the assignment fails. configureMutation
   patches the spool first and assigns second, so a failed assign -- printer offline,
   slot busy, a 409 -- leaves the colour changed while the modal reports only an
   assignment error. Assign first and update after, or make it clear in the UI that the
   colour is saved immediately.

3. The custom colour picker stores the hex as the colour name. setSelectedColorName('#' +
   customHex) writes color_name: "#FF00AA", and resolveSpoolColorName treats that as a
   user-set name, so it shows in the inventory from then on. getColorName(hex) /
   hexToColorName(hex) already exist for this, or leave the name alone when only the
   swatch changes.

4. Please drop the built frontend from the PR. static/assets/index-*.js is about 10 MB,
   plus the CSS and static/index.html. It cannot be reviewed, and it conflicts with dev on
   every commit that rebuilds it. It is also not reproducible: building this exact branch
   here produces index-CJo626YE.js rather than the committed index-7qJ7LJSX.js. The two
   are 40 bytes apart and the CSS hash matches exactly -- the difference is 20 extra \r
   escapes inside className template literals, from a Windows checkout with CRLF endings
   (every source file in the repo is LF). Harmless at runtime, since a carriage return is
   ASCII whitespace and the classes still apply, but it shows the artifact belongs to the
   build step rather than the PR. Source changes only, please.

Smaller:

- There is no test for the case that is actually broken: assigning without touching the
  colour. "persists a changed spool colour" only covers the changed path. A test for the
  failure path leaving the colour changed would be worth having too.
- The new role="button", tabIndex and Enter/Space handling on SpoolSlot is real progress,
  but the only accessible name is a hardcoded English title={`AMS Slot ${n}`}. The new
  tests select on getAllByTitle(/AMS Slot/i), so translating it would break them -- worth
  doing properly with an i18n aria-label and a selector that does not depend on English.
- The #ID search now excludes spools whose note or colour name contains "#3". That reads
  intentional, just confirming it is.

PR #3011, opened 2026-08-31
Let a spool keep its own empty weight (issue #2908)

### LINKED ISSUE #2908: [Bug]: Per-spool core weight is silently discarded in Spoolman mode — the edit field renders, accepts input, and never persists

ojimpo:
## Description

`core_weight` was declared on both Spoolman write schemas and dropped after validation, with a
comment saying so. Spoolman has the field — `spool.spool_weight`, its tare — and Bambuddy already
reads it with priority over the filament-level value. Only the write was missing.

Nothing in the UI showed the edit was ignored, because the read path derives the displayed value
from `spool.spool_weight ?? filament.spool_weight ?? 250`: an edit that went nowhere came back as
the inherited number and read as "no change".

It is not cosmetic. The weigh endpoint resolves the tare the same way and subtracts it from the
measured gross weight, so a spool whose real empty weight differs from its filament's recorded a
wrong remaining weight on every weigh-in — 70 g for my third-party 180 g spools against Bambu's
250 g reusable ones.

## Related Issue

Fixes #2908

## Documentation

**Pick one**:
- [x] No docs update required — reason: the wiki already documents this field as working
  (`docs/features/inventory.md:174`, "**Empty Spool Weight** | Select from the spool catalog or
  enter manually (for accurate remaining calculations)"). This makes the behaviour match what is
  already written rather than changing it.

## Type of Change

## Changes Made

- `SpoolmanClient.create_spool` takes `spool_weight` and sends it when given. Guarded on
  `is not None`, not truthiness: 0 g is a real tare (a bare coil) and is not the same answer as
  omitting the key, which is what tells Spoolman to inherit.
- The single-create handler passes `core_weight` through, but only when the request actually set
  it. The schema defaults it to 250, so writing it unconditionally would stamp an explicit tare on
  every spool created without one and silently detach it from its filament — a worse bug than this
  one, and invisible, since the displayed number would not change.
- `bulk_create_spools` takes the same schema and dropped the field the same way, so it gets the
  same treatment.
- The update handler passes `data.core_weight` straight through. No `model_fields_set` guard here:
  that schema already defaults to `None`, and `None` is what `update_spool_full` reads as "leave
  the tare alone", so a guard would be a second spelling of the same condition. I measured this
  rather than assuming it — see below.
- Both schema comments replaced with what the field now does, and why the create side needs the
  guard while the update side does not.

## Testing

- [ ] I have tested with my printer model: not printer-related — this is the Spoolman write path.

Ten tests, seven at the route level and three on the client. `backend/tests/unit`:
**8613 passed, 22 skipped**. The two files this touches, unit and integration together: **217
passed**. `ruff check` and `ruff format --check` clean.

Mutations, run against those two files:

```
head                                                     217 passed
defect restored (write removed from all three handlers)    6 failed, 211 passed
create guard removed (writes the 250 default always)       2 failed, 215 passed
create_spool guard changed to truthiness                   1 failed, 216 passed
```

Each mutation fails only the tests that exist for it. The second one is worth
reading as a pair with the fourth bullet above: removing the guard from the
*create* handlers breaks two tests, and there is nothing to remove on the update
handler, which is how I know a guard there would have been decoration.

The route tests cover an edited tare reaching `spool_weight`, an edit that never mentions it
leaving the spool inheriting, the same pair on create, bulk create, `0` as a value rather than an
absence, and the weigh endpoint subtracting the tare that was just written — with a 180 g spool a
800 g gross weight records 620 g remaining, where inheriting 250 g would have recorded 550 g.

## Checklist

## Additional Notes

**Why this survived.** There were already three tests naming `core_weight`
(`test_create_spool_with_non_default_core_weight_accepted`,
`test_update_spool_with_non_default_core_weight_accepted`, and the no-`core_weight` case). They
assert status codes only — they were written for an earlier fix that stopped the route rejecting
values other than 250, and "accepted" is all they check. They pass identically with the write
present or absent. Every new test here asserts what reached Spoolman.

**One thing I left out.** There is no way to go back to inheriting through this route: omitting the
field means "leave alone", and the schema has no null-to-clear. `update_spool_full` already has
`clear_spool_weight`, and the filament-level route uses it, so the mechanism exists if you want the
per-spool editor to offer it — but it needs a UI affordance to mean anything, and I did not want to
invent one inside a bug fix. Happy to follow up if you want it.

**A correction on the issue text, already posted there.** The reproduction steps in #2908 are
written as if I executed them; I did not. What I verified was the source of the running build and
the Spoolman side (that `PATCH /api/v1/spool/<id> {"spool_weight": 180}` is picked up immediately
by Bambuddy's tare resolution). The UI round-trip was inferred from the handlers never referencing
`core_weight` after validation. The code-level claim is the part I stand behind, and it is
checkable without the UI.

### DIFF
--- a/backend/app/api/routes/spoolman_inventory.py
+++ b/backend/app/api/routes/spoolman_inventory.py
@@ -310,9 +310,12 @@ class SpoolmanInventoryCreate(BaseModel):
     color_name: str | None = Field(None, max_length=64)
     rgba: str | None = Field(None, max_length=8, description="6-digit hex (RRGGBB) or 8-digit (RRGGBBAA)")
     label_weight: int = Field(1000, ge=1, le=100_000)
-    core_weight: int = Field(
-        250, ge=0, le=10_000
-    )  # Accepted for schema parity but not persisted to Spoolman (stored on filament type, not spool)
+    # Persisted to the Spoolman spool's own `spool_weight` (tare), which takes
+    # priority over the filament-level value both in _map_spoolman_spool and in
+    # the weigh endpoint. Only written when the request actually sets it: the
+    # 250 default is the display fallback, and writing it on every create would
+    # stamp an explicit tare on spools that should keep inheriting one (#2908).
+    core_weight: int = Field(250, ge=0, le=10_000)
     weight_used: float = Field(0.0, ge=0.0, le=100_000.0)
     note: str | None = Field(None, max_length=1000)
     cost_per_kg: float | None = Field(None, ge=0.0, le=1_000_000.0)
@@ -351,9 +354,11 @@ class SpoolmanInventoryUpdate(BaseModel):
     color_name: str | None = Field(None, max_length=64)
     rgba: str | None = Field(None, max_length=8, description="6-digit hex (RRGGBB) or 8-digit (RRGGBBAA)")
     label_weight: int | None = Field(None, ge=1, le=100_000)
-    core_weight: int | None = Field(
-        None, ge=0, le=10_000
-    )  # Accepted for schema parity but not persisted to Spoolman (stored on filament type, not spool)
+    # Persisted to the spool's own `spool_weight` (see the Create schema).
+    # Omitted / null leaves the current value alone, as with every other field
+    # here. There is no per-spool "go back to inheriting" through this route;
+    # the filament-level route already owns that decision (#2908).
+    core_weight: int | None = Field(None, ge=0, le=10_000)
     weight_used: float | None = Field(None, ge=0.0, le=100_000.0)
     note: str | None = Field(None, max_length=1000)
     cost_per_kg: float | None = Field(None, ge=0.0, le=1_000_000.0)
@@ -551,6 +556,7 @@ async def create_spool(
                 remaining_weight=remaining,
                 comment=data.note or None,
                 location=storage_location or None,
+                spool_weight=(data.core_weight if "core_weight" in data.model_fields_set else None),
             )
     except HTTPException as exc:
         if exc.status_code == 404 and data.spoolman_filament_id is not None:
@@ -640,6 +646,7 @@ async def bulk_create_spools(
                 remaining_weight=remaining,
                 comment=data.note or None,
                 location=storage_location or None,
+                spool_weight=(data.core_weight if "core_weight" in data.model_fields_set else None),
             )
         except (SpoolmanUnavailableError, SpoolmanClientError, SpoolmanNotFoundError) as exc:
             logger.warning("Bulk spool creation: one spool failed: %s", exc)
@@ -853,6 +860,11 @@ async def update_spool(
                 extra=extra,
                 location=storage_location or None,
                 clear_location=storage_location_changed and not storage_location,
+                # No model_fields_set guard here, unlike create: this schema
+                # already defaults core_weight to None, and None is what
+                # update_spool_full reads as "leave the tare alone". A guard
+                # would be a second spelling of the same condition.
+                spool_weight=data.core_weight,
             )
 
     # Persist BambuStudio slicer preset AND color_name under spool.extra.
--- a/backend/app/services/spoolman.py
+++ b/backend/app/services/spoolman.py
@@ -405,6 +405,7 @@ class SpoolmanClient:
         lot_nr: str | None = None,
         comment: str | None = None,
         extra: dict | None = None,
+        spool_weight: float | None = None,
     ) -> dict:
         """Create a new spool in Spoolman."""
         data: dict = {"filament_id": filament_id}
@@ -416,6 +417,11 @@ class SpoolmanClient:
             data["lot_nr"] = lot_nr
         if comment:
             data["comment"] = comment
+        # `is not None`, not truthiness: 0 g is a legitimate tare (a spool-less
+        # coil), and it is not the same answer as "inherit from the filament",
+        # which is what leaving the field off means to Spoolman.
+        if spool_weight is not None:
+            data["spool_weight"] = spool_weight
         if extra:
             data["extra"] = extra
             await self._ensure_extra_fields(extra)
--- a/backend/tests/integration/test_spoolman_inventory_api.py
+++ b/backend/tests/integration/test_spoolman_inventory_api.py
@@ -2813,3 +2813,129 @@ class TestGetAllSlotAssignmentsEnriched:
         assert data[0]["printer_id"] == 1
         assert data[0]["printer_name"] == "P1"
         assert data[0]["spoolman_spool_id"] == 201
+
+
+class TestPerSpoolCoreWeight:
+    """The per-spool tare reaches Spoolman now (#2908).
+
+    `core_weight` was declared on both write schemas and dropped after
+    validation, with a comment saying so. The read path never showed it: it
+    derives the value from ``spool.spool_weight ?? filament.spool_weight ?? 250``
+    (_spoolman_helpers.py), so an edit that went nowhere came back as the
+    inherited value and looked like it had simply not changed.
+
+    It is not cosmetic, because the same resolution is the tare the weigh
+    endpoint subtracts. A spool whose real empty weight differs from its
+    filament's produced a wrong remaining weight on every weigh-in -- 70 g for
+    the reporter's third-party spools against Bambu's 250 g reusable ones.
+
+    Spoolman already has the field and already gives it priority. Only the
+    write was missing.
+    """
+
+    @pytest.mark.asyncio
+    @pytest.mark.integration
+    async def test_an_edited_tare_is_written_to_the_spools_own_field(
+        self, async_client: AsyncClient, spoolman_settings, mock_spoolman_client
+    ):
+        response = await async_client.patch("/api/v1/spoolman/inventory/spools/42", json={"core_weight": 180})
+
+        assert response.status_code == 200
+        assert mock_spoolman_client.update_spool_full.call_args.kwargs["spool_weight"] == 180
+
+    @pytest.mark.asyncio
+    @pytest.mark.integration
+    async def test_an_edit_that_does_not_mention_the_tare_leaves_it_inheriting(
+        self, async_client: AsyncClient, spoolman_settings, mock_spoolman_client
+    ):
+        """The reason this keys off model_fields_set rather than the value.
+
+        `core_weight` carries a default, so a PATCH that never mentions it still
+        arrives at the handler holding one. Writing that would stamp an explicit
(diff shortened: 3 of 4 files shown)


maziggy (changes requested):
The backend half of this is right, and the write-up is the most careful I have had on a contributor PR - the 0-versus-absent distinction, the model_fields_set reasoning, the mutation table, and the correction about the repro steps being inferred rather than executed. Verified here on 45af1a94: ruff clean, format clean, 12170 backend tests pass. I checked every claim in the description against the code rather than taking it on trust; they all hold.

Two things to settle, and the first one has to be fixed before this can land.

The kiosk sends the field, so the guard does not protect it. SpoolBuddyDashboard.tsx:425 creates a spool from an unknown tag with a hardcoded core_weight: 250 posted straight to createSpoolmanInventorySpool, and SpoolBuddyWriteTagPage.tsx:747 sends formData.core_weight, which defaults to 250. An explicitly-sent 250 lands in model_fields_set - I measured it - so after this change every spool the kiosk creates in Spoolman mode gets an explicit tare stamped on it and silently stops inheriting. That is the exact bug your fourth bullet is about, reaching the handler through a caller the guard cannot see. Today those values are dropped harmlessly. The fix is the idiom SpoolFormModal:867 already uses: strip core_weight in Spoolman mode unless the user actually set it.

The repro in #2908 is a different screen from the one you fixed. SpoolFormModal passes spoolmanMode into AdditionalSection, which at line 215 swaps the Empty Spool Weight picker for the note "managed per filament type in Spoolman", and handleSubmit strips the field from the payload in that mode. So Inventory - Edit - change - save cannot happen: there is no input and nothing is sent, and this PR does not change that path. The field you saw is on SpoolBuddyWriteTagPage, which renders AdditionalSection without spoolmanMode and therefore shows the picker even in Spoolman mode - that is the screen where your description is literally true, and the one this fixes. Good find, incidentally: the two forms disagree with each other about whether a per-spool tare exists.

Which way that goes is my call, not a defect in your patch, so I will make it: the main form should show the picker in Spoolman mode and send the value only when the user touched it, the same way it already handles weight_used and location_id. Spoolman has the field, our read path already prefers it, and the filament-level editor already manipulates it per spool - the note claiming otherwise is the thing that is wrong. If you would rather keep this PR to the backend I will do the frontend separately; if you want to carry it, take it.

Small one: test_spoolman_inventory_api.py:2201, :2208 and :2743 still carry docstrings saying the field is accepted for schema parity and ignored server-side. They pass either way - which is how you found them - but they now describe the opposite of what happens.

Nothing else. The is not None guard in create_spool, the decision not to add a redundant guard on the update handler, and leaving null-to-clear alone rather than inventing a UI affordance for it inside a bug fix are all the right calls, and the tests assert what reached Spoolman rather than a status code, which is what the three tests that let this through did not do.

ojimpo:
Both of the ones that are mine are done. The third is yours.

**The kiosk no longer sends a tare in Spoolman mode.** You were right that the
guard cannot see those callers: an explicitly sent 250 is in `model_fields_set`,
so every spool the kiosk created would have been stamped with the form default
and stopped inheriting. The two screens needed different fixes.

`SpoolBuddyDashboard.tsx` quick-add has no picker, so it drops the field
outright. The `createSpool` branch a few lines below still sends it, which is
correct — that path owns the tare itself.

`SpoolBuddyWriteTagPage.tsx` renders `AdditionalSection` without `spoolmanMode`,
so the picker is on screen and a value someone actually chose should survive. It
now tracks whether the picker was touched and sends the field only then, which is
the `weight_used` / `location_id` rule from `SpoolFormModal:867`. I track
`core_weight` alone and not `core_weight_catalog_id`: `SpoolWeightPicker`
auto-selects a catalog row on mount when exactly one entry matches the weight, so
the catalog id moves without anyone touching it and every form would have looked
touched.

**The three docstrings now say what the tests do** — :2201, :2208 and :2743. I
left :2755 alone, since it only claims the field is no longer rejected, which is
still what it asserts.

**The main form is yours.** I have no disagreement with the call, and showing the
picker with a touched-check is the same rule I just applied on the kiosk side.
But it is a UI decision in your app rather than a defect in this patch, and I
would rather not hold the fix behind me writing it — please take the frontend
separately.

Frontend `npm run lint` and `npm run build` are clean. I did not re-run the
backend suite: the backend change here is three docstrings.

maziggy (changes requested):
Reviewed. The backend half is right and the tests behind it are the good kind - they assert the request body, and the 0-versus-absent distinction is pinned where it matters. is not None rather than truthiness, model_fields_set on create against default-None on update, key omitted entirely when nothing was set: all correct, and the reasoning in the comments matches what the code does.

The frontend half needs another pass, because as it stands the two halves of this PR contradict each other.

AdditionalSection.tsx:215 replaces the tare picker with t('inventory.spoolWeightManagedBySpoolman') whenever spoolmanMode is set, and SpoolFormModal.tsx:867 strips core_weight and core_weight_catalog_id from the payload in Spoolman mode - and that one data object serves create, bulk create and update. So after this merges, a Spoolman user still cannot set a per-spool tare from Inventory, Edit. The only place it is reachable is the SpoolBuddy touch form, which renders AdditionalSection without spoolmanMode and so still shows the picker.

This is where your correction on the issue text lands. Step 2 of the repro, "the core weight field shows a value", is not true of dev: dev already implements the second option your Expected Behavior offered, which is not rendering an editable field in this mode. This PR implements the first option on the backend and leaves the second in place on the frontend, so the message left on screen - "Empty spool weight is managed per filament type in Spoolman", in all 15 locales - now asserts exactly what your own evidence disproves.

So the ask is to pick one and finish it. Either show the picker in Spoolman mode, stop stripping the field in SpoolFormModal, and retire or reword those strings; or keep the desktop form as it is and say in the PR why the kiosk may set a tare and the Inventory editor may not. I would take the first, since the 70 g weigh-in error you opened this for is felt by desktop users, not kiosk users.

Two smaller things.

The new comment in SpoolBuddyDashboard.tsx says "The tare is a property of the filament type in Spoolman, not of the spool, and Bambuddy already reads it from there." Both clauses are the misconception this PR exists to correct - Spoolman has spool.spool_weight and Bambuddy reads it with priority. The true reason is narrower: the kiosk quick-create has no tare input, so its 250 was a placeholder rather than a user choice.

And the frontend change has no test. Ten backend tests, each mutation-checked, and nothing covering either frontend edit. SpoolBuddyDashboard.test.tsx mocks createSpoolmanInventorySpool and never inspects the argument, so removing core_weight: 250 from that payload cannot fail anything, and coreWeightTouched is unexercised. Given the standard you have been holding yourself to on the other four, that gap stands out.

Credit on the part that is easy to get backwards: flipping coreWeightTouched on core_weight and not on core_weight_catalog_id is the right hook. SpoolWeightPicker auto-selects a catalog entry on mount when the catalog loads and one entry matches the current weight, so keying on the catalog id would have marked an untouched form as touched - while both real user actions call onChange and do register.

Minor: core_weight_catalog_id is not a field on SpoolmanInventoryCreate, so the kiosk sending it in Spoolman mode is silently dropped. Harmless, but a catalogue selection cannot round-trip there, which may be worth a word in the comment.

PR #3022, opened 2026-09-03
[Fix] Dual-UID spool identification, use tray_uuid end-to-end (rebase of #1200)

### LINKED ISSUE #984: [Bug]: Respect both BL spool tags

Sawtaytoes:
## Description

Rebase of #1200 onto current `dev`, with @maziggy's review items applied. The original work is @Keybored02's — this carries it forward because the branch had gone stale (last touched 2026-07-27) and no longer merges. Happy to close this if @Keybored02 would rather update the original.

**What it does** (unchanged from #1200):

- **Read block 9.** The Bambu tray UID is a 16-byte value at block 9 (sector 2, block 1). `_extract_tray_uuid` read blocks 4+5, which hold the *detailed filament type* — so on a PLA Basic spool it returned `504C4120426173696300000000000000`, the hex of the ASCII string `PLA Basic`, identically for every such spool. `BAMBU_BLOCKS` now includes `9` in the daemon driver and in `read_tag.py`.
- **Add `tag_uid_2`** so one spool can hold both of its physical tag UIDs.
- **Save `tray_uuid`** when a spool is created through SpoolBuddy, and match on it first.
- **Clear SpoolBuddy state on tag removal** so the spool card does not linger.

## Related Issue

Fixes #984. Supersedes #1200.

## Changes made while rebasing

1. **Conflict resolved** — `backend/app/api/routes/inventory.py`: `dev` added `IntegrityError`, the branch added `or_`. Kept both.
2. **Review item 2** — `InventorySpool.tag_uid_2` is now optional (`tag_uid_2?: string | null`) rather than required, which is what you asked for and matches how other additive columns are introduced here.
3. Reverted an unrelated em dash change in the `pn5180.py` module docstring, to keep the diff to the fix.

Review items **1** and **4** were already handled on the branch before it went stale: `tag_uid_2` is `String(32)` with a `VARCHAR(32)` migration, and the redundant `setShowLinkModal(false)` on the success path is gone.

**Review item 3 — Spoolman scoping, stated explicitly as requested:** the dual-UID feature is **local-DB only**. Spoolman's REST API exposes no second tag field and `_map_spoolman_spool` has nothing to read one from, so with `spoolman_enabled=true` the secondary UID column is unreachable — setting it via the form persists locally only, and no Spoolman spool will ever match against it. `tray_uuid` matching *does* work in Spoolman mode, via `find_spool_by_tag`.

## Independent confirmation on hardware

Verified against a live Bambuddy instance with a USB PN532 rather than a PN5180, so this is a second reader confirming the same layout:

```
TAG 2bb197fc  sak=8  tray_uuid=5376FCE9EBB44929A0E45733A6256139
TAG f500d20c  sak=8  tray_uuid=5376FCE9EBB44929A0E45733A6256139
```

Two different tag UIDs, one spool, one block-9 value — and that value is byte-for-byte the `tray_uuid` already stored for that spool from the AMS. Blocks 4+5 on the same tag decode to `PLA Basic`. The full block-9 decode also yields PLA / PLA Basic, RGBA `FFFFFFFF`, 1000 g, 1.75 mm, 190-230 °C, produced `2025_03_20_15_53`.

## One thing I did not change, and think is worth a look

`_fuzzy_uid_match` keeps a "first-character variance" fallback that matches when `candidate_uid[1:] == tag_uid_norm[1:]`, commented as handling *"readers that report different first bytes for the same tag"*. I left it alone rather than redesign someone else's PR, but I believe the premise is a misdiagnosis of the very problem this PR fixes properly: two sides of a spool have entirely different UIDs (see @Thomansky's `2C0273DD` vs `8E3A00A2` in #984), not UIDs differing in one nibble. As written it can match the wrong spool, and it logs a warning admitting so.

The mismatch it is actually papering over looks like padding: the AMS reports a 4-byte UID padded to 8 bytes with a constant `00000100` suffix (`2BB197FC00000100`), while a direct reader reports `2BB197FC`. `normalize_tag_uid` truncates above 16 characters but never pads, so the two can never be equal. A **prefix** rule against that known suffix would be principled; the nibble rule is a guess. Your call whether to address it here or separately.

## Documentation

**Pick one**:
- [ ] Docs PR(s) linked above
- [x] No docs update required — reason: backend fix and behaviour fix, no user-visible config, port, flag or install change.

## Type of Change

## Checks run on this tree

| Gate | Result |
| --- | --- |
| `ruff check backend/` | All checks passed |
| `ruff format --check backend/` | 641 files already formatted |
| `pytest backend/tests/` | **7254 passed** in 138 s |
| `npx tsc -b` | clean |
| `npm run test:run` | **2432 passed** across 183 files |
| `check-i18n-parity` | all 11 locales in parity with en |


Keybored02:
As mentioned in #1200, the scope of the fix goes way beyond this single PR and it's being worked on. So at the current state is not mergeable, and will be included in a future PR once we settle on a way to do it.

Issue #2685, opened 2026-07-27
[Feature]: Add Bambu spools to inventory from SpoolBuddy RFID scans


BenBBSLA:
## Motivation

I use SpoolBuddy to identify and weigh spools. When it scans an unknown genuine Bambu tag, I want to log that spool into Bambuddy from SpoolBuddy without manually re-entering information already present on the RFID tag.

Today, [Quick Add creates a generic PLA record](https://github.com/maziggy/bambuddy/blob/3dcbbdd25e5a86ba267357e7fa4397c5a6569784/frontend/src/pages/spoolbuddy/SpoolBuddyDashboard.tsx#L384-L454) with blank brand and colour fields.

A live test on [SpoolBuddy 1.2.5 (`3dcbbdd`)](https://github.com/maziggy/bambuddy/commit/3dcbbdd25e5a86ba267357e7fa4397c5a6569784) using a Bambu PLA Basic spool returned:

| Field | Decoded value |
| --- | --- |
| Variant / profile | `A00-K0` / `GFA00` → Bambu PLA Basic |
| Material / subtype | PLA / PLA Basic |
| Colour | RGBA `000000FF` → Bambuddy colour catalogue: Black |
| Label weight | 1000 g |
| Diameter | 1.75 mm |
| Tray identity | Valid 16-byte block-9 UUID (redacted) |

The connected scale also sends readings successfully.

## Proposed solution

For an unknown authenticated Bambu tag:

1. Read and parse blocks 1, 2, 4, 5 and 9.
2. Match an existing spool by the canonical block-9 tray UUID first.
3. If unmatched, prefill an **Add to Inventory** confirmation using the existing AMS inventory mapping and current scale weight.
4. If metadata is incomplete, keep the current manual/link flow. Do not silently overwrite an existing spool.

This should prevent the two RFID sides from creating duplicate spools, consistent with [#984](https://github.com/maziggy/bambuddy/issues/984).

## Acceptance criteria

- [ ] The tested spool is prefilled as Bambu Lab / PLA Basic / Black / 1000 g / 1.75 mm.
- [ ] Saving creates one Bambu-tagged record containing its canonical tray UUID, profile, RGBA and current scale weight.
- [ ] Scanning either tag side finds the same spool; incomplete reads fall back safely.

## Alternatives considered

AMS sync or manual entry works, but adds unnecessary steps. Barcode and photo-based onboarding remain useful for third-party spools.

## Related issues and references

[...]

PR #3031, opened 2026-09-03
[Fix] Clear tray_uuid values read from the wrong Bambu tag blocks

grengojbo:
## Description

Complements #3022, which stops the SpoolBuddy daemon reading `tray_uuid` from the
wrong Bambu tag blocks. This handles the values already written by that bug:
existing databases still hold them, and daemons that have not been updated keep
sending new ones.

The tray UUID lives in block 9. A daemon reading blocks 4-5 picks up the filament
description instead, so a white PLA Basic spool reports
`504C4120426173696300000000000000` — the hex of `"PLA Basic"` — and so does every
other spool of that product. Since the backend identifies spools by `tray_uuid`,
a scan resolves to an arbitrary one of them and its weight and assignment land on
the wrong row.

Nothing currently rejects these values: they are 32 characters of valid hex and
pass the `^[0-9A-Fa-f]*$` pattern on `TagScannedRequest`. What sets them apart is
their shape — printable ASCII followed by NUL padding, which a random 16-byte
UUID practically never produces. `is_text_artifact_tray_uuid()` keys on that.

## Changes

- `backend/app/utils/tag_normalization.py` — `is_text_artifact_tray_uuid()`.
- `backend/app/core/database.py` — `_migrate_clear_text_artifact_tray_uuids()`
  nulls affected rows on upgrade. Clearing is the only repair available: the real
  UUID exists solely on the tag and comes back on the next scan by a fixed
  daemon. Until then the spool is still found by its own `tag_uid`.
- `backend/app/schemas/spoolbuddy.py` — drop the field on `tag_scanned` rather
  than reject the request, so an un-updated device stays usable.
- `backend/app/services/spool_tag_matcher.py` — `get_spool_by_tag()` falls
  through to `tag_uid` instead of matching on an artifact.
- `backend/tests/unit/test_spoolbuddy_tray_uuid_artifact.py` — 14 tests covering
  detection, the schema validator, lookup fallback and the migration.

The migration is deliberately not gated behind a `settings` flag, unlike the
one-shot backfills next to it. Those are gated because re-running would undo a
value the user has since chosen; here there is nothing to undo, and while
un-updated daemons are still in the field they keep writing new bad rows that a
later upgrade should also clean up. Rows holding a real UUID never match the
filter, so repeated runs are no-ops — covered by a test.

## Verification

Block layout confirmed against a real tag (MIFARE Classic 1K, UID `E35FACA1`,
white PLA Basic 1 kg), read with HKDF/Crypto1 authentication:

```
block 1: 4130302D573100004746413030000000  "A00-W1", "GFA00"
block 2: 504C4100000000000000000000000000  "PLA"
block 4: 504C4120426173696300000000000000  "PLA Basic"
block 5: FFFFFFFFE80300000000E03F00000000  colour RGBA, 1000 g, 1.75 mm
block 9: 6B9170BB5BD84DDBA748713C159367DF  tray UUID
```

Full backend suite passes; `ruff check` and `ruff format --check` are clean.

## Related Issue

Part of #984. Complements #3022 — no overlapping files. Relevant to #2685, which
needs canonical tray-UUID matching before it can prefill from a tag.

### DIFF
--- a/backend/app/core/database.py
+++ b/backend/app/core/database.py
@@ -4957,6 +4957,10 @@ async def run_migrations(conn):
     # Spoolman and the location sync then imported as storage locations.
     await _migrate_drop_ams_slot_locations(conn)
 
+    # Migration: clear tray_uuid values a SpoolBuddy daemon read from the wrong
+    # tag blocks (#984). They collide across every spool of one filament type.
+    await _migrate_clear_text_artifact_tray_uuids(conn)
+
 
 async def _migrate_rename_ha_sensor_alert_template(conn) -> None:
     """Rename the ha_sensor_alert template to "Printer Sensor Alert" (#2824).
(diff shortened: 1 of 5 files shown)


Keybored02:
Please note our rules in CONTRIBUTING.md.

In particular:

> Open a new issue or comment on an existing one describing what you'd like to work on
> Wait for agreement — discuss the approach with a maintainer so we're aligned on scope and direction
> Get assigned — once we agree, a maintainer will assign the issue to you
> Then start coding — only open a PR for an issue that is assigned to you

As discussed in #1200, the fix for this issue is much more boraded-scoped.

Issue #3046, opened 2026-09-06
[Feature]: Answer the post-print outcome prompt by reacting with a thumbs up/down to the Telegram message


Thomansky:
### Context

#1898 (implementation PR opened right after this issue — linked in a comment below) adds the post-print outcome confirmation: opt-in per job, Bambuddy asks "How did your print come out?" and the answer can be given one-tap from a notification. On Telegram that's inline Good/Reject buttons; on ntfy action buttons; elsewhere plain links.

All of these share one constraint: **the tapping device must be able to reach the Bambuddy host**, because the buttons/links carry `external_url`-based capability URLs. Away from the LAN, behind NAT, or with the VPN down, the buttons go dead — the notification still arrives (it comes via Telegram's servers), but the answer can't get back.

### Proposal

Let the user answer by **reacting to the Telegram message itself**: 👍 on the confirmation message → good, 👎 → reject. The phone then talks only to Telegram; Bambuddy needs **zero inbound connectivity**.

Mechanics — everything needed already exists in the Bot API:

- Bambuddy already gets the sent `Message` (incl. `message_id`) back from `sendMessage`/`sendPhoto`; store it next to the pending confirmation (chat_id + message_id → archive), guarded by `message_id > 0` (the docs reserve 0 for ephemeral/scheduled cases)
- A lightweight long-poll worker calls `getUpdates` with `allowed_updates=["message_reaction"]` and a long timeout — one background task per Telegram provider with the feature enabled, started/stopped with the provider (note `allowed_updates` is sticky across calls and doesn't apply retroactively to updates queued before the call)
- On a `message_reaction` update matching a stored pending message: 👍 → good, 👎 → reject, same single-use semantics as the token links (first verdict from any path wins, later ones no-op); optionally confirm by editing the message ("✅ marked as good" — `editMessageText` for text messages, `editMessageCaption` for the photo variant)
- Unknown emojis and reactions on other messages are ignored

### Settings

A per-provider **verdict mode** choice in the notification provider settings (that's where the channel is configured, so the choice belongs there):

- **Inline buttons** (current behaviour, default)
- **Reactions** — message without buttons, answer by reacting
- **Both** — buttons for on-LAN convenience, reactions as the fallback

### Constraints / open points

[...]

Issue #1444, opened 2026-05-19
[Bug]: Failure Analysis widget shows "Unknown" for archives with failure_reason set after the fact


needo37:
### Component

Bambuddy

### Bug Description

Stats page → Failure Analysis widget → "Top Failure Reasons" groups failed prints as `Unknown` even when the archive has a `failure_reason` set via the Edit Archive modal.

Two `failure_reason` columns exist and they desync:
  - `print_archives.failure_reason` — written by Edit Archive modal (`PATCH /archives/{id}`).
  - `print_log_entries.failure_reason` — read by the widget.

`PrintLogEntry.failure_reason` is set once at print completion (`backend/app/main.py:3603-3620`) by copying `archive.failure_reason`. At that point the archive value is still NULL (user hasn't classified the failure yet). 

Later, picking a reason in Edit Archive writes only to `print_archives` (`backend/app/api/routes/archives.py:1315-1349` — generic `setattr` loop, no log-entry sync), so the log entry stays NULL and `failure_analysis.py:88-97` keeps grouping it as `Unknown`.

Same desync hits the `status` field — flipping status in the modal won't update the log entry either.

DB confirms:
sqlite> SELECT id, failure_reason FROM print_archives WHERE id=22;
22|Adhesion failure
sqlite> SELECT id, archive_id, failure_reason FROM print_log_entries WHERE archive_id=22;
26|22|              -- empty, never updated

### Expected Behavior

  After picking "Adhesion failure" in Edit Archive and saving, the Failure Analysis widget shows that print under "Adhesion failure" (count 1), not under `Unknown`.

### Steps to Reproduce

  1. Run a print, let it fail (or mark a completed one as Failed in Edit Archive).
  2. Open Edit Archive → set Failure Reason = "Adhesion failure" → Save.
  3. Stats page → Failure Analysis → "Top Failure Reasons".
  4. Print is grouped as `Unknown`, not `Adhesion failure`.

### Printer Model

Not printer-related

### Bambuddy Version

0.2.4.2

### Installation Method

Docker

### Operating System

Docker

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I am using the latest version of Bambuddy
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled

maziggy:
[...]

PR #3047, opened 2026-09-06
Post-print outcome confirmation: good/reject verdicts, one-tap links, yield stats

### LINKED ISSUE #1898: [Feature]: Post-print quality outcome — mark a completed print as reject/scrap
### Problem or Use Case

Right now a print's lifecycle effectively ends at completed / failed / cancelled. But "the printer finished the job" and "the part is actually usable" are two different things. A print can complete 100% and still be a reject — stringing, warping, layer shift, a blemish, a dimensional issue found on inspection. There is currently no way to record that a finished print is scrap.

This matters for production: my project progress says "47/130 parts done", but if 4 of those completed prints were cosmetic rejects, I actually have 43 usable parts and need to reprint 4. Today Bambuddy has no place to capture that, so the counts overstate good output and I have to track scrap separately on paper.

### Proposed Solution

Add a quality outcome to a completed print archive, set manually after inspection:

- A control on the archive (and ideally on the print-complete notification / archive card) to mark the finished print as Good or Reject.
- Optionally a reason (free text, or a small preset list like stringing / warping / layer shift / adhesion / other) and/or a photo — reusing the existing notes, tags, photos, failure_reason fields.
- Rejected-but-completed prints should be excluded from "good parts" counts in project/statistics progress (or at least surfaced separately), so production numbers reflect usable output rather than machine completions.

Defaults keep current behaviour: an unmarked completed print counts as good, exactly as today. The QC step is opt-in.

Why it fits the existing model: print_archives already carries most of what's needed — status, failure_reason, notes, tags, photos, is_favorite, quantity. This could be a new nullable quality/qc_result field (or reuse of an existing one) plus a small UI affordance, no schema upheaval. The statistics layer already aggregates per print event, so filtering good vs. reject slots into the existing per-event model.

### Alternatives Considered

Tracking scrap on an external spreadsheet — works but lives outside Bambuddy and drifts out of sync with the real print history. Using the existing notes/tags field manually — possible today, but it's free-form, not counted, and doesn't affect any progress totals, so it doesn't actually correct the "usable parts" number.

### Feature Category

Print Archiving

### Priority

Would improve my workflow

### Mockups or Examples

a Good/Reject toggle on the completed-print archive card, with an optional reason field

### Contribution

### Checklist

Thomansky:
## Description

Adds an opt-in post-print outcome confirmation, Formlabs-style: for jobs that request it, Bambuddy asks "How did your print come out?" after completion — in the app and through notifications with one-tap Good/Reject links — and the verdict feeds the archive, the statistics (new yield rate), and project part counts. A print that finished but produced scrap finally stops counting as a success.

## Related Issue

Fixes #1898

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#84

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- **Per-print opt-in** — "Ask for Outcome" toggle in the print dialog's print options (pre-filled from a new `default_confirm_outcome` setting), stored on the queue item and copied onto the archive at dispatch, following the `plate_id` shape; the option is also surfaced as a green "Ask for Outcome" pill right above the print dialog's action row (create, edit and re-queue), as an "Outcome prompt" badge on the queue card, and as a tri-state toggle in the queue's bulk edit — so it can be set or checked without opening the print-options panel
- **Prints Bambuddy did not start** — a print started at the printer, in Bambu Studio or in the Handy app is archived by Bambuddy but has no queue item to carry the opt-in, so it never asked. New setting `confirm_outcome_external_prints` (default off) gives those archives the same prompt; a queue item that deliberately has the flag off is never overridden, and a queue item still printing on that printer suppresses it (a restart mid-print is the one case where the two paths could collide)
- **Completion flow** — when requested, completion emits a `print_confirm_request` WebSocket event that opens a confirmation dialog anywhere in the app (finish photo, Good / Reject with failure reason, Later, and "Print again" on reject), plus a notification event of the same name (per-provider toggle, default on, editable template)
- **One-tap verdict links** — unauthenticated `GET /api/v1/archives/confirm/{token}/{verdict}` guarded by a single-use 256-bit capability token minted per request and retired on the first verdict from *any* path; responds with a small HTML confirmation page; exempted from the auth middleware following the existing `/dl/` token-in-path precedent; the token is never exposed in API responses
- **Channel affordances** (when `external_url` is absolute): ntfy action buttons, Telegram inline Good/Reject buttons, Pushover supplementary URL, Bark tap-URL — every other channel gets the plain links from the template body
- **Verdict data** — `user_verdict` on the archive, mirrored to the matching `PrintLogEntry` (the #1444 mirror pattern); editable later in Edit Archive (verdict select on completed prints, failure reason now also for rejects)
- **Archives page** — amber "waiting" badge that opens the dialog, green thumbs-up for confirmed good, red for rejected; an "Unconfirmed" filter chip; a context-menu entry; `?confirm=<archive id>` deep link
- **Printer card** — pending verdicts surface at the plate-clear spot: 👍 records good right there, 👎 opens the dialog for the optional reject reason — or just clear the plate as before; an optional `confirm_default_good_on_plate_clear` setting counts the latest unanswered prompt as good once the plate is cleared
- **Statistics** — rejected-print count and yield rate on the Stats page (only shown when rejects exist); the machine failure rate is deliberately untouched — a machine-successful print that produced scrap is a quality problem, not a printer failure; project good-part counts exclude rejects
- **Migrations** — six idempotent `_safe_execute` ALTERs (print_archives ×3, print_log_entries, print_queue, notification_providers)
- **i18n** — 41 new keys in all 14 locales, `npm run check:i18n` green

## Screenshots

**The confirmation dialog** (opens on completion, from the badge, or via the notification deep link):

| Good path | Reject path (reason + print again) |
|---|---|
| after: confirmation dialog | after: reject with reason |

**Archives** — before/after (badges + Unconfirmed filter):

| Before | After |
|---|---|
| before: archives | after: archives with badges |

**Settings** — before/after (default toggle + plate-clear behaviour):

| Before | After |
|---|---|
| before: settings | after: settings |

**Stats** — yield rate next to the existing success metrics:

after: stats with yield rate

## Testing

- Backend: `test_print_confirmation.py` with 9 integration tests and `test_external_print_confirmation_1898.py` with 10 more (the real `on_print_start`, both archive-creating branches, the queue-item precedence) — opt-in plumbing and defaults, PATCH verdict validation and log mirroring with token retirement, the token endpoint (single-use, unknown token → 404, bad verdict → 400), the plate-clear default-good helper (latest-pending-only, no-op safe), the failure-analysis reject counters, and a pin that the confirm route stays exempt from the auth middleware
- Frontend: full suite green (3674 tests, 266 files) incl. i18n parity; the pill / badge / bulk toggle are pinned in `PrintModal.test.tsx` (4) and `QueuePage.test.tsx` (4)
- Live end-to-end on a real instance **with auth enabled**: dialog flow, printer-card verdicts, one-tap links, and Telegram inline buttons against a real bot — the auth-middleware exemption fix came out of exactly this test round
- Backend suite green apart from pre-existing Windows-only platform failures that reproduce identically on a clean `dev` checkout

- [x] I have tested with my printer model: H2S/H2D/X1E farm; the feature adds no printer-facing commands (it hooks the existing completion/archive path)

## Checklist

## Additional Notes

- **Why the auth exemption is safe**: the confirm URL *is* the credential — a per-request `secrets.token_urlsafe(32)` capability token that is single-use, retired on the first verdict from any path (link, dialog, printer card, plate-clear default), never included in any API response, and useless for anything but setting one verdict on one archive. This mirrors the existing `/dl/` slicer-download token links (token in the path, auth middleware steps aside); unlike those it has no TTL, but it is retired on first use and only ever grants one verdict on one archive.
- **Follow-up idea, deliberately not in this PR**: answering by *reacting* 👍/👎 to the Telegram message itself (via `getUpdates` long-polling) — that would work even when the phone has no route back to the Bambuddy host. Filed as #3046 with a per-provider verdict-mode setting sketch; kept out to keep this PR reviewable.
- No new dependencies; the notification-channel changes are additive (new optional parameters defaulting to `None`/absent).


maziggy:
Why don't I see cover images for all your archives?

Thomansky:
Fair question — those four are seeded fixtures, not real prints.

I built a scratch instance with one archive per verdict state so all four badges are visible in a single shot: unanswered, rejected, good and failed. They have no printer, no 3MF and no run behind them, which is why the cover falls back to the placeholder. Anything real would only have shown one or two of the four states at a time.

On a real install the covers are there. For what it is worth I checked my own rather than asserting it: 342 archives, 294 with a cover. The 48 without are all prints Bambuddy did not start — no archived 3MF, so no plate thumbnail to take. Every archive that has a file has its cover, no exceptions either way.

I would rather not swap in a production screenshot: the file and project names there are customer part numbers. If the placeholders are distracting I can re-seed the fixtures with a plate thumbnail attached so the cards look like the real thing.

Thomansky:
Update after running this for a few days rather than only testing it. Three things turned up in use, all now on the branch.

**1. Prints Bambuddy did not start never got asked.** The prompt is driven off the queue item's `confirm_requested`, and a print started at the printer or from Studio has no queue item — so on a machine where a good share of jobs begin that way, the feature looked switched off. There is now an opt-in setting for it: when it is on, a finished print with no queue item behind it gets the same prompt, the same one-tap links and the same yield accounting. Off by default, because on an install that only ever prints from the queue it would be noise.

**2. A one-tap link that had already been answered said "invalid".** With `confirm_default_good_on_plate_clear` on, releasing the plate records the verdict and retires the token. Pressing the thumbs-up in the notification afterwards then landed on an error page that reads like a broken link rather than "this one is already settled". It now answers with the verdict it already holds and when it was recorded. Unknown tokens still 404 — that distinction is the point.

**3. The archive could not say where a verdict came from.** Once a verdict can arrive from the dialog, a link, a Telegram reaction, the printer card, the API or the plate-clear default, "confirmed good" and "assumed good because somebody cleared the plate" are not the same claim, and the yield rate reads differently depending on which it was. Archives now carry the source and the timestamp of the verdict itself, and every path goes through one helper, so a new path cannot forget to stamp it.

One more, smaller: when a prompt is re-sent for a print whose token was already spent, a fresh token is minted instead of re-sending the dead one.

`dev` is merged in, including today's Swedish locale — the 49 new keys are written, and `check:i18n` is green at 15 locales. Taking it out of draft.

PR #3055, opened 2026-09-07
Add per-printer notification aliases

bazza2000:
## Summary
- add an optional, persisted notification alias to printer configuration
- expose `{printer_alias}` to notification templates while retaining `{printer}` for the canonical name
- use the alias in default notification templates with a canonical-name fallback

## Testing
- tested in an isolated BambooBuddy sandbox
- tested against the production print farm with aliases configured for two printers

This lets notification messages use concise names such as `Printer 1` without changing canonical printer identifiers.


maziggy:

Issue #3052, opened 2026-09-06
[Feature]: Feature request: Add Swedish (sv) language support


AntonPalmqvist:
### Problem or Use Case

There's no Swedish translation.

### Proposed Solution

I'd like to contribute Swedish (sv) translation

New file: frontend/src/i18n/locales/sv.ts
Register sv in frontend/src/i18n/index.ts
Extend frontend/scripts/check-i18n-parity.mjs with an SV_COGNATES allow-list

### Feature Category

UI/UX

### Priority

Nice to have

### Contribution

- [x] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

AntonPalmqvist:
PR is ready for review here: https://github.com/maziggy/bambuddy/pull/3060

AntonPalmqvist:
New PR ready for review: https://github.com/maziggy/bambuddy/pull/3062

maziggy:
Thanks.

Issue #3051, opened 2026-09-06
[Bug Report] When editing an archived file that was part of a project, I should be able to ma


maziggy:
When editing an archived file that was part of a project, I should be able to mark the items printed as 0 as a total print plate failure. It won't let me go below 1. In this case the printer jammed in a way that it thought it completed successfully, but everything on the plate was ruined.

### Screenshot
![Bug Report Screenshot](https://raw.githubusercontent.com/maziggy/bambuddy/bug-report-assets/screenshots/72e09eee4c314d5c8371bd5e23f03ab0.png)

<details>
<summary>Reporter Contact</summary>

Email: cartman75@gmail.com

</details>

<details>
<summary>System Information</summary>

```json
{
  "generated_at": "2026-09-06T23:01:04.001182+00:00",
  "app": {
    "version": "1.2.6b1",
    "debug_mode": false
  },
  "system": {
    "platform": "Linux",
    "platform_release": "7.0.0-30-generic",
    "platform_version": "#30-Ubuntu SMP PREEMPT_DYNAMIC Fri Jul 31 18:22:54 UTC 2026",
    "architecture": "x86_64",
    "python_version": "3.13.15"
  },
  "environment": {
    "docker": true,
    "data_dir": "/app/data",
    "log_dir": "/app/logs",
    "timezone": ""
  },
  "database": {
    "archives_total": 245,
    "archives_completed": 217,
    "printers_total": 4,
    "filaments_total": 0,
    "projects_total": 3,
    "smart_plugs_total": 0
  },
  "printers": [
    {
      "index": 1,
      "model": "P1S",
      "nozzle_count": 1,
      "is_active": true,
      "mqtt_connected": true,
      "state": "FINISH",
      "firmware_version": "01.10.00.00",
      "wifi_signal": -45,
      "reachable": true,
      "ams_unit_count": 1,
      "ams_tray_count": 4,
      "has_vt_tray": true,
      "external_camera_configured": false,
      "plate_detection_enabled": false,
      "obico_enabled": true,
      "hms_error_count": 0,
      "developer_mode": true,
      "nozzle_rack_count": 0
    },
    {
      "index": 2,
      "model": "P1S",
      "nozzle_count": 1,
      "is_active": true,
      "mqtt_connected": true,
      "state": "RUNNING",
      "firmware_version": "01.10.00.00",
      "wifi_signal": -74,
      "reachable": true,
      "ams_unit_count": 1,
      "ams_tray_count": 4,
[...]

Issue #1334, opened 2026-05-14
[Bug Report] Up goes down and down goes up on the Bed stepper. I clicked 50 then up and it pr


maziggy:
Up goes down and down goes up on the Bed stepper. I clicked 50 then up and it proceeded to confidently ram the nozzle into my bed. uh oh :(

<details>
<summary>Reporter Contact</summary>

Email: william.filipcic@gmail.com

</details>

<details>
<summary>System Information</summary>

```json
{
  "generated_at": "2026-05-13T22:44:16.923793",
  "app": {
    "version": "0.2.4",
    "debug_mode": false
  },
  "system": {
    "platform": "Linux",
    "platform_release": "6.12.47+rpt-rpi-2712",
    "platform_version": "#1 SMP PREEMPT Debian 1:6.12.47-1+rpt1~bookworm (2025-09-16)",
    "architecture": "aarch64",
    "python_version": "3.11.2"
  },
  "environment": {
    "docker": false,
    "data_dir": "/home/[user]/bambuddy",
    "log_dir": "/home/[user]/bambuddy/logs",
    "timezone": ""
  },
  "database": {
    "archives_total": 0,
    "archives_completed": 0,
    "printers_total": 1,
    "filaments_total": 0,
    "projects_total": 0,
    "smart_plugs_total": 0
  },
  "printers": [
    {
      "index": 1,
      "model": "A1 Mini",
      "nozzle_count": 1,
      "is_active": true,
      "mqtt_connected": true,
      "state": "IDLE",
      "firmware_version": "01.06.00.00",
      "wifi_signal": -65,
      "reachable": true,
      "ams_unit_count": 0,
      "ams_tray_count": 0,
      "has_vt_tray": true,
      "external_camera_configured": false,
      "plate_detection_enabled": true,
      "hms_error_count": 0,
      "developer_mode": true,
      "nozzle_rack_count": 0
    }
  ],
  "settings": {
    "advanced_auth_enabled": "false",
    "smtp_auth_enabled": "true",
    "auth_enabled": "true",
    "setup_completed": "true",
    "auto_archive": "true",
    "save_thumbnails": "true",
    "capture_finish_photo": "true",
    "default_filament_cost": "25.0",
    "currency": "USD",
    "energy_cost_per_kwh": "0.15",
    "energy_tracking_mode": "total",
    "disable_filament_warnings": "false",
    "prefer_lowest_filament": "false",
    "check_updates": "true",
    "check_printer_firmware": "true",
[...]

PR #3062, opened 2026-09-08
Feat: add Swedish language

AntonPalmqvist:
## Description

Adds Swedish (sv) translation 

## Related Issue

Fixes [#3052](https://github.com/maziggy/bambuddy/issues/3052)

## Documentation

**Pick one**:
- [x] No docs update required — reason: Pure translation addition. The new entry appears in the existing language picker (no new UI surfaces, no new config, no new ports/URLs, no installation step change).

## Type of Change

## Changes Made

- Added frontend/src/i18n/locales/sv.ts - full Swedish translation
- Updated frontend/src/i18n/index.ts: register sv
- Extended frontend/scripts/check-i18n-parity.mjs with SV_COGNATES

## Screenshots

image

## Testing

- [x] I ran `npm run test:run` with passing results
- [x] I ran `npm run lint` with passing results

## Checklist

## Additional Notes

Corrections from previous PR:
- Used `dev` as base branch instead of `main`
- Set the correct name for the branch as `feature/add-swedish-translation`
- Added screenshots

Sorry about the previous mistakes, please point me in the right direction if I’m still missing something, thank you!


maziggy (changes requested):
Thanks Anton, and no need to apologise - the base branch, the branch name and the screenshots are all right this time, and the work underneath them is good. I checked the structure mechanically rather than reading the description: no duplicate keys, no keys sv has that en does not, plural suffixes exact, and the Trans-component tags preserved on every string but two. French has quietly dropped <strong> from five strings and nobody caught it, so that is a real bar you cleared. I also checked all 76 SV_COGNATES entries against the strings that are actually identical to English: every one is load-bearing, none dead. The comment saying "curated, not a shortcut" is accurate, which is not something I can usually verify.

eslint, build and the 3664 frontend tests all pass. Two things before merge.

The parity gate is red now, and that is on us, not you. npm run test:run fails with 13 missing keys and 3 placeholder mismatches. I traced each one: they all come from commits that landed on dev between 18 and 20 September, after your 8 September commit - #3051, #3063, #1334, #3090, #3092, #3114 and #3110. Your submission was clean on the day. Pull dev again and add the 13 keys, plus {{id}} in inventory.tagAlreadyLinked and {{runtime}} in the two diagnostic.check.network_mode strings.

A few literal identifiers got translated along with the prose. device/<serial>/report in diagnostic.check.printer_publishing.fail became enhetsämnets/<serienummer>/rapport, and {{prefix}}/printers/<serial>/status in settings.topicPrefixHint became <serienummer>. Those are MQTT topics on the wire - German and the rest keep them verbatim. The one that actually costs a user something is inventory.csv.dragHint: material, brand, subtype, color_name, rgba are the CSV column names the importer matches on, so märke, subtyp, färgnamn produces a file that will not import. de, fr, nl and it all translate the word "Header" and leave the columns alone.

While you are in there, four small ones:
- uppäckning at sv.ts:5630 and :7039 - the file spells it upptäckt correctly everywhere else.
- ett felaktigt eller felaktigt skrift serienummer in the same publishing string; the English is "a wrong or mis-cased serial number".
- slicer vs the skivare family is split roughly 51/63 across the file. Worth picking one - and skivaren sits one letter from skrivaren in an app that says skrivare 393 times, with both words in the printer_publishing.pass sentence.
- exempel.com in the four backup.repoUrlPlaceholder* values should stay example.com; that one is reserved for documentation, exempel.com is somebody's real domain.

Optional: gcodeViewer.feature.support reads as helpdesk support rather than "Stöd", settings.accentTeal is the only English colour among Grön/Blå/Lila/Röd, and spoolbuddy.settings.tabDisplay is the only English tab among Enhet/Våg/Uppdateringar. Translating those three lets you drop three SV_COGNATES entries.

One note on process, for your sake rather than mine: our CI only runs on PRs into main, so the green checks you see here are just the security scans - the lint/test/build job never ran on this PR. npm run test:run locally is the real gate until I fix that.

AntonPalmqvist:
Thanks for pointing out all this. They were all valid points and I've now fixed them.

Good call about "skivare" being so close to "skrivare". I changed it to "slicer", as it's common to create English loan words like that.

Hope I didn't miss anything!

maziggy (approved):

PR #3099, opened 2026-09-14
feat: add optional printer model to stream overlay

### LINKED ISSUE #3080: [Feature]: Add optional printer model to the stream overlay
### Problem or Use Case

It would be nice to be able to add printer model to the stream overlay as not everyone can tell a printer just by looking at it

### Proposed Solution

I’d like to add a Printer model checkbox under Settings → API Keys → Stream overlay. Both would be off by default and included in the generated overlay URL when selected. The details would display alongside the printer name and work with OBS streaming tokens. May I work on this?

### Alternatives Considered

_No response_

### Feature Category

Other

### Priority

Nice to have

### Mockups or Examples

Image

### Contribution

### Checklist

adamspicedev:
## PR stack

1. #3099 — optional printer model and narrow-source layout fix **(this PR)**
2. #3134 — friendly names for legacy model codes

Merge order: **#3099 → #3134**. The follow-up branch descends from #3099. Both PRs target `dev` because the parent branch exists only in the contributor fork; #3134 therefore includes the parent changes until they are merged/rebased.

[Review only the follow-up changes](https://github.com/adamspicedev/bambuddy/compare/feature/stream-overlay-printer-model...fix/stream-overlay-legacy-model-names)

## Description

Add an opt-in **Printer model** checkbox under Settings → API Keys → Streaming Overlay. Selecting it adds `model` to the generated overlay URL. With the printer name also enabled, the overlay shows `Big Mumma · H2D`; with the name disabled, it shows `H2D` alone. Unknown models are omitted.

The token-authenticated overlay feed now includes the printer model for OBS/browser sources. Existing default fields and token authentication remain unchanged.

## Related Issue

Fixes #3080

## Documentation

Companion wiki PR: https://github.com/maziggy/bambuddy-wiki/pull/87

- [x] Docs PR linked above

 The camera guide documents the checkbox, URL parameter, and OBS workflow. No website update is needed because public feature claims are unchanged.

## Type of Change

- [x] Test addition or update

## Changes Made

- Add the model checkbox and label in all 14 locales.
- Render the selected printer name/model independently, joining them with a middle dot when both exist.
- Return the nullable model in connected and disconnected token-feed responses without exposing printer credentials.
- Cover URL/preview updates, default visibility, missing models, OBS data access, and API payload boundaries.

## Screenshots

These captures use the real components with simulated printer data and the camera disabled.

| Before | After |
| --- | --- |
| Before | After |

[Model-only mobile preview](https://raw.githubusercontent.com/adamspicedev/bambuddy/4af7a88324112e958479ebf8c08c0aa3d89196f1/model-only-mobile.png)

### Narrow OBS source (review fix)

| Before | After |
| --- | --- |
| Overflow | Truncated identity |

## Testing

- [x] Tested locally with simulated printer data
- [ ] Tested with a physical printer (not performed)

- Overlay frontend suite: 55 tests passed after the review updates (including missing-model token cases).
- Maintainer verification of the reviewed branch: full frontend suite passed, 265 files / 3,673 tests; SlicerSettingsPanel, QueuePage and PrinterQueueWidget passed 98/98.
- Chromium at 320px width and size=large: long identity reproduced the overflow before the fix and now truncates within the panel.
- Overlay API integration suite: 14 tests passed.
- TypeScript, changed-file ESLint, Ruff lint/format, and all-locale parity passed.
- Production build and Safari 16.0 bundle compatibility check passed.
- Companion wiki built and its new section was checked in Chromium.

## Checklist

- [x] Implementation follows existing overlay configuration and authentication patterns
- [x] New behavior has regression tests
- [x] New label is translated in every supported locale
- [ ] CI checks pass (pending GitHub Actions)

## Additional Notes

Issue #3080 is assigned to adamspicedev. No database migration is required. Full backend, Docker, security, physical-printer, and real OBS checks were not run.

Legacy model-code display is handled separately in https://github.com/maziggy/bambuddy/pull/3134 (dependent follow-up).


maziggy (changes requested):
Nice, clean PR. Verified here: ruff clean, the backend overlay suite 14/14,
eslint clean, build plus the Safari 16.0 baseline check green, i18n parity
green, and the two overlay suites at 53 tests - exactly the number you quoted.

First, the CI caveat in your description: I cannot reproduce it. The full
frontend suite passes on your branch, 265 files and 3673 tests, in 61 seconds,
and SlicerSettingsPanel, QueuePage and PrinterQueueWidget are 98/98 on their
own. Given your run was stopped at 28m30s on an ARM host, those five look like
load-induced timeouts on your machine rather than anything on dev. You can drop
that caveat.

The tests here are better than the size suggests - default opt-in, canonical
show= ordering, both orderings, model-only, no stray separator on a missing
model, and asserting the token path never touches /api/v1/printers/:id. Leaving
DEFAULT_FIELDS alone so existing OBS URLs keep working is the right call, and
the wiki PR says so explicitly. I read all 14 locale strings; they are correct.

One thing before merge:

- The identity row has no truncate and no min-w-0 on the flex item, so a long
  name plus the model at size=large (text-xl, p-6) runs off the right edge of
  the gradient panel in a narrow OBS source. The filename div right below it
  already has truncate. Pre-existing for long names alone, but this change
  lengthens the string by design, so it is the moment to add it.

Two notes, neither blocking:

- A saved printer's model is not guaranteed to be a friendly name. PrintersPage
  runs printer.model through mapModelCode() at :3537, :6720 and :7260 because a
  legacy row can hold a raw code like BL-P001 or O1D, which the overlay would
  render verbatim. Both supported add paths normalise (discovery at :7776,
  manual add via the dropdown), so this only affects older rows - and the same
  file compares printer.model raw at :4153 and :4878, so the codebase is already
  inconsistent about it. Not yours to fix, just worth knowing.
- The missing-model case is covered on the non-kiosk path only; the token test
  only exercises a model that is present. Two lines if you want it.

Cosmetic: fieldModel in fr.ts uses the typographic apostrophe (U+2019) while
fieldPrinter directly above it uses the straight one, as do 103 of the 109
apostrophes in that file.

adamspicedev:
@maziggy — addressed in 41afc46925a8ad81b27e4dfdc014705ebd2a922b, following [your review](https://github.com/maziggy/bambuddy/pull/3099#pullrequestreview-5266733795):

- Added `min-w-0 truncate` to the identity text and `shrink-0` to the printer icon. Verified the overflow before the fix and ellipsis afterward at 320px width with `size=large`; before/after screenshots are in the description.
- Added null and empty-model cases to the token-feed test, retaining the assertion that it never requests printer details.
- Changed the French label to use the straight apostrophe.
- Removed the old timeout caveat and recorded your full-suite result with attribution.

Validation for this update: 55 overlay tests pass, TypeScript and changed-file ESLint pass, and all 14 locales remain in parity.

The legacy model-code note is handled separately in draft #3134, built on this branch: it reuses the existing `mapModelCode` mapping for both overlay data paths. That follow-up passes 61 overlay tests plus 71 printer-page tests, TypeScript, and lint. Merge this PR first, then the follow-up. I know you said it wasn't mine to fix but figured I would help as much as possible

Issue #1268, opened 2026-05-11
[Feature]: Improve File Manager with recursive search, tags, and markdown preview


zumik3-del:
### Problem or Use Case

I frequently use the File Manager tab to search and print models (I have a large collection). It would be great to have a few improvements:

1. Recursive search in subfolders Currently, search only works within the current folder. It would be helpful to have the ability to search recursively through all nested subfolders.

2. Tags for model files It would be convenient to add tags to model files for grouping and searching by tags. For example: "toys", "furniture", "organizers", etc.

3. Markdown preview for models It would be great to have the ability to upload a .md file to a model folder, and have it displayed on a right panel with the model description, printing specifications, and other useful information.

### Proposed Solution

I have a large collection of 3D models organized into folders. Currently, to find the right model I have to constantly open nested folders. With tags, I could quickly find all models in a specific category. And markdown descriptions would help store important notes about each model (material, print settings, etc.) right next to the file.

### Feature Category

Print Archiving

### Priority

Nice to have

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

unLieb:
Tags are something I'm missing too.

maziggy:
Before we commit to building this, I'd like to gauge community interest. If you'd find this feature useful, please give this issue a thumbs up (👍) reaction so we can prioritize accordingly.

maziggy:
1. and 3.

Available/Fixed in branch dev and available with the next release or daily build. Please let me know if it works for you.

-----
If you find Bambuddy useful, please consider giving it a ⭐ on [GitHub](https://github.com/maziggy/bambuddy) — it helps others discover the project!

maziggy:
2.

Available/Fixed in branch dev and available with the next release or daily build. Please let me know if it works for you.

-----
If you find Bambuddy useful, please consider giving it a ⭐ on [GitHub](https://github.com/maziggy/bambuddy) — it helps others discover the project!

PR #3119, opened 2026-09-18
Fix File Manager filename column collapse

### LINKED ISSUE #3105: [Bug]: File Manager list view can collapse filenames to zero width
## Product

Bambuddy

## Area

UI / display / theme / i18n — File Manager list view

## Bug description

In File Manager's list view, the filename column can collapse to zero width when authentication is enabled and the available content width is narrower than the table's fixed columns. The thumbnail remains visible, but the filename disappears and the "Uploaded By" value renders immediately beside the thumbnail.

The list wrapper already uses horizontal scrolling, but the grid itself has no minimum width. Its fixed columns consume the available width while the filename uses `1fr`, so that track is squeezed away instead of the grid overflowing.

## Expected behavior

The Name column keeps enough width to show a useful, truncated filename. On narrower layouts, the table should overflow horizontally using its existing scroll container.

## Steps to reproduce

1. Enable authentication.
2. Open **File Manager**.
3. Switch to list view.
4. Use a viewport where the content pane is narrower than all list columns combined (approximately the width shown on a 1512 CSS-pixel desktop with the sidebar and folder pane open).
5. Observe that file thumbnails are visible, but filenames are not.

## Environment

- Bambuddy: v1.2.5.5 (current latest release)
- Installation method: not relevant to this client-side layout issue
- OS/browser: viewport-independent CSS issue; reproduced from the current `dev` source layout
- Printer model/firmware: not printer-related

## Troubleshooting already performed

- Searched open and closed issues for File Manager filename/name/layout/column reports; no matching issue found.
- Checked the current `dev` implementation.
- Root cause is the authenticated list grid definition: fixed tracks plus a collapsible `1fr` filename track inside an `overflow-x-auto` wrapper.

## Support package / logs

Not applicable: this is a deterministic frontend CSS layout issue and does not involve backend state, printer connectivity, or logs.

## Proposed fix

Give the shared header/row grid a minimum width and make the Name track an explicit `minmax(..., 1fr)`, preserving header/body alignment and allowing the existing wrapper to scroll horizontally.

A regression test should assert the header and rows retain the same minimum-width/grid contract.

akshayy29:
## Description

Prevent the File Manager list view from collapsing its filename column to zero width when authentication and the folder sidebar leave less room than the table's fixed columns.

The shared header/row grid now gives filenames a 240px minimum, uses one fixed 24px leading track to keep every subsequent header and body column aligned, and sizes each auth variant with `min-w-min` inside the existing horizontal overflow wrapper.

## Related Issue

Fixes #3105

## Documentation

**Pick one**:
- [ ] Docs PR(s) linked above
- [x] No docs update required — this is a localized layout bug fix; the File Manager documentation already describes the filename column correctly and contains no affected screenshot.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition or update

## Changes Made

- Give the filename grid track a 240px minimum in authenticated and unauthenticated list views.
- Use a shared 24px leading track so header and row columns remain aligned at the filename floor.
- Use intrinsic `min-w-min` sizing so each auth variant scrolls only when its own content requires it; retain the existing `minmax(0,200px)` tag behavior.
- Add authenticated-list regression coverage for the shared header/row class contract.

### Before

The filename track collapses between the thumbnail and **Uploaded By** columns.

Before: filename column collapsed

### After

The same constrained layout retains a readable, truncated filename column.

After: filename column remains visible

The after screenshot uses local mock file records only; no printer or production data is shown.

## Testing

- [ ] I have tested with my printer model: Not printer-related

Commands and checks run after rebasing onto current `dev`:

- `npm run test:run` — 266 files / 3,678 tests passed; i18n parity passed
- `npm run lint`
- `npx tsc --noEmit`
- `npm run build` — production build and Safari 16 baseline check passed
- Real-browser authenticated grid: 1033.59px intrinsic width; all header/body column offsets 0
- Real-browser unauthenticated grid: 899.19px intrinsic width; all header/body column offsets 0

## Checklist

## Additional Notes

No backend or API behavior changes. The production build retains the repository's existing chunk-size advisory; this change introduces no new warning.


maziggy (changes requested):
Thanks - the diagnosis in #3105 is correct and better written than most bug reports, and the fix does work. I measured it in a real browser at the app's actual root font-size (we set html { font-size: 14.4px }, so gap-4 is 14.4px rather than 16px): auth mode at a 900px content pane gives a 0px filename track before your change and 240px after. That part is right.

Three things to change before this can land.

The min-w-[1120px] is not what fixes the bug. With your minmax(240px,1fr) and no min-width at all, the name track measures 240px at every container width from 390 to 1100 and the wrapper scrolls - the minmax alone does the work. What the min-width actually does is stretch the row box so the hover and selection backgrounds and the bottom border span the scrolled width, which is worth doing but is not what the description says.

And one constant cannot serve both grids. The authenticated grid's own minimum is 1131.2px, so 1120 squeezes the right-hand padding from 14.4px down to 3.3px. The unauthenticated grid's minimum is only 996.7px, so 1120 is about 123px too wide: at a 1008px content pane the list used to fit with a 151px name column and no scrollbar, and now it is forced to 1120 and grows one. Auth-disabled installs on a 1024px window get a horizontal scrollbar on a page that never had this bug.

It also introduces a header/body misalignment. The first track is auto, and the header's spacer is w-6 (21.6px) while the row's checkbox is w-5 (18px). The 1fr name track used to absorb that 3.6px, so columns 3 onward aligned. Pinning the name track at its floor removes the slack and the offset propagates: measured at 1100px, header-minus-row left edges go from [0, 3.6, 0, 0, 0, 0, 0, 0] to [0, 3.6, 3.6, 3.6, 3.5, 3.6, 3.6, 3.6]. That is the same class of bug as the one the comment above the header already documents, in the width band the issue is about.

All three go away with one shape - first track fixed at 24px instead of auto, name minmax(240px,1fr), and min-w-min instead of the constant. Measured: name stays 240px, header/row offsets are 0 at every width, the box sizes itself to 1033.5px with auth and 899.1px without, and the unauthenticated list keeps its scrollbar-free range. No magic number to drift when someone adds a column.

Two smaller ones. You changed the tag track from minmax(0,200px) to minmax(100px,200px), which reverses a deliberate call from #1268 - that is fine to propose, but it needs to be in the description rather than only in a code comment, and the comment at line 2585 still describes the old value and a rationale that is now inverted. And please drop the stray blank line at line 368.

On the test: it is load-bearing in the sense that it fails on the unfixed page, but it fails at getByTestId on the markers this PR adds, and everything it asserts is the literal class string the component sets. jsdom has no layout engine so it cannot see a collapsed column; rewriting 240px as 15rem would break the test without changing any behaviour. I am not going to ask you to delete it, but it is worth knowing it guards the implementation rather than the bug.

akshayy29:
Addressed the review in ac50c026 and rebased the branch cleanly onto current `dev` (58ea7a36):

- replaced the shared `min-w-[1120px]` constant with `min-w-min`
- fixed the first grid track at `24px` for header/body alignment
- retained the deliberate `minmax(0,200px)` tag track
- removed the stray test blank line and updated the class-contract assertions
- updated the PR description to distinguish the filename fix from intrinsic row sizing

Verification after the rebase: 266 frontend test files / 3,678 tests pass, i18n parity passes, ESLint and TypeScript pass, and the production build/Safari 16 baseline pass. Real-browser measurements at the app's 14.4px root size are 1033.59px authenticated and 899.19px unauthenticated, with 0px header/body offset for every column in both modes.

Issue #1292, opened 2026-05-12
[Bug]: Groups resetting for LDAP users on login


Fuechslein:
### Component

Bambuddy

### Bug Description

When assigning LDAP users any other groups than the default group setup in advanced LDAP settings, they get removed again on next login.

It happens only on login, when a user receives a new group while already logged in, he will keep the group and its associated permissions until next login.

### Expected Behavior

LDAP users keep their manually assigned groups.

### Steps to Reproduce

1. Assign an LDAP user a group manually
2. Have the LDAP user login
3. Check their permissions and see that their groups have been reset.

### Printer Model

None

### Bambuddy Version

0.2.4

### Installation Method

Docker

### Operating System

Linux (Ubuntu/Debian)

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I am using the latest version of Bambuddy
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled

maziggy:
Available/Fixed in branch dev and available with the next release or daily build. Please let me know if it works for you.

-----
If you find Bambuddy useful, please consider giving it a ⭐ on [GitHub](https://github.com/maziggy/bambuddy) — it helps others discover the project!

Issue #1569, opened 2026-05-30
[Bug]: OIDC auto-provisioning does not populate email when Email Claim is set to preferred_username


anderl1969:
### Product

Bambuddy

### Area

Authentication / users / permissions

### Bug Description

When OIDC auto-provisioning is used and the **Email Claim** is set to `preferred_username`, the user's email field is not populated even though the ID token contains a valid `email` claim.

### Details
- ID token includes both email and preferred_username
- Email Claim is configured as: preferred_username
- Auto user creation is enabled

### Result
- Username is correctly set from `preferred_username`
- Email field on the created user is empty

### Expected behavior

The user’s email field should be populated from the `email` claim when present, regardless of the configured Email Claim used for username/identity mapping.

### Notes

This prevents the common setup of:

- username from `preferred_username`
- email from `email`

in OIDC auto-provisioning scenarios.

### Steps to Reproduce

1. Configure an OIDC Identity Provider (e.g. Authentik) with a client in BamBuddy.
2. Ensure the ID token contains at least the following claims:
- `email`
- `preferred_username`
3. In BamBuddy, enable:
- OIDC authentication
- Automatic user creation / provisioning
4. Set **Email Claim** to `preferred_username`.
5. Log in with a user that does not yet exist in BamBuddy.
6. Inspect the created user in BamBuddy’s user administration.

### Troubleshooting steps already taken

- Verified that the ID token contains both `email` and `preferred_username` claims.
- Confirmed that `email` is correctly populated by the identity provider (Authentik).
- Tested different configurations of the **Email Claim**:
  - `email` → email field is populated correctly, but username is derived from email local part
  - `preferred_username` → username is set correctly, but email field remains empty
- Verified that required OIDC scopes (`openid`, `email`, `profile`) are included.
- Checked that no additional user mapping or transformation rules are applied in BamBuddy.

### Printer Model

None

### Bambuddy Version

0.2.4.4

### Installation Method

Manual (git clone)

### Operating System

Linux (Ubuntu/Debian)

### Support Package (.zip) and / or relevant logs

[bambuddy-support-20260530-165239.zip](https://github.com/user-attachments/files/28423528/bambuddy-support-20260530-165239.zip)

### Screenshots

<img width="405" height="579" alt="Image" src="https://github.com/user-attachments/assets/45384359-6dd4-4a5e-9c26-6224a269e61a" />

### Final checks

- [x] I searched existing (open AND closed) issues and this bug hasn't already been reported or resolved
[...]

PR #3122, opened 2026-09-18
Sync OIDC provider groups to BamBuddy groups on every login (issue #3107)

### LINKED ISSUE #3107: [Feature]: Sync OIDC provider groups to Bambuddy groups via configurable mapping
### Problem or Use Case

I run Bambuddy with OIDC login and the group situation is a daily annoyance. LDAP logins already have this solved: the `ldap_group_mapping` setting lets an admin map a directory group to a Bambuddy group, and the sync runs on every login. OIDC has no equivalent. An auto-created OIDC user gets the provider's default group and nothing else, and their Bambuddy groups never follow what the IdP says after that. I end up moving users between groups by hand in the admin UI, and it drifts again as soon as anything changes on the IdP side.

Part of this was asked for in #1173. The username part was fixed in #1176, the groups part wasn't.

### Proposed Solution

Two new per-provider settings:

- Group claim name. Defaults to `groups`. Providers disagree on where groups live and what shape they're in (Keycloak sends a JSON array, Authentik sends a space separated string), so I'd accept both shapes.
- A mapping of IdP group to Bambuddy group. The admin picks the pairs, same style as the LDAP group mapping. No name matching.

I'd run the sync on every OIDC login, not just when the account is first created. Only groups named in the mapping would be managed, so manual assignments to other groups survive logins. That matches what the LDAP sync does since the #1292 fix. A provider with no mapping configured behaves exactly as it does today.

This goes together with #1727: group sync puts the user in the right group, printer scoping decides what that group can see.

### Alternatives Considered

Auto-create Bambuddy groups from IdP group names. Fills the group list with IdP internals and gives no control over which groups actually get access.

Sync only at account creation. Covers the first login and nothing after.

### Feature Category

Other

### Priority

Would improve my workflow

### Contribution

### Checklist

willuhmjs:
## Description

An auto-created OIDC user gets the provider's `default_group_id` (or Viewers) and nothing else after that. Whatever the IdP says about group membership never reaches Bambuddy, so every group change has to be done by hand in the admin UI and drifts again on the next IdP-side change. LDAP already solves this: `ldap_group_mapping` maps directory groups to Bambuddy groups and `_sync_ldap_user` applies it on every login. This gives OIDC the same thing, following the LDAP contract rather than inventing a second one.

Two per-provider settings, in the SSO/OIDC form and in the `BAMBUDDY_OIDC_*` env path from #2593:

- **Group Claim** (default `groups`): which JWT claim holds the user's IdP groups. Providers disagree on the shape, so the sync accepts a JSON array (Keycloak, Authentik) and a space or comma separated string (Logto and others).
- **Group Mapping**: explicit IdP group to Bambuddy group pairs. Names don't have to match, the admin picks both sides.

### Only the mapped slice is managed

Same rule the LDAP sync settled on in #1292. Groups named in the mapping values follow the IdP on every login, everything else on the user is a manual assignment and is left alone. So revocation at the IdP propagates on the next login, but promoting a user into a non-mapped group sticks instead of being wiped. One deliberate difference from LDAP: there is no default-group fallback in the sync. `default_group_id` applies once at account creation, and re-asserting it on every login would revert any manual promotion out of Viewers on the next SSO sign-in.

Matching is case insensitive on the IdP side, since group casing is not stable across providers and a silent case mismatch is the failure an admin can least diagnose from the UI. A missing claim means "no mapped groups" and never blocks a login that already authenticated; a sync failure is logged and the user keeps the groups they had.

### The form

The mapping is edited as rows (IdP group name plus a dropdown of the real Bambuddy groups) rather than raw JSON, so an unknown group can't be typed in to begin with. Since most IdPs only return a claim for a scope that was requested, the Email Claim and Group Claim fields now warn inline when the claim name isn't part of the configured Scopes. Deleting a Bambuddy group doesn't rewrite any provider's mapping, so a row pointing at a deleted group shows the stale name with a warning instead of silently showing no selection.

## Related Issue

Fixes #3107

## Documentation

- Wiki: maziggy/bambuddy-wiki#89
- Website: not required. `features.html` already lists Single Sign-On (OIDC / SSO); this adds configuration within an existing capability, same call as #2625.

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- **`group_claim` + `group_mapping` columns** on `oidc_providers`, inline migration in `run_migrations()`. Defaults reproduce the old behavior (`groups` / empty mapping), so an upgrade is a no-op and a provider with no mapping configured never touches group state.
- **`backend/app/services/oidc_group_sync.py`** holds the sync. It runs in `oidc_callback` after the user is resolved and before the exchange token is minted, so every login applies it, existing linked users included. It computes the mapped slice, keeps everything outside it, and commits only when the set actually changed.
- **Provider CRUD** rejects mapping values with no matching group (422, the same answer `default_group_id` gets), checked as a set so three entries naming the same group is one lookup.
- **`BAMBUDDY_OIDC_GROUP_CLAIM` / `BAMBUDDY_OIDC_GROUP_MAPPING`** in `core/oidc_env.py`. Values are group names, not ids, for the same reason as `BAMBUDDY_OIDC_DEFAULT_GROUP`. Unknown names or invalid JSON refuse the whole provider config at boot: a half-resolved mapping would grant exactly the groups whose names happened to match and silently drop the rest. The reason is logged, the app still starts, an already running provider is left as it was.
- **Settings UI** (`OIDCProviderSettings.tsx`): claim input, row based mapping editor, scope mismatch warning on the claim fields, orphaned row flagging, Group Sync line on the provider summary card.
- **i18n**: new strings in all 14 locales; `.env.example` documents the two new variables in the existing `BAMBUDDY_OIDC_*` block.

## Screenshots

| Screenshot | What it shows |
| --- | --- |
| Provider edit form | Group Claim field and Group Mapping as rows (IdP group name plus a dropdown of the real BamBuddy groups), with Add Mapping below.<br>OIDC provider edit form with Group Claim input and Group Mapping rows mapping syskids to Administrators and consultants to Consultants |
| Provider summary card | The saved provider shows its Group Sync status (claim name included) next to Default Group.<br>Saved Keycloak provider summary card with Group Sync On and claim groups listed |
| SSO/OIDC settings overview | The provider in context in Settings → Authentication → SSO/OIDC.<br>SSO OIDC settings tab with the Keycloak provider enabled and group sync active |

## Testing

- [ ] I have tested with my printer model: n/a, this doesn't touch printers

- backend: new suite `backend/tests/integration/test_oidc_group_sync.py`, 26 tests (claim shape parsing, case insensitive matching, manual group survival, revocation propagation, dangling mapping tolerance, no mapping no-op, schema validation, column round trip); existing suites re-run green, including the full mocked OIDC callback e2e in `test_mfa_api.py`
- frontend: tsc, eslint, vitest (including `OIDCProviderSettings.test.tsx`) and the i18n parity check pass
- verified on a Kubernetes instance against a real group based OIDC IdP

## Checklist

- [x] Code follows the project's style guidelines (ruff check + format clean)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective / my feature works
- [x] New and existing unit tests pass locally


maziggy (changes requested):
Thanks, this is a well-argued PR. The managed-slice contract really does mirror
_sync_ldap_user line for line, dropping LDAP's default-group fallback is the
right call for OIDC, and refusing the whole provider config on an unresolvable
env mapping rather than half-applying it is the right disposition. Verified
green here: ruff, 233 backend tests across the OIDC and LDAP group-sync suites,
eslint, build + baseline check, i18n parity.

One blocker:

1. GET /api/v1/auth/oidc/providers is public (mfa.py:1323) and now returns both
   new fields. With no Authorization header at all I get back
   "group_mapping": {"fablab-staff": "Operators", "it-team": "Administrators"} -
   so anyone who can reach the login page learns which IdP group name grants
   Administrators here. The login page only needs id, name, icon and
   is_autologin; please keep group_claim and group_mapping off the public
   serialisation.

Cheap and user-visible, would like these in the same round:

2. The scope warning fires on every provider by default. I rendered the stock
   Add Provider form (scopes "openid email profile", group_claim "groups") and
   the warning is already showing before you type anything. Claim names and
   scope names are different namespaces - only "email" coincides. It also
   false-alarms on the documented email_claim=preferred_username setup from
   #1569, and Keycloak usually delivers groups via a client mapper with no
   dedicated scope. I'd just drop the check.
3. 13 of the 14 locales label the field "Group Mapping (JSON)" while en says
   "Group Mapping" and the UI is a row editor with a dropdown.
4. 10 locales also reword the unrelated LDAP groupMappingHint. Please revert
   those - and note the ko rewrite drops the trailing space that
   LDAPSettings.tsx:416 depends on, so it now renders
   "사용 가능한 그룹:Administrators, Viewers".

Smaller, in the callback and the service:

5. getattr(user, "groups", None) is not None is dead - User.groups is
   lazy="selectin" (user.py:74), so it is always a loaded list. And if it could
   lazy-load, getattr would not have helped: MissingGreenlet is not
   AttributeError.
6. The db.refresh comment says it is for "the permissions baked into the login
   below", but nothing below reads user.groups - the callback only puts
   user.username on the exchange token, and /oidc/exchange re-selects with
   selectinload. With expire_on_commit=False (database.py:163) it is not needed
   on the success path either. Fine to keep as a post-rollback guard, but move
   it inside the group_mapping branch so sync-off installs do not pay two
   queries per login.
7. _MAX_CLAIM_ITEMS only bounds the list shape: 5000 entries as a list gives
   500 back, the same 5000 space-separated gives 5000.
8. _missing_group_names: I tested the case-only mismatch and it is correctly
   rejected, because exact comes back empty. So the case-insensitive cohort only
   fires when one request submits both spellings, and then it admits the
   wrong-cased one that the sync's exact Group.name.in_() will never resolve -
   the failure the docstring says it prevents. Either drop the branch (matching
   the env path's Group.name == target) or make both sides insensitive.
9. The migration adds both columns nullable while create_all makes them NOT NULL
   on fresh installs, so the comment "Both nullable-free with defaults" only
   holds on one of the two paths.

On tests: the 26 are good but they all sit on the helpers, the service and the
schema. Neither route-level 422 is covered, the env path is not covered at all
(that is the code most likely to strand an operator at boot), and oidc_callback
is never exercised end to end with a mapping. OIDCProviderSettings.test.tsx
still passes 20/20 but was not touched, so the row editor, the orphan option and
the warning have no frontend coverage.

Two notes rather than requests: group_claim reuses the email-claim validator, so
namespaced claims are refused and Auth0 group sync cannot be configured at all -
worth relaxing for this field or at least saying so in the wiki PR. And a mapping
that manages Administrators can empty an install of admins on the next login,
since the last-admin guards in users.py only cover role change, deactivate and
delete. That is identical to the LDAP sync, so not yours to fix here.

willuhmjs:
Everything's addressed in 53a7b176, including the route/env/callback tests and the missing bound + sync-failure tests. Heads-up that test_download_deadline_and_gate_2957 fails on a clean dev checkout too, so that failure isn't from this PR.

| Review point | Before | After |
|---|---|---|
| 1. Public provider list leaks group sync config | `GET /oidc/providers` returned the full provider, mapping included | New `OIDCProviderPublicResponse` with `id`, `name`, `has_icon`, `is_autologin` only; full shape stays on `/providers/all` |
| 2. Scope warning fires by default and false-alarms on `preferred_username` / Keycloak mappers | Warning rendered on the stock form before anything was typed | Dropped entirely, component + all 14 locales, with a test pinning it stays gone |
| 3. 13 locales label the field "Group Mapping (JSON)" while en says "Group Mapping" | Stale label from the JSON-textarea version | Label matches en in every locale |
| 4. 10 locales reword the LDAP `groupMappingHint`; ko drops the trailing space `LDAPSettings.tsx` depends on | Divergence from upstream | Reverted byte-for-byte to upstream in every locale, ko trailing space included |
| 5. `getattr(user, "groups", None) is not None` is dead (lazy="selectin") | Guard that never fired | Removed; the block keys off `provider.group_mapping` |
| 6. `db.refresh` comment claims token freshness; runs on sync-off installs | Refresh on every login | Moved inside the mapping branch; comment now states the real job (post-rollback guard) |
| 7. `_MAX_CLAIM_ITEMS` bounds the list shape only | 5000 space-separated fragments returned 5000 | Slice applies after splitting; both shapes bounded, tests pin the bound |
| 8. `_missing_group_names` CI cohort admits a case variant the sync's exact lookup never resolves | Case-insensitive acceptance | Exact match only, same as the sync and the env path |
| 9. Migration adds nullable columns; create_all builds NOT NULL | Fresh and upgraded installs disagreed | Explicit `NOT NULL DEFAULT` on the ALTER path |
| Note: namespaced claims refused, so Auth0 group sync can't be configured | Email-claim charset reused | Group claim takes `/` (`app/roles`); test added |
| Tests: no route 422s, no env path, no callback e2e, no frontend coverage | Helpers/service/schema only | `test_oidc_group_sync_routes.py`: both 422s, case-variant rejection, public-shape assertion, env apply/refuse/bad-JSON/clear, two callback e2e (mapped grant at creation; revocation + manual + creation-default survival). `OIDCGroupSyncForm.test.tsx`: row editor, empty save, row removal, orphan flagging, summary card, warning-absence |

Issue #1528, opened 2026-05-25
[Feature]: WLED Integration for Printer Status Visualization


d3nn3s08:
### Problem or Use Case

Currently Bambuddy already provides rich printer and queue state information internally, but there is no simple way to visualize those states externally in real time.

Many users already use WLED devices in workshops, print farms and smart home environments as status indicators. Right now this requires custom Home Assistant automations, MQTT bridges or external scripts.

A native WLED integration would allow Bambuddy to directly communicate printer states visually without requiring additional middleware.

This would especially help with:

* monitoring printers from across a room
* identifying printer errors immediately
* visualizing queue states
* showing print completion states
* improving print farm monitoring
* simplifying smart-home integrations

### Proposed Solution

It would be great to have native WLED integration in Bambuddy to use LED strips or lamps as real-time printer status indicators.

Possible use cases:

* Printing → animated effect / progress color
* Idle → soft white or off
* Error / HMS alert → flashing red
* Finished print → green pulse
* Paused print → yellow
* Queue waiting / plate clear required → orange
* Filament issue → purple/red blinking

This could work either via:

* WLED JSON API
* HTTP presets
* MQTT integration
* or simple webhook support

A lightweight implementation could already be very useful by allowing users to map Bambuddy printer states to:

* WLED preset IDs
* colors
* brightness
* effects

This would integrate especially well with print farms, workshops and smart home setups.

### Feature Category

Monitoring & Stats

### Priority

Would improve my workflow

### Contribution

- [x] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Before we commit to building this, I'd like to gauge community interest. If you'd find this feature useful, please give this issue a thumbs up (👍) reaction so we can prioritize accordingly.

d3nn3s08:
> Before we commit to building this, I'd like to gauge community interest. If you'd find this feature useful, please give this issue a thumbs up (👍) reaction so we can prioritize accordingly.

I already implemented a small local prototype while testing the idea and it works surprisingly well so far.
The implementation was intentionally kept small and fully optional.
If there is enough community interest, I would be happy to open a draft PR for feedback.
(Sorry for my bad English, I use a translator.)

maziggy:
Plese note our rules from CONTRIBUTING.md. Thanks!

d3nn3s08:
[...]

PR #3124, opened 2026-09-19
feat(wled): add preset-based WLED integration for printer status visualization

d3nn3s08:
## Description

This PR adds an optional **preset-based WLED integration** for Bambuddy.

Instead of configuring colors and effects inside Bambuddy, printer states are mapped to existing WLED presets. This keeps all colors, effects, playlists and future WLED features managed entirely within WLED, while Bambuddy only triggers the selected preset.

The implementation also includes several usability improvements to simplify configuration and testing.

## Related Issue

Related to #1528

This PR implements the requested WLED integration and is submitted for review and community feedback.

## Documentation

**Pick one**:

- [ ] Docs PR(s) linked above
- [x] No docs update required — reason: The feature is self-explanatory and does not introduce new installation steps or configuration outside the existing settings page.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Test addition or update

## Changes Made

- Added optional preset-based WLED int
- Added per-printer WLED configuration
- Added printer state → preset mapping
- Added preset loading directly from WLED
- Added connection testing
- Added per-state test buttons
- Added optional Finished timeout with automatic return to the Idle preset
-Added responsive two-column settings layout
- Added backend and frontend tests
- Added WLED connection testing
- Added configurable printer state → preset mapping

## Screenshots

See attached screenshots below.

## Testing

- [x] I have tested with my printer model: A1 Mini

## Checklist

## Additional Notes

The preset-based approach intentionally keeps WLED as the single source of truth for colors, effects and animations.

This avoids duplicating WLED functionality inside Bambuddy while still allowing users to fully customize their LED behavior using native WLED presets.

The implementation was designed to keep the integration simple, maintainable and future-proof.

wled-settings-two-column-desktop
Screenshot 2026-09-19 093944


Thomansky:
Really nice to see this land — the preset approach is exactly the right cut; WLED owning the visuals keeps Bambuddy simple. I went through the diff with my own setup in mind, and there's one case I think deserves a small addition (happily as a follow-up, if you'd rather keep this PR as it is):

**Several printers on one WLED device.** `wled_config` lives per printer with its own `base_url`, so N printers with N strips work fine. A *shared* strip doesn't: `send_preset` POSTs `{"ps": id}`, and a WLED preset applies to the whole device — so every status change of any printer repaints the strip for all of them, last change wins. That's how I run it, one bar above the door with a numbered segment per printer, and I'd guess it's the common shape once there are more than two printers:

WLED status bar with one segment per printer

**Proposal (backwards compatible):** an optional `segment` (WLED segment id) in `WLEDConfig`. When it is set, WLED still defines the look — Bambuddy reads the mapped preset from `/presets.json`, takes its first `seg` entry as the *template* for that status, and applies it to the printer's segment only: `POST /json/state {"seg": [{"id": <segment>, ...template minus id/start/stop}]}`. One preset per status, shared by every printer on the strip; colours and effects stay in WLED; no segment configured → today's global `ps` path, so nothing changes for single-strip users. The finished-timeout return to idle goes through the same send path, so it just works. UI-wise it's one numeric field next to the base URL (the segment count is right there in `/json/state` for validation).

I'd be glad to build that on top of this once it's merged, or send it as a commit to your branch if that's easier — whichever you prefer.

d3nn3s08:
Thank you very much for the kind words!

I'm really happy that you like the preset-based approach. My goal was to keep WLED responsible for colors, effects and animations while Bambuddy simply maps printer states to presets.

Your idea about optional segment support makes a lot of sense, especially for shared WLED installations with multiple printers. I intentionally kept the first version as simple as possible, but I agree that this would be a great backwards-compatible extension.

Also, this is actually my very first Pull Request to an open source project, so I'm really happy and encouraged that the overall approach is being received so positively.

If the maintainers like the current implementation, I'd be happy to extend it with optional segment support afterwards. Of course, I'd also be happy if you'd like to contribute that part directly.

d3nn3s08:
I'm currently simplifying the implementation based on the results of the live WLED validation. The cleanup is in progress and I'll push the final changes tomorrow.

Issue #3077, opened 2026-09-10
[Feature]: File Manager: Similar to Archives add Fields: External Link and Option to upload images


SergioFuchs:
### Problem or Use Case

Currently, in File Manager, I can only edit a file's name and folder. It would be nice to also have the option to add more information, such as an external link, images, and notes, similar to the options available when editing files in Archive.

### Proposed Solution

Similar to the options available in Archives, it would be helpful to add support for external links, notes, and images in File Manager, if possible.

Thanks!

### Feature Category

File Manager

### Priority

Nice to have

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

Thomansky:
Big +1 — and part of this is closer than it looks: `library_files` already has a `notes` column, and `PUT /api/v1/library/files/{id}` already accepts `notes`; the File Manager UI just never surfaces it. The rest maps cleanly onto machinery the Archives already have. Here is a worked-out design; I'd be willing to implement it.

**Data model** (mirrors the archive columns):

- `library_files.external_url VARCHAR(500)` — same shape as `print_archives.external_url` (one user-defined link: Printables, MakerWorld, supplier page, …). (Related: `library_files.source_url` already records MakerWorld provenance on import but isn't part of the library file responses today — it only surfaces in the MakerWorld recent-imports list — so the details view could show it read-only next to the user-defined link.)
- `library_files.photos JSON` — a list of stored filenames, exactly like `print_archives.photos`.
- Notes: no schema change needed; the list response gains a lightweight indicator (`notes` or a `has_notes` flag) so grid/list can show a badge without fetching every file.
- Upgrades: two idempotent add-column migrations in the established `_safe_execute` pattern.

**API** (mirrors the archive photo routes):

- `POST /library/files/{id}/photos` — multipart upload, extension allowlist (`.jpg/.jpeg/.png/.webp`), stored under the library data dir with a generated name; ownership-gated like the existing file update.
- `GET /library/files/{id}/photos/{filename}` — membership check against the file's photo list first, containment-checked path resolution, and served with the media-token pattern from #3025 (`require_media_token_ownership`, `?token=` query param) so a plain `<img src>` works with auth enabled and ownership is still enforced — same as archive photos and library thumbnails today.
- `DELETE /library/files/{id}/photos/{filename}`.
[...]

Issue #3025, opened 2026-09-03
[Bug]: `camera:view` is load-bearing for every image in the app — 13 non-camera media routes are gated by the camera stream token


lonix:
### Product

Bambuddy

### Area

Authentication / users / permissions

### Bug Description

# [Bug]: `camera:view` is load-bearing for every image in the app — 13 non-camera media routes are gated by the camera stream token

**Product:** Bambuddy
**Area:** Authentication / users / permissions
**Printer Model:** X2D
**Bambuddy Version:** v1.2.6b1 (daily, image built 2026-08-30)
**Installation Method:** Docker
**Operating System:** Linux (Other) — AlmaLinux 10.2

### Bug Description

`RequireCameraStreamTokenIfAuthEnabled` guards 16 route handlers in 1.2.6b1. Only
**three** of them are camera routes:

| route | file:line | camera? |
|---|---|---|
| `GET /library/files/{id}/thumbnail` | `library.py:5296` | no |
| `GET /library/files/{id}/plate-thumbnail/{plate}` | `library.py:3216` | no |
| `GET /archives/{id}/thumbnail` | `archives.py:2323` | no |
| `GET /archives/{id}/timelapse` | `archives.py:2552` | no |
| `GET /archives/{id}/photos/{filename}` | `archives.py:3268` | no |
| `GET /archives/{id}/qrcode` | `archives.py:3356` | no |
| `GET /archives/{id}/plate-preview` | `archives.py:3697` | no |
| `GET /archives/{id}/plate-thumbnail/{plate}` | `archives.py:4230` | no |
| `GET /archives/{id}/project-image/{path}` | `archives.py:4681` | no |
| `GET /projects/{id}/cover-image` | `projects.py:1415` | no |
| `GET /print-log/{id}/thumbnail` | `print_log.py:139` | no |
| `GET /external-links/{id}/icon` | `external_links.py:242` | no |
| `GET /printers/{id}/cover` | `printers.py:1149` | no |
| `GET /printers/{id}/camera/stream` | `camera.py:849` | **yes** |
| `GET /printers/{id}/camera/snapshot` | `camera.py:1168` | **yes** |
| `GET /printers/{id}/camera/plate-detection/references/{i}/thumbnail` | `camera.py:1731` | **yes** |

The only way to mint that token is

```python
# app/api/routes/camera.py:832
@router.post("/camera/stream-token")
async def create_stream_token(
[...]

PR #3128, opened 2026-09-20
File Manager: external link, notes and photos on library files

Thomansky:
## Description

Gives library files the same trio the Archives already have: an external link, notes and photos. A new **File details** entry in the card menu (and an info icon on list rows) opens a modal with the file's facts, an editable notes field and link, and a small photo gallery; the grid shows a globe icon, a notes indicator and a photo count so the information is visible without opening anything.

`library_files.notes` and the `PUT /library/files/{id}` route already existed — the UI just never surfaced them. The rest mirrors the archive implementation.

## Related Issue

Fixes #3077

Builds on #2990 (the preview work touches the same context menu and modal mounts); this PR's own diff against that branch is 29 files.

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#90

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- **Data** — `library_files.external_url` (VARCHAR 500) and `library_files.photos` (JSON list of stored filenames), two idempotent `_safe_execute` migrations; `FileUpdate` accepts `external_url` (http/https only, empty string clears it); `FileResponse` gains `external_url`, `photos` and the read-only `source_url` (MakerWorld provenance, previously never exposed); `FileListResponse` gains `external_url`, `has_notes` and `photo_count` so the grid needs no extra requests
- **Photos** — `POST /library/files/{id}/photos` (multipart, `.jpg/.jpeg/.png/.webp`, 10 MB cap, ownership-gated like the file update), `GET /library/files/{id}/photos/{filename}` (membership check against the stored list first, containment-safe path, served with the #3025 media-token pattern exactly like library thumbnails) and `DELETE`; files live under `<library dir>/photos/<file_id>/` and are removed when a trashed file is purged; external-folder files are supported the same way
- **Details modal** (`LibraryFileDetailsModal`) — filename and type, size, print facts from the detail response, the MakerWorld source as a link when present, notes textarea, link input, photo grid with upload tile and delete, lightbox via `PhotoGalleryModal` (which now takes either an `archiveId` or a `getPhotoUrl` — a typed either/or, both existing callers unchanged)
- **Entry points** — *File details* in the card's three-dot menu and an info icon in the list row's actions; globe icon (opens the link in a new tab, same affordance as Archives), notes indicator and photo-count badge on the card
- Unsaved edits survive background refetches of the file (photo upload/delete, window focus)
- i18n: new keys under `fileManager.details.*` in all 14 locales, `npm run check:i18n` green
- Out of scope, as discussed on the issue: folders (the README convention covers them), multiple links, the thumbnail pipeline

## Screenshots

| Card menu | Grid indicators (globe, notes, photo count) |
|---|---|
| context menu with Details | grid indicators |

**Details modal**

details modal

## Testing

- Backend: `test_library_file_metadata_3077.py` (15 tests) — upload/serve/delete round-trip, unknown filename → 404 even when such a file exists on disk, traversal-looking names, wrong extension, oversize, link round-trip and clearing, non-http schemes → 422, list fields, trash purge removing the photo directory; the SSRF body-URL guard lists the new field as stored-never-fetched
- Frontend: `LibraryFileDetailsModal.test.tsx` and `FileManagerFileDetails.test.tsx` (menu entry on card and row, indicators, save path, refetch-safety); full suite green, i18n parity green
- Manually on a running instance: notes, link and photo upload/display on a library file (the screenshots above); the external-folder and trash-purge paths are covered by the tests

- [x] I have tested with my printer model: no printer-facing changes

## Checklist

## Additional Notes

- `ArchiveUpdate.external_url` still accepts any string; tightening it to http/https the same way would be a small follow-up rather than a change in this PR.
- If #2990 is merged first, this PR rebases to its own 29 files automatically.


maziggy (changes requested):
Reviewed the whole thing on the branch as it stands (448375961), ran the gates and proved the migration against a real PostgreSQL instance. This is careful work - the photo routes are better than the archive ones they copy, the trash interaction is right, and the wiki PR is unusually thorough. A handful of things before it goes in.

Blocking: the i18n parity gate is red. npm run check:i18n exits 1 - sv has 6302 leaves against 6340 everywhere else. All 38 missing keys are this branch's: 27 fileManager.details.* and 11 fileManager.preview.* from the preview groundwork, including the photoCount_one/_other pair. Not the translation's fault - today's dev merge brought sv.ts in and the keys were never added to it. The other 14 locales are complete and pass the identical-to-en check, and the new FR_COGNATES entries are legitimate.

The lightbox always opens on the first photo. Clicking the third thumbnail shows "Photo 1 of 3". PhotoGalleryModal has no initialIndex, which Archives never noticed because they open the gallery from a single button rather than a per-photo grid. An optional initialIndex and the clicked index is the whole fix.

Cancelling the photo-delete confirmation by clicking outside it closes the entire details modal, discarding exactly the unsaved notes and link that seededForId goes to trouble to protect. PhotoGalleryModal and its ConfirmModal render inside the details overlay, whose root carries onClick={onClose}, so the backdrop click bubbles all the way up. stopPropagation on the gallery wrapper, or portalling it out of the overlay, closes it.

Two deletion paths leave the photo directory behind. Four are covered and tested; these are not:

- library.py:2045, scan_external_folder - removes rows for files that vanished from the share, unlinks the thumbnail but not library_photos_dir(id).
- print_scheduler.py:6455 - a queue item with cleanup_library_after_dispatch deletes the library row after archiving. The photos are orphaned on disk and disappear from the UI without following the file into the archive that replaced it.

The wiki says photos "leave the disk together with the file", which is true for the four and not for these two.

window.open(file.external_url!, '_blank') at FileManagerPage.tsx:890 omits 'noopener,noreferrer', so the target page gets window.opener on a URL another user stored. Every <a> in this PR gets it right, and Layout.tsx:482 - the closest analogue, a user-supplied sidebar link - passes the flags.

Smaller things, take or leave:

- No Escape-to-close on the details modal; 51 of the 60 *Modal.tsx components handle it.
- Deleting a photo from the grid is immediate while the same delete inside the lightbox asks for confirmation. The wiki documents both, so presumably deliberate.
- delete_file_photo gates on LIBRARY_UPDATE_* where the archive twin uses ARCHIVES_DELETE_*. Defensible as metadata editing, worth being a conscious call.
- The photo routes return 403 for a file the caller does not own while the read paths return 404 to avoid id enumeration. Consistent with update_file, so consistent inside library; archives route everything through _ensure_archive_visible.
- No tests cover ownership on the three photo routes, the new cleanup in delete_folder/bulk_delete, or a thumbnail click opening the gallery - the last of which would have caught the index issue.

What I verified rather than took on faith:

- Both migrations run against a live PostgreSQL install with 268 real library_files rows inside a rolled-back transaction: external_url lands as character varying(500) nullable, photos as json, all 268 rows read NULL. A repeat ALTER raises 42701, which _safe_execute's savepoint swallows without poisoning the transaction, and nothing persists after the rollback.
- The JSON column comes back as a Python list, not a string, on that same server - checked against the existing print_archives.photos rows. Worth confirming, because a str would silently turn filename not in library_file.photos into a substring match.
- Backup really does cover this: archive/ is copied wholesale and the Postgres-to-SQLite export walks metadata.sorted_tables, so both the photo files and the new columns are captured. backup.md's claim checks out.
- list_files reads notes and photos off an already-loaded entity, so the two new indicators add no queries, and there is exactly one FileListResponse construction site.
- Bundle impact is fine: dev's main chunk is already 9.91 MB, this branch 9.94 MB, with xlsx, pdf and the STEP worker split into their own chunks.
- ruff clean, backend suite 12251 passed, eslint clean, vite build clean, vitest 270 files / 3698 tests passed. check:i18n is the only failure.

Thomansky:
Thanks — all five are real and all five are fixed. Here is what each one was and what now pins it.

**The i18n gate.** Reproduced exactly: sv at 6302 leaves against 6340, all 38 of them this branch's 27 `fileManager.details.*` and 11 `fileManager.preview.*` keys including the `photoCount_one`/`_other` pair. I merged `dev` in — the gate has to pass on the merged result, so there was no way around it — and wrote the 38 Swedish strings in `en.ts` key order. Real translations, not copies; the only value that matches en is `Filament`, which is already in the locale's cognate list. `npm run check:i18n` is green with all 15 locales at 6340.

**The lightbox opening on the first photo.** `PhotoGalleryModal` had no `initialIndex` and started at 0, so the third thumbnail showed "Photo 1 of 3". It now takes an optional `initialIndex`, the details grid passes the clicked one, and the value is clamped so a stale index cannot render a missing photo. `LibraryFileDetailsModal.test.tsx` clicks the third thumbnail and asserts both the counter and the image URL — the test you said would have caught it.

**The delete confirmation closing everything.** Also right, and slightly worse than described: the confirmation's backdrop click reached the gallery's own backdrop as well, so it closed the confirmation, the gallery and the details modal in one go, taking the unsaved notes and link with them. I fixed it in `PhotoGalleryModal` rather than at the call site: the gallery stops its own backdrop click from leaving it, and the confirmation is wrapped so its dismissal stops at the gallery. That also fixes the same dismissal in the two archive galleries, which had it too. The test types a draft into notes, opens the lightbox, opens the confirmation, clicks its backdrop, and asserts the draft and the lightbox are both still there and `onClose` was never called.

**`scan_external_folder`.** Confirmed at the row-removal loop: the thumbnail was unlinked, the photo directory was not. It now calls `remove_library_photos_dir(db_file.id)` before the delete, like the other four. The test uploads a photo onto a scanned external file, deletes the file from the share, rescans, and asserts the directory is gone — it fails on the assertion without the fix.

**The dispatch cleanup.** This is the one where I did more than remove the directory. The sentence that mattered was "without following the file into the archive that replaced it": `_repoint_siblings_at_archive` already moves everything else onto the archive, and deleting the photos would have fixed the orphan while still losing the pictures of a print that still has a record. So they now move into `archive_photos_dir(archive)` and get appended to `archive.photos`, and show up in that archive's own gallery. `move_library_photos` renames around a name the destination already holds (both sides draw from the same 8-hex alphabet) and keeps the source directory if any photo could not be moved, so a failure orphans the pictures rather than destroying them. Three scheduler tests cover it: photos move and the directory goes, an external file keeps them and the archive gets none, and an archive-creation failure leaves them untouched.

**The `window.open`.** Fixed — `'noopener,noreferrer'` on the card menu's Open link, matching `Layout.tsx`. A test spies on `window.open` and asserts the third argument.

On the smaller list: I took the **Escape-to-close** (it stands down while the lightbox is open, so one press does not close both) and I added the **three missing test groups** — ownership on all three photo routes, the folder-delete and bulk-delete cleanups, and the thumbnail-click index. I left the **immediate grid delete vs confirmed lightbox delete** alone; the wiki documents both and the asymmetry is deliberate. I left **`delete_file_photo` on `LIBRARY_UPDATE_*`** as a conscious call: photos are metadata edits like notes and the link, all three follow `update_file`, and moving the delete to `LIBRARY_DELETE_*` would let someone with update-but-not-delete add a photo they could never remove. I left the **403/404 split** as it is, and it is now pinned: 403 from the two write routes, 404 from the read route through `_ensure_library_file_visible`.

The wiki sentence is corrected in a separate commit — the scan case added to the delete list, and a new paragraph for the dispatch case, which is the one flow where the photos move instead of leaving. I also updated the two lines about opening the gallery and about Escape.
---

Two things from re-reading the branch afterwards that you did not list, both real:

- **A seventh deletion path.** `bulk_delete`'s folder branch hard-deletes the library rows without removing their photo directories. Your list of four plus the two you named made six; this is the one nobody had looked at. Fixed and tested with the rest.
- **The lightbox's index was never reset** when the last photo was deleted from inside it, which left Escape unable to close the details modal afterwards.

The wiki PR is updated in the same shape: the scan case added to the delete list, and a paragraph of its own for the dispatch case, since that is the one flow where the photos move rather than leave.

Issue #3127, opened 2026-09-20
[Feature]: Run printer calibration automatically as a scheduled maintenance action (weekday schedule or when due, once the plate is released)


Thomansky:
### Problem or Use Case

Printer calibration (bed levelling, vibration compensation, motor noise — and on the H2D nozzle offset) is the one recurring maintenance task that Bambuddy could actually *perform* rather than only remind about, yet today it still means walking up to each printer, or opening Bambu Studio, and starting it by hand at a moment when the printer happens to be free. With several printers that quickly becomes a Saturday-morning ritual: wait for a printer to finish, clear the plate, start the calibration, come back later, repeat.

What I'd love: tell Bambuddy "calibrate this printer every Saturday, as soon as the plate has been released" (or "when the calibration is due") and have it happen on its own, with the maintenance log updated afterwards.

### Proposed Solution

What struck me while reading the code is that almost everything for this already exists — it just isn't connected:

- `POST /printers/{id}/calibration` already starts a calibration with a chosen option set (`bed_leveling`, `vibration`, `motor_noise`, `nozzle_offset`, `high_temp_heatbed`), but nothing in the UI calls it.
- The Maintenance page already tracks per-printer items with intervals (print hours / days) and a "performed" reset that writes history.
- Scheduled drying already has exactly the executor shape needed: a pending row with `start_after`, evaluated on every scheduler pass, deferred with a `waiting_reason` (`printer_busy`, `printer_offline`, …) until the printer is idle.
- The plate-clear gate (`awaiting_plate_clear` + `require_plate_clear`) already expresses "the plate has been released".

So the proposal is to make calibration an **actionable maintenance item**:

**1. A maintenance type can carry an action.** A new system maintenance type *Printer calibration* (per-printer option set: bed levelling / vibration / motor noise; nozzle offset / high-temperature heatbed where the model supports them — the five flags the endpoint already has). The H2D's vision-encoder calibration isn't exposed by the endpoint yet; I'm recording what Studio sends for it and would add it as a sixth flag if it turns out to be just another bit in the same command. It shows up on the Maintenance page like every other item — with a "Run now" button, which is also the missing UI for the existing endpoint.

**2. Trigger modes per item**, chosen on the maintenance card:
- **Manual** — "Run now" only.
- **When due** — the existing interval (print hours / days; print count once #3086 lands) doesn't just raise the reminder, it queues the run.
[...]

Issue #3081, opened 2026-09-11
[Bug]: Failed calibration depletes all filament weights in AMS


Arn0uDz:
### Product

Bambuddy

### Area

Filament / AMS / Spoolman

### Bug Description

Was having a couple issues with my P2S and was sorting out some stuff by running calibrations on bed level and while running the calibration I cut off the power because the hotend was hitting the bed.

I had 2 filament rolls in my AMS at the time and in the filament tab I can see that the failed calibration recorded a 1kg usage on both rolls, setting their remaining weight to 0.

### Expected Behavior

Don't touch filament weight during a calibration. It was not even loaded in the toolhead. No idea why it counted a 1000.0g weight for a calibration.

### Steps to Reproduce

Don't wanna mess up my calibration again to test. But I assume run a calibration and cut off the power. While doing this have a spools sitting in the AMS.

### Troubleshooting steps already taken

Re-added the weight of the spools.

### Printer Model

P2S

### Bambuddy Version

v1.2.5.5

### Installation Method

Docker

### Operating System

Docker

### Support Package (.zip) and / or relevant logs

Don't wanna mess up my calibration again to test. No support package. But the screenshots speak for themselves in my opinion. 

### Screenshots

<img width="913" height="1171" alt="Image" src="https://github.com/user-attachments/assets/34db9fb6-e63e-4a68-84f6-a2ff5e308541" />

<img width="907" height="1252" alt="Image" src="https://github.com/user-attachments/assets/eaa4bd8d-beb5-4f82-bce7-c6199e67f7e4" />

### Final checks

- [x] I searched existing (open AND closed) issues and this bug hasn't already been reported or resolved
- [x] I checked the [Troubleshooting wiki](https://wiki.bambuddy.cool/reference/troubleshooting/) and the relevant feature page for my issue
- [x] I am using the latest version of Bambuddy (or the latest daily build)
- [x] My printer is set to LAN Only mode
- [x] My printer has Developer Mode enabled
- [x] For any connection / printing / camera issue, I ran the in-app Connection Diagnostic and included the result above

maziggy:
Please upload a support package -> https://wiki.bambuddy.cool/features/system-info/?h=debug#enable-debug-logging

Issue #3086, opened 2026-09-12
[Feature]: Maintenance interval type: Number of prints


GennPen:
### Problem or Use Case

Some type of maintenance need to track number of prints (like purge poop container).

### Proposed Solution

Add Interval Type: Number of prints (all and successful only)

### Feature Category

Maintenance Tracking

### Priority

Nice to have

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Before we commit to building this, I'd like to gauge community interest. If you'd find this feature useful, please give this issue a thumbs up (👍) reaction so we can prioritize accordingly.

Issue #2900, opened 2026-08-21
[Feature]: Force maintanace before printing


Nick-C130:
### Problem or Use Case

The maintenance is nice but would be better with an option for number of prints instead of print hours (especially for bed cleaning) and something more then just a warning for maintenance needed

### Proposed Solution

A toggle then when enabled will not let a print start on a printer unless its maintenance is up to date
Also an option for a number of prints between maintenance requests instead of just print hours

### Feature Category

Print Queue & Scheduling

### Priority

Would improve my workflow

### Contribution

- [ ] I would be willing to help implement this feature

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested

maziggy:
Before we commit to building this, I'd like to gauge community interest. If you'd find this feature useful, please give this issue a thumbs up (👍) reaction so we can prioritize accordingly.

PR #3130, opened 2026-09-20
Run printer calibration as a scheduled maintenance action

Thomansky:
## Description

Makes printer calibration something Bambuddy can *do* on your behalf rather than only remind you about: a new system maintenance item **Printer Calibration** with a **Run now** button, a **When due** trigger (the existing interval queues the run instead of only raising the reminder) and an **On a schedule** trigger (weekdays + time). A queued run waits until the printer is idle — and, with *Require plate clear* on, until the plate has been released — then sends the calibration, and marks the item performed when the printer reports it finished. "Saturday morning, as soon as the plate is clear" no longer needs anyone in the workshop.

On the way it fixes how the printer's own calibration runs are handled at completion (#3081): they no longer deduct filament, raise the plate-clear prompt or get matched against the queue.

## Related Issue

Fixes #3127
Fixes #3081

## Documentation

**Companion docs PRs**:
- Wiki: maziggy/bambuddy-wiki#92

**Pick one**:
- [x] Docs PR(s) linked above

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition or update

## Changes Made

- **Calibration flags** — `start_calibration` / `POST /printers/{id}/calibration` gain `micro_lidar` (bit 0) and `nozzle_clumping` (bit 6), completing Bambu Studio's seven-option set (verified against Studio's `Calibration.cpp` and an H2S MQTT capture); the five existing flags are untouched
- **Actionable maintenance type** — `MaintenanceType.action` ("calibration"), a seeded system type *Printer Calibration* (100 print hours); per item: `action_options` (the seven flags, default bed levelling + vibration + motor noise), `trigger_mode` (`manual` | `when_due` | `schedule`), `schedule_days`, `schedule_time` (server-local, like the local backup schedule), computed `schedule_next_at`, `last_auto_run_at`
- **Runs** — new `maintenance_runs` table (pending → running → completed / failed / cancelled, with `source`, `waiting_reason`, `error_message`); one active run per item, enforced by a partial unique index; routes `POST /maintenance/items/{id}/run`, `GET /maintenance/items/{id}/runs`, `DELETE /maintenance/runs/{id}`; overview/status responses carry the new fields and the current run
- **Executor** — `_check_maintenance_runs` in the scheduler pass, right after scheduled drying and with the same deferral shape: offline / drying / queue-reserved (a queue item in *printing*, an upload in flight, a dispatch hold) / not idle → the row stays pending with a `waiting_reason` (`printer_offline`, `already_drying`, `printer_busy`, `awaiting_plate_clear`); the plate-clear gate is honoured through `_is_printer_idle(require_plate_clear=<setting>)`; the printer counts as busy for the print queue while it calibrates; stale *running* rows are closed after 2 h
- **Completion** — the printer's own job (`/usr/etc/print/<model>/auto_cali_for_user_param.gcode`, `print_type: system`) closes the run: FINISH → completed, history entry "Automatic calibration" at the current hours, interval reset; a cancel from the printer screen → cancelled (the cancel code `50348044` arrives seconds after the FAILED edge, so the outcome is decided after a short grace period); any other FAILED → failed with the message. Cancelling from Bambuddy stops the printer only when it is actually on the calibration
- **Internal-job handling in `on_print_start` / `on_print_complete`** — detected first, before every print-specific side effect: no archive lookup, no usage tracking or session discard, no plate-clear gate, no queue reconciliation, no SD cleanup, no notification, no kill-switch reconciliation; the state and hour bookkeeping that must still happen is kept. `is_internal_printer_job` learns the `auto_cali_for_user_param` spelling current firmware reports; the kill-switch exemption keys on the `/usr/` path (the strong signal), not on the bare name
- **Vision Encoder Calibration (H2 series)** — a second actionable system type (`action = "motion_precision"`, 7 days, H2S / H2D / H2D Pro / H2C only) that runs the printer's own `calibrate_motion_precision.gcode` through the `gcode_file` command, so the manual "vision encoder" reminder people set up is no longer needed. The model-specific directory under `/usr/etc/print/` is learned from the printer's own jobs (every H2S reports `O1S`; the H2D/H2C fallbacks are unverified guesses), the printer's reply to the command is awaited and a refused path fails the run at once with the firmware's code, and completion is matched per action, so a vision-encoder FINISH cannot close a levelling run. The cover route answers the printer's own jobs with 404 without the FTP sweep, and neither status producer hands out a `cover_url` for them
- **Start condition “only when the bed is below … °C”** — optional per item (both actions), stored as `bed_temp_below` in `action_options` (0 < value ≤ 120, one decimal). The executor checks it last, after every other gate, against the live bed temperature the bed-cooled notification already uses: warmer or unknown → the run stays pending with `bed_too_warm` (the current temperature in the new `waiting_detail` JSON, shown as “Waiting: bed still warm (34 °C)”) or `bed_temp_unknown`, re-evaluated every pass, never timing out. The threshold is read from the item, so changing it releases a waiting run
- **Notifications** — a bell on every maintenance card mutes that item (`notifications_enabled`, default on): a muted item raises no due/warning reminder. New provider event *Maintenance run finished* (`on_maintenance_run`, off by default so nobody gets new messages after the upgrade) reports completed / failed / cancelled runs with printer, item and message through the normal template machinery
- **Maintenance page** — the calibration card shows the option checkboxes (nozzle offset only on dual-nozzle printers), the trigger selector — the two automatic triggers say what gates the start, “(only once the plate has been released)” with *Require plate clear* on, “(once the printer is idle)” without —, weekday chips + time, a status line (next run, waiting reason, running since, last result) and Run now / Cancel; the Settings tab marks the type as one that runs a calibration
- Migrations: idempotent `_safe_execute` ALTERs for the new columns, `CREATE TABLE IF NOT EXISTS` + partial unique index for `maintenance_runs`, placed before the must-run-last blocks
- i18n: 42 new keys in all 14 locales, `npm run check:i18n` green

## Screenshots

| Before — a reminder-only card | After — the calibration card |
|---|---|
| before | after |

**Maintenance page with the calibration card, a Saturday schedule and a run waiting for the printer**

maintenance page

## Testing

- Backend, 268 new tests: `test_scheduler_maintenance_runs.py` (71 — future `start_after` not dispatched; offline / drying / busy / queue-reserved / awaiting-plate-clear deferrals with the right reason; `require_plate_clear` off dispatches on a FINISH printer; bitmask incl. bits 0/6; one run per item; due-trigger fires once; schedule math incl. week wrap and DST; bed condition: warm/exact/unknown wait, dispatch once below, no condition ignores a hot bed), `test_maintenance_actions.py` (77), `test_maintenance_runs_api.py` (83 — run 409/400, cancel with and without a real calibration on the printer, PATCH validation, overview fields), `test_internal_job_lifecycle_3081.py` (8 — drives the real `on_print_start` / `on_print_complete`: no plate-clear gate, no usage tracker, no archive, no notification; replays the capture's two-message screen-cancel sequence), `test_printer_kill_switch.py` (19), `test_cover_internal_job_3127.py` (10)
- Regression suites re-run: print lifecycle, printers API, print queue, maintenance API, scheduled drying, clear-plate scheduler tests — green apart from the known Windows-only path/timing cases
- Frontend: `MaintenancePageCalibration.test.tsx` (21), `NotificationProviderCardMaintenanceRun.test.tsx` (4), `AddNotificationModal.test.tsx` (+3); full suite green (3689 tests, 268 files), i18n parity green
- Protocol facts (option bits, the printer's own job name and path, FINISH / FAILED edges, the late cancel code) come from an H2S MQTT capture of Studio-driven calibrations, replayed in the lifecycle tests
- Live on an H2S: *Run now* queued the run, the scheduler dispatched it on the next pass, the printer ran bed levelling + vibration compensation + motor noise cancellation (52 min), and on the printer's FINISH the run closed as completed with the "Automatic calibration" history entry and the interval reset — no archive, no filament deduction, no plate-clear prompt along the way
- A vision-encoder calibration started on another printer's screen during that time was recognised as an internal job and left alone

- [x] I have tested with my printer model: H2S (H2D and X1E connected, not yet used for a run)

## Checklist

## Additional Notes

- Deliberately narrow: two executable maintenance items with a trigger, not an automation engine (#1016). Out of scope, as listed in the issue: PA / flow-dynamics calibration.
- The vision encoder item is a 7-day interval and, like every day-based item, counts as due until it has been performed once — so every H2 printer shows it as **Due** right after the upgrade. Its trigger defaults to *Manual*, so nothing runs by itself; switching it to *When due* queues a run straight away, and *Reset* clears it without one.
- The maintenance page now shows the seeded system types under their translated names (the locale keys existed but were never read); custom and renamed types are shown as stored.
- After a run that failed or was cancelled, a *When due* item does not re-queue itself for the same due period; Run now (or Reset) restarts it. Documented in the wiki PR.
- Composes with #2900 (block printing while maintenance is due) and #3086 (print-count intervals) on the same item model.


maziggy:
Does it respect all the various printer models and their specifics? For example: Micro Lidar calibration is not avialble for all printer models.

Thomansky:
Partly, and you picked exactly the one that was wrong. Pushed a fix.

What was already gated per model:

- **Nozzle offset** — only offered on dual-nozzle models, through `is_dual_nozzle_model`.
- **Vision encoder calibration** — refused outright on a model without one (`has_vision_encoder`, H2 series), and the model gate runs at two levels: the printer picker in "Add type" only lists printers the action can run on, and a run on an ineligible printer is turned down rather than dispatched. There is also a fallback path for the internal G-code directory, which logs a warning when it has to guess the model's directory instead of using the one the printer reported for one of its own jobs.

What was **not** gated, and should have been:

- **Micro Lidar.** `available_calibration_options` offered it everywhere, on the reasoning that firmware ignores bits for hardware it does not have. That is true of the wire and useless to the user: a P1S owner was shown a tick box for a calibration the machine cannot run. What makes it worse is that `start_calibration` has documented bit 0 as "(X1 series)" since it was written, so the knowledge was already in the file — the gate was simply never applied to the options the UI offers. There is now a `MICRO_LIDAR_MODELS` list next to `VISION_ENCODER_MODELS` and `DUAL_NOZZLE_MODELS`, and a test that pins the box to X1 / X1C / X1E and away from P1S, A1 and the H2 series.

One I deliberately left alone: the **high-temperature bed**. In practice it reads like a P2S / H2 feature, but not from a source I would gate on, and guessing wrong there hides a calibration from someone who needs it rather than merely offering a useless one. If you know the model list, I will add it the same way.

Keeping this in draft for now: the vision encoder path has been exercised against the MQTT capture but not yet on a real H2D. I would rather run it once on hardware before asking you to merge it.

